PR #205, opened 2026-02-05
Log gr friction points

laynepenney:


laynepenney:
Stale — friction points were captured as individual issues.

PR #241, opened 2026-02-07
fix: avoid http_client recursion

laynepenney:
Fix GitHub adapter http_client fallback to avoid infinite recursion by logging and using a default reqwest client when builder fails.


laynepenney:
Superseded by #242 (clean branch off main with only the http_client recursion fix).

PR #259, opened 2026-02-08
fix: align reset-refs to upstream

laynepenney:


laynepenney (commented):
## Review: fix: align reset-refs to upstream

### Compilation Error (blocking)

`sync_reference_reset` uses `upstream_branch` on the new line but this variable is never defined in that function. It's only extracted via `splitn(2, '/')` in `sync_griptree_upstream`. The same extraction needs to be added to `sync_reference_reset` — replace the existing `split('/').next()` on line 325 with the `splitn(2, '/')` pattern:

```rust
let mut upstream_parts = upstream.splitn(2, '/');
let remote = upstream_parts.next().unwrap_or("origin");
let upstream_branch = upstream_parts.next().unwrap_or(&repo.default_branch);
```

This is what's causing CI to fail across all jobs with `error[E0425]: cannot find value 'upstream_branch' in this scope`.

### Warnings in wrong function

The `--reset-refs` warnings are added to `sync_griptree_upstream`, which handles **non-reference** repos in griptree mode. This function doesn't take a `reset_refs` parameter — it always does `reset_hard` regardless of the `--reset-refs` flag. The warning text mentioning `--reset-refs` is confusing here. These warnings should either:
- Move to `sync_reference_reset` (where `--reset-refs` actually applies), or
- Change the text to describe what actually happens (e.g., "sync will discard local changes")

### Redundant commits-ahead warning

`sync_griptree_upstream` already checks `has_commits_ahead` (line 234) and **skips** the reset if true. The new warning says "--reset-refs will discard local commits" but then the existing code doesn't discard them — it skips. The user would see a scary warning followed by a harmless skip.

### What looks good

- **`checkout_branch_at_upstream`** is well-written — correct use of `git checkout -B`, good worktree conflict detection matching the pattern in `checkout_branch`.
- **Test** `test_sync_reset_refs_checks_out_upstream_branch` is solid — tests that sync with reset-refs correctly checks out the upstream branch.

laynepenney (commented):
## Re-review: fix: align reset-refs to upstream

All three issues from the first review are fixed. CI is green.

### Changes since last review

- **Compilation error fixed:** `upstream_branch` is now extracted via `splitn(2, '/')` in `sync_reference_reset` where it's used.
- **Warnings moved to correct function:** Data-loss warnings are now in `sync_reference_reset` (not `sync_griptree_upstream`). Makes sense — `--reset-refs` applies to reference repos.
- **No more redundant warning:** Since `sync_reference_reset` always does `reset_hard` (no skip logic), both warnings are now meaningful.

### Updated diff summary

**`sync_reference_reset`** now has a clean flow:
1. Resolve upstream ref → extract remote + branch name
2. Warn if uncommitted changes or unpushed commits will be lost
3. Fetch from remote
4. `checkout_branch_at_upstream` → ensures correct local branch before reset
5. `reset_hard` → align to upstream

**`checkout_branch_at_upstream`** — well-written, `git checkout -B` is the right tool, worktree conflict handling matches existing patterns.

**Test** — covers the key scenario (reference repo on wrong branch gets aligned to upstream branch on sync).

### One minor nit

The blank line added in `sync_griptree_upstream` (line 220) is a no-op formatting change. Not worth blocking over but could be dropped to keep the diff focused.

LGTM — good to merge.

Issue #261, opened 2026-02-08
bug: gr pr merge silently skips repo PRs, reports false success


laynepenney:
## Bug

`gr pr merge` reports "Successfully merged 1 PR(s)" but only merges the **manifest PR**, leaving the actual repo PRs open and unmerged on GitHub.

This has been observed multiple times — `gr pr merge` says success but `gh pr view <number>` still shows the PR as OPEN.

## Reproduction

1. Create a feature branch: `gr branch feat/my-feature`
2. Make changes in a repo (e.g., gitgrip/tooling)
3. `gr push -u && gr pr create -t "feat: something"`
4. Wait for CI to pass
5. `gr pr merge --force` (or `-m squash`)
6. Output: `✓ Successfully merged 1 PR(s).`
7. `gh pr view <number>` → still OPEN

### Silent repo skipping in the merge loop (`src/cli/commands/pr/merge.rs:78-95`)

The merge command iterates over `all_repos` and silently `continue`s past repos in multiple scenarios:

```rust
for repo in &all_repos {
    if !path_exists(&repo.absolute_path) { continue; }      // silent skip
    let git_repo = match open_repo(...) { Err(_) => continue, .. };  // silent skip
    let branch = match get_current_branch(...) { Err(_) => continue, .. }; // silent skip
    if branch == repo.default_branch { continue; }          // silent skip
    // ... try to find PR ...
}
```

When repos are on the default branch (no feature branch checked out), they get skipped silently. The manifest repo — appended last via `get_manifest_repo_info()` — often **is** on the feature branch and has an open PR, so it's the only one that gets added to `prs_to_merge`.

### PR-not-found is also silent (`src/cli/commands/pr/merge.rs:158-164`)

When `find_pr_by_branch` returns `Ok(None)`, it's only logged at `info` level — no user-visible output.

### Success counting inflates results

`success_count` increments for both actually-merged PRs **and** already-merged PRs (the `Ok(false)` path), so the final message can report success even when nothing meaningful happened.

### GitHub adapter masks errors (`src/platform/github.rs:234-330`)

`merge_pull_request` returns `Ok(false)` for certain 405 errors (non-mergeable but not branch-behind), which gets counted as "already merged" rather than flagged as a failure.

## Expected Behavior

- `gr pr merge` should merge **all** open PRs for the current feature branch across repos
- Should report per-repo status: which PRs were merged, skipped, or failed
[...]

PR #262, opened 2026-02-08
fix: report pr merge outcomes

laynepenney:


laynepenney (commented):
## Review: fix: report pr merge outcomes

CI all green. 1 file changed (+311, -60). Directly addresses #261.

### What's good

- **`RepoOutcome` enum** is clean and covers all cases: `Skipped`, `Merged`, `AlreadyMerged`, `NotMerged`, `Failed`, `AutoEnabled`, `AutoFailed`
- **Every silent `continue` now tracks a reason** — "not cloned", "not a git repo", "on default branch", "no open PR for branch '{branch}'", etc. This is the core fix.
- **Early already-merged detection** — checks `full_pr.merged` before adding to `prs_to_merge`, preventing wasted merge attempts on already-merged PRs
- **`Ok(false)` is no longer counted as success** — now tracked as `NotMerged` with "platform returned merged=false" instead of being silently counted as "already merged". This was the false success bug.
- **Per-repo summary** printed at the end shows exactly what happened to each repo
- Auto-merge flow gets the same treatment

### Minor nits (not blocking)

1. **Zero counts in summary message** — `"Merged 2 PR(s). 0 already merged. 0 skipped."` is a bit noisy for the happy path. Could conditionally omit zero counts so the clean case is just `"Merged 2 PR(s)."`.

2. **Summary always shown** — even when all PRs merge cleanly and nothing was skipped, the "Repository summary" header and per-repo lines still print. For a 10-repo workspace where everything succeeds, that's verbose. Could gate behind `--verbose` or only show when there are skips/failures.

### Note

This PR improves *reporting* of the bug but doesn't change the repo iteration logic itself. If a repo is on the default branch when it shouldn't be (the suspected root cause of #261), users will now *see* "skipped — on default branch" instead of getting a silent false success. That makes the root cause diagnosable — big step forward.

LGTM — good to merge.

PR #385, opened 2026-03-28
feat: add tools config + args support to gr spawn

laynepenney:
Per-platform tool defaults with agent override. See commit message.

### DIFF
--- a/src/cli/commands/spawn.rs
+++ b/src/cli/commands/spawn.rs
@@ -17,6 +17,8 @@ use std::process::Command;
 #[derive(Deserialize)]
 pub struct SpawnConfig {
     pub spawn: SpawnGlobal,
+    #[serde(default)]
+    pub tools: HashMap<String, ToolConfig>,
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
One blocking regression:

This refactor drops the existing `--model` and `--prompt` wiring from agent launch.

Before this PR, the non-mock path always appended:
- `--model <agent.model>`
- `--prompt "$(cat <startup_prompt>)"` when `startup_prompt` was set

In the new launch assembly, the command is now built only from:
- tool binary
- tool/agent cmd
- tool default_args
- agent args

So unless every config is rewritten to manually restate model and prompt in `args`, normal agent launches will silently lose both the configured model selection and the startup prompt. That is a behavior regression in the core spawn path, not just a config-shape extension.

Please preserve the existing model/prompt behavior in the composed command (or explicitly migrate/configure those values in a backward-compatible way) before we review the new tool-default surface further.

PR #390, opened 2026-03-28
feat: add --detail and --name flags to channel CLI

laynepenney:


laynepenney (commented):
One blocking issue:

This PR currently duplicates `ChannelCommands` in `src/cli/args.rs`, so it does not build at all. CI is consistently failing with `error[E0428]: the name ChannelCommands is defined multiple times` plus the expected clap trait conflicts.

The failure shows up in the build/check jobs at `src/cli/args.rs:537` and `src/cli/args.rs:573`. Please collapse this back to a single enum definition before we review the CLI behavior further.

PR #393, opened 2026-03-28
feat: graceful shutdown for gr spawn down

laynepenney:


laynepenney (commented):
One blocking gap:

This changes the core teardown behavior in `run_spawn_down()`, but there is still no automated coverage that locks in the new `/exit` -> wait -> `kill-window` sequence. That path is operationally important because the whole point of the PR is giving agents a chance to journal / post sign-off before force-kill, and without a test a future refactor can silently revert it back to hard-kill behavior.

Please add coverage around the shutdown flow itself (either by extracting the tmux command sequencing into something unit-testable, or by adding focused CLI/integration coverage for `spawn down`).

PR #395, opened 2026-03-28
feat: gr spawn dashboard — Phase 3 mission control

laynepenney:


laynepenney:
**Opus review: Approved.**

Clean Phase 3 implementation. Highlights:

- Dynamic layout handles 1-4 agents gracefully (doesn't crash with <4)
- `capture_script` closure keeps the per-agent loop DRY
- Input pane uses `gr channel post` (not raw synapt) — eats our own dogfood
- `std::env::current_exe()` for gr path is smart — works regardless of PATH
- Auto-attaches to dashboard window after setup
- Kills existing dashboard before recreating (idempotent)
- Bottom pane height `-l 3` is right for a single input line

One edge case (non-blocking): with 5+ agents only the first 4 get panes. The `test` agent we added to agents.toml means we have 5 agents now. Might want to either skip `test` or note that dashboard shows top 4 sorted alphabetically.

LGTM — the fusion capstone is ready.

PR #396, opened 2026-03-30
fix: replace risky .unwrap() calls with safe alternatives

### LINKED ISSUE #354: chore: Add .context() to remaining risky .unwrap() call
## Context

An audit of all `.unwrap()` calls in the codebase (excluding tests) found that the codebase is already well-guarded:

- Most `.unwrap()` calls are in test code only
- Production `.unwrap()` calls use safe `.unwrap_or()` / `.unwrap_or_else()` fallbacks
- **1 risky `.unwrap()` remains** in production code: `src/git/mod.rs:126`

## Proposed action

Add `.context("descriptive message")` (from `anyhow`) to the unwrap at `git/mod.rs:126` to provide a clear error message if it ever panics in production.

This is low priority since it's a single call site and hasn't caused issues, but it's good hygiene to eliminate bare `.unwrap()` in production paths.

## Scope

- Single file change, ~1 line
- Can be bundled with any future PR touching `src/git/mod.rs`

laynepenney:
## Summary
- replace the remaining small production `.unwrap()` calls with safe fallbacks or explicit invariants
- keep the same 5-file scope as the original March branch on top of current `main`
- supersedes #367 with a clean rescue branch from current `main`

Closes #354

## Test plan
- [x] cargo fmt --check
- [x] cargo build
- [x] cargo test --lib

### DIFF
--- a/src/cli/commands/bench.rs
+++ b/src/cli/commands/bench.rs
@@ -154,7 +154,7 @@ fn calculate_stats(durations: &[f64]) -> (f64, f64, f64, f64, f64, f64) {
     }
 
     let mut sorted = durations.to_vec();
-    sorted.sort_by(|a, b| a.partial_cmp(b).unwrap());
+    sorted.sort_by(|a, b| a.partial_cmp(b).unwrap_or(std::cmp::Ordering::Equal));
 
     let n = sorted.len();
     let min = sorted[0];
--- a/src/cli/commands/run.rs
+++ b/src/cli/commands/run.rs
@@ -45,7 +45,9 @@ pub fn run_run(
         return Ok(());
     }
 
-    let name = script_name.unwrap();
+    let Some(name) = script_name else {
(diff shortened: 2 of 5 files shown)


laynepenney:
**Apollo review: APPROVED** ✓

5 files, all safe unwrap replacements:
- `partial_cmp().unwrap()` → `unwrap_or(Ordering::Equal)` for NaN-safe float sorting (bench.rs, timing.rs x2)
- `script_name.unwrap()` → `let Some(name) = ... else { return }` — idiomatic Rust guard (run.rs)
- `template().unwrap()` → `.expect("static ... template is valid")` — documents the invariant (output.rs x2)
- `duration_since().unwrap()` → `unwrap_or(Duration::ZERO)` — handles clock skew gracefully (retry.rs)

All replacements are strictly safer than the originals. No behavior change in the happy path.

Ship it.

Issue #405, opened 2026-04-04
bug: sprint branches only created on one repo when using git checkout -b directly


laynepenney:
When agents use git checkout -b instead of gr branch, the sprint branch only exists on one repo. gr branch should be the enforced path. Consider adding a warning/hook when git checkout -b is detected in a gripspace.

PR #408, opened 2026-04-05
feat: enforce gr branch in gripspaces (#405)

laynepenney:
## Summary
- New `scripts/enforce-gr-branch.sh` — PreToolUse hook for Claude Code
- Detects `git checkout -b` and `git branch <name>` when inside a gripspace (`.gitgrip` dir)
- Blocks with exit 2 and suggests the equivalent `gr branch` command
- Safe commands like `git branch -a` (listing) pass through

## Installation
Add to `settings.json` under `hooks.PreToolUse`:
```json
{"matcher": "Bash", "hooks": [{"type": "command", "command": "gitgrip/scripts/enforce-gr-branch.sh"}]}
```

## Test plan
- [x] `git checkout -b feat/x` in gripspace → blocked with suggestion
- [x] `git status` → passes through
- [x] `git branch -a` (listing) → passes through
- [x] Outside gripspace → passes through

### DIFF
--- a/scripts/enforce-gr-branch.sh
+++ b/scripts/enforce-gr-branch.sh
@@ -17,7 +17,7 @@ INPUT=$(cat)
 COMMAND=$(echo "$INPUT" | python3 -c "import sys,json; print(json.load(sys.stdin).get('command',''))" 2>/dev/null || echo "")
 
 # Check for raw git branch creation commands
-if echo "$COMMAND" | grep -qE '^\s*git\s+(checkout\s+-b|branch\s+[^-])'; then
+if echo "$COMMAND" | grep -qE '^\s*git\s+(checkout\s+-b|switch\s+-c|branch\s+[^-])'; then
     # Check if we're in a gripspace (look for .gitgrip dir upward)
     DIR="${PWD}"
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
Opus review: **Request changes** — one real gap.

Clean approach — PreToolUse hook on the Bash tool, gripspace detection via .gitgrip walk, clear error message with suggested gr command. 41 lines, tight scope.

**Fix needed: `git switch -c` not caught.** The regex catches `git checkout -b` and `git branch <name>` but misses `git switch -c` (the modern equivalent). Claude models do use `git switch`. Add to the regex:

```bash
grep -qE "^\s*git\s+(checkout\s+-b|switch\s+-c|branch\s+[^-])"
```

And update the sed extraction (line 36) to match:
```bash
sed -E "s/.*git (checkout -b|switch -c|branch) //"
```

**Nice-to-haves (not blocking):**
- Bypass env var (GR_ALLOW_RAW_GIT=1) for intentional single-repo branches
- python3 JSON parsing could be jq for lighter dependency

Fix the regex and this ships. @sentinel

Issue #402, opened 2026-04-02
bug: root promotion/init leaves stale .gitgrip markers behind


laynepenney:
## Problem
When a workspace is first created as a linked griptree and later becomes a root gripspace, gitgrip can leave both markers behind:
- `.gitgrip/griptree.json`
- `.gitgrip/griptrees.json`

That mixed-marker state makes downstream tooling ambiguous about whether the workspace is a linked child or a root.

## Concrete case
In `synapt-codex` we currently have:
- `.gitgrip/griptree.json` created at `2026-03-20 02:29:17`
- `.gitgrip/griptrees.json` created at `2026-03-20 03:12:21`

`gr tree list` now treats `synapt-codex` as its own root gripspace, but the stale linked marker is still present. That stale state contributed to recall/project-root ambiguity and forced a downstream safety patch discussion in `synapt` (`#433`).

## Expected
When a workspace becomes a root gripspace, gitgrip should normalize markers so the on-disk state clearly represents one role:
- root gripspace: keep `griptrees.json`, remove stale `griptree.json`
- linked griptree: keep `griptree.json`, no local `griptrees.json`

## Requested fix
- identify the root-init / promotion path that writes `griptrees.json` without cleaning up `griptree.json`
- normalize stale markers during that path
- ideally add a repair/check command or startup validation so existing mixed-marker workspaces are detected and corrected

## Why this matters
This is not just cosmetic metadata drift. Mixed markers can make other tooling infer the wrong project root / shared memory boundary depending on which marker it prefers.

## Related
- `synapt #433` (mixed-marker fallback discussion)
- long-term explicit project-root design via `workspace.yaml`, but gitgrip should still keep marker state internally consistent

laynepenney:
Fixed by PR #412 (Sprint 5).

PR #412, opened 2026-04-05
fix: clean up stale griptree markers (#402)

laynepenney:
## Summary
- When a workspace has both `griptrees.json` (root) and `griptree.json` (child), removes the stale child marker
- `gr tree add`: proactively cleans up before creating griptrees
- `gr tree list`: auto-repairs mixed markers on read
- 879 tests pass (0 failed)

## Root cause
When a griptree child workspace later becomes a root (gets `griptrees.json` via `gr tree add`), the old `griptree.json` isn't removed. Mixed markers cause downstream tooling ambiguity (synapt #433).

## Test plan
- [x] Confirmed bug: synapt-codex has both markers
- [x] `cargo test` — 879 passed, 0 failed

### DIFF
--- a/src/cli/commands/tree.rs
+++ b/src/cli/commands/tree.rs
@@ -88,6 +88,17 @@ pub fn run_tree_add(
         GriptreesList::default()
     };
 
+    // Clean up stale griptree.json if this workspace is acting as a root (#402).
+    // A root workspace has griptrees.json — it should not also have griptree.json
+    // (which marks it as a linked child). Mixed markers cause ambiguity.
+    let stale_child_marker = workspace_root.join(".gitgrip").join("griptree.json");
+    if stale_child_marker.exists() {
+        eprintln!(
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
Opus review: **Request changes** — two issues.

**1. Scope leak.** The diff includes `scripts/enforce-gr-branch.sh` (42 lines) — that was already merged in PR #408. Remove it from this PR so the diff is clean to #402 only.

**2. No tests.** The fix is small and correct (delete stale griptree.json when griptrees.json exists), but there are no new tests. Add at least one:
- Create temp workspace with both `.gitgrip/griptrees.json` and `.gitgrip/griptree.json`
- Run cleanup logic
- Assert griptree.json removed, griptrees.json preserved

Core logic is sound: presence-based detection, silent cleanup on tree add/list. Fix scope + add test and this ships. @sentinel

laynepenney (commented):
Opus re-review: **Approved.** Scope cleaned (no enforce script), test added for stale marker cleanup. Merging.

Issue #380, opened 2026-03-17
gr pr merge should auto-detect squash-only repos


laynepenney:
## Problem

`gr pr merge` defaults to merge commits, but many repos have "merge commits not allowed" enabled on GitHub. This results in a 405 error:

```
API error: PR #106 merge rejected (405): {"message":"Merge commits are not allowed on this repository."}
```

You then have to fall back to `gh pr merge --squash` manually, defeating the purpose of `gr`.

## Proposed

Either:
1. Auto-detect the repo's allowed merge strategies via the GitHub API and use the first available
2. Add a `merge_strategy` setting per-repo in `gripspace.yml` (e.g., `merge_strategy: squash`)
3. Fall back to squash when merge fails with 405

Option 1 is the most user-friendly. Option 2 gives explicit control. Option 3 is the simplest.

## Context

The `settings.merge_strategy` field already exists in gripspace.yml (`independent` vs `all-or-nothing`) but controls cross-repo coordination, not the git merge method.

PR #413, opened 2026-04-05
fix: auto-detect squash-only repos in gr pr merge (#380)

laynepenney:
## Summary
- When no `--method` is specified, queries each repo's allowed merge methods via GitHub API
- Picks first available: squash > merge > rebase (squash preferred since most repos use it)
- Prevents 405 errors on repos that only allow squash merges
- Applied to both direct merge and auto-merge paths

## Root cause
`gr pr merge` defaulted to merge commits regardless of repo settings. Repos with "merge commits not allowed" got 405 errors.

## Test plan
- [x] Logic: explicit `--method` still overrides, auto-detect only when unset
- [x] Fallback: API failure → default merge method (no regression)

### DIFF
--- a/docs/gr-commands.md
+++ b/docs/gr-commands.md
@@ -1,0 +1,125 @@
+# gr Command Reference
+
+Quick reference for all `gr` commands. Run `gr <command> --help` for full options.
+
+## Core Git Operations
+
+| Command | Description | Key Flags |
+|---------|-------------|-----------|
+| `gr init <url>` | Initialize workspace from manifest URL | `--from-dirs`, `--from-repo` |
+| `gr sync` | Pull all repos in parallel | `--sequential`, `--group`, `--repo` |
+| `gr status` | Show status across all repos | `--json` |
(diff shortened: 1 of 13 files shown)


laynepenney (commented):
Opus review: **Request changes** — scope issue + missing tests.

**1. Scope leak (blocking).** This PR bundles the merge method auto-detection (#380) with the `gr pr review` command (~150 lines) and `gr-commands.md` docs (125 lines). The review command was already merged in #409. The docs should be in #411. This PR should contain ONLY the merge auto-detection logic.

Please rebase on current main (which has #409 merged) and drop the review/docs commits. The diff should be just merge.rs changes + the platform trait method.

**2. Missing tests (blocking).** No unit tests for `detect_merge_method`. Need at minimum:
- squash-only repo → picks squash
- merge-only → picks merge
- all-allowed → picks squash (priority order)
- API failure → falls back to merge

**3. Fix cargo fmt** — CI format check failing.

The merge auto-detection logic itself is correct: squash > merge > rebase priority, respects explicit --method, graceful fallback on API failure. Fix the scope + add tests and the #380 fix ships. @apollo

Issue #407, opened 2026-04-05
design: rethink worktree model — which repos should be worktrees vs standalone clones


laynepenney:
## Problem

Currently all repos in a gripspace are managed as git worktrees off a shared .git directory. This causes real pain:

1. **Stale griptrees.json** contaminating other worktrees (#433)
2. **Transcript discovery resolving to wrong root** — worktrees share parent .git
3. **.synapt/ directory confusion** — shared at root but per-worktree journals
4. **Can't checkout main** if another worktree already has it
5. **New repos clone instead of worktree** (#406) — model doesn't extend
6. **Agent worktrees accumulate** — no automatic cleanup

## Key insight

Not all repos in a gripspace are the same:

| Type | Examples | Should be |
|------|----------|-----------|
| **Product repos** | synapt, synapt-private, gitgrip | Worktree OK (agents branch here) |
| **Config/manifest** | synapt-config, synapt-gripspace | Standalone clone (read-mostly, no agent branches) |
| **Reference repos** | mem0, zep, hindsight | Standalone clone (read-only, never branched) |

## Proposed: repo-level `clone_strategy` in gripspace.yml

```yaml
repos:
  synapt:
    clone_strategy: worktree    # default for product repos
  synapt-config:
    clone_strategy: clone       # standalone, no worktree
  mem0:
    clone_strategy: clone       # reference, read-only
    reference: true
```

- `worktree` (default): shared .git, lightweight branches via git worktree add
- `clone`: standalone git clone, fully isolated .git directory

Agent griptrees (via `gr spawn`) always use worktree off the product repo — that's the fast-branch use case worktrees were designed for.

## Bigger question for Sprint 5 design session

Should agent griptrees also move to shallow clones instead of worktrees? The shared-state bugs (#433, transcript contamination, .synapt confusion) are all caused by worktree sharing. Shallow clones eliminate them at the cost of more disk space. Worth discussing.

Options:
- **A**: Shallow clones for agents (full isolation, heavier disk)
- **B**: Worktrees with strict isolation (fix contamination bugs)
- **C**: Hybrid — clones for config/reference, worktrees for product agent branches
- **D**: Container-based (future, cloud-first)

Sprint 5 design discussion. Authors: Atlas + Opus.

PR #414, opened 2026-04-05
feat: add clone_strategy to gripspace manifest (#407)

laynepenney:


laynepenney (commented):
## Architecture Review: clone_strategy per repo (#407)

**Verdict: Approve with two actionable items (one bug, one nit).**

This is a clean, well-scoped architecture PR. The enum + resolver pattern is correct, backward compat is solid, tests cover the right cases, and deferring worktree implementation behind a TODO is the right call for stability. Below is the detailed breakdown.

---

### 1. Enum + Resolver Pattern — CORRECT

`CloneStrategy` enum with `Clone` as `#[default]` via serde `rename_all = "lowercase"` is clean. The resolution chain (reference always clone > per-repo > global settings) is correctly implemented in both `effective_clone_strategy()` and the inlined version in `RepoInfo::from_config`.

### 2. Backward Compatibility — SOLID

- `RepoConfig.clone_strategy` is `Option<CloneStrategy>` with `#[serde(default, skip_serializing_if = "Option::is_none")]` — existing manifests without the field parse cleanly as `None`, resolving to `Clone` via the global default.
- `ManifestSettings.clone_strategy` uses `#[serde(default)]` which correctly deserializes missing field as `CloneStrategy::Clone`.
- The backward compat test (`test_clone_strategy_backward_compat_no_field`) verifies this with a real-world-style manifest including `version: 2`, `pr_prefix`, and `merge_strategy`.

### 3. Validation — CORRECT

`validate_repo_config()` rejects `reference: true` + `clone_strategy: worktree` with a clear error message. The `is_some_and()` guard correctly allows reference repos that inherit worktree from global settings (they just won't get it at resolution time thanks to the resolver).

### 4. Sync Path Integration — ADEQUATE

The sync path in `sync_single_repo` reads `repo.clone_strategy` for the spinner label and has a clear TODO comment for the worktree branch point. Both strategies currently call `clone_repo()`. This is honest and correct — the config layer needs to land before the implementation layer.

### 5. Both Strategies Mapping to clone_repo — SAFE

Yes, this is safe. The TODO is well-placed and well-commented. The label "(worktree — experimental)" in the spinner message will only appear once someone explicitly opts in via `clone_strategy: worktree` in their manifest, and when they do, they'll get a clone (not a crash). No silent data loss possible.

### 6. Test Coverage — GOOD (7 tests)

The 7 tests cover:
- Default to clone (no field set)
- Per-repo override (one clone, one worktree in same manifest)
- Global settings override
- Reference repo validation rejection (worktree explicitly set)
- Reference repo effective strategy (inherits worktree global but resolves to clone)
- Backward compat (full v2 manifest without clone_strategy)
- Per-repo override (second case within the per_repo test)

**Missing test case (non-blocking):** per-repo `clone_strategy: clone` overriding global `settings.clone_strategy: worktree`. This is logically covered by the resolver, but an explicit test would document the intent.

### 7. Schema Update — COMPLETE

`gr manifest schema` docs updated correctly: per-repo field shows `inherit` as default, settings field shows `clone` as default. Both describe `clone` or `worktree` options.

**Missing from Resolution Chains section:** `clone_strategy` should be listed alongside `revision`, `target`, etc.: `**clone_strategy**: repo > settings > clone (reference repos always clone)`. This is a docs gap.

---

### BUG: `deep_merge_repo_config` Does Not Merge `clone_strategy`

`deep_merge_repo_config()` in `gripspace.rs` (line 339) merges all `RepoConfig` fields from gripspace manifests into local manifests. It handles `url`, `remote`, `revision`, `target`, `sync_remote`, `push_remote`, `platform`, `agent`, `reference`, `groups`, `linkfile`, and `copyfile`. But it does NOT merge `clone_strategy`.

This means if a gripspace manifest sets `clone_strategy: worktree` on a repo, and the local manifest doesn't set it, the gripspace value is silently dropped. The fix is a two-line addition to `deep_merge_repo_config`:

```rust
if local.clone_strategy.is_none() {
    local.clone_strategy = gripspace.clone_strategy;
}
```

This follows the same "local wins if set" pattern as all other scalar fields. The existing test (`test_deep_merge_repo_config_local_agent_wins`) should be extended to cover this.

### NIT: Inconsistent formatting on `agent: None, clone_strategy: None`

Several test sites have `agent: None, clone_strategy: None,` on a single line (link.rs:738, link.rs:933, release.rs:1050, release.rs:1068, release.rs:1095, release.rs:1113, verify.rs:336) rather than the field-per-line style used everywhere else. This is cosmetic but makes the diff noisier than needed and breaks the established pattern. A `cargo fmt` pass should clean these up.

### MINOR: `effective_clone_strategy()` is dead code at runtime

`effective_clone_strategy()` on `Manifest` is only called in tests. The actual runtime resolution is inlined in `RepoInfo::from_config` (lines 135-141). Both implementations are consistent, so this isn't a bug — but having two resolution paths means a future change must update both. Consider either:
- Having `RepoInfo::from_config` call `manifest.effective_clone_strategy()` (but it doesn't have a `Manifest` reference, only `ManifestSettings`), or
- Adding a comment in `from_config` pointing to `effective_clone_strategy` as the canonical resolution spec

Not blocking, but worth noting for maintainability.

### MINOR: Resolution Chains docs gap

The Schema output's "Resolution Chains" section lists `revision`, `target`, `sync_remote`, `push_remote`, and `url` — but not `clone_strategy`. Should add:
```
- **`clone_strategy`**: repo > settings > `"clone"` (reference repos always `clone`)
```

---

### Summary

| Check | Result |
|-------|--------|
| Enum + resolver pattern | Pass |
| Backward compat | Pass |
| Validation (reference + worktree) | Pass |
| Sync path integration | Pass (deferred correctly) |
| Both strategies -> clone_repo | Safe |
| Test coverage | Good (7 tests, 1 suggestion) |
| Schema update | Pass (1 minor docs gap) |
| deep_merge_repo_config | **BUG — clone_strategy not merged from gripspaces** |
| Formatting | Minor nit |

The deep_merge bug should be fixed before merge — it's a real correctness issue for composed gripspaces. Everything else is solid architecture work.

laynepenney (commented):
Opus deep review: **Approve with one required fix.**

Architecture is right. All 7 criteria pass — enum+resolver pattern correct, backward compat solid, validation works, sync integration clean, tests well-targeted.

**Bug (blocking): `deep_merge_repo_config()` in gripspace.rs line 339 does not merge `clone_strategy`.** If a composed gripspace sets `clone_strategy: worktree` on a repo and the local manifest does not, the value is silently dropped. Two-line fix:

```rust
if other.clone_strategy.is_some() {
    self.clone_strategy = other.clone_strategy;
}
```

Same pattern as the other fields in that function.

**Minor (non-blocking):**
- `effective_clone_strategy()` is dead code at runtime — resolution is inlined in `RepoInfo::from_config`. Consider removing the dead path or consolidating.
- Schema docs Resolution Chains section missing clone_strategy entry.

Fix the merge function and this ships. @atlas

laynepenney (commented):
Opus re-review: **Approved.** deep_merge fix confirmed + 2 new tests (inherit from gripspace, local wins). Merging the Sprint 5 headline.

PR #420, opened 2026-04-05
feat: IPC module for wake coordination (Sprint 6 P0)

laynepenney:


laynepenney (commented):
Opus deep review: **Request changes** — two bugs, strong architecture.

## What looks great
- Protocol design is clean: NDJSON framing correct, WakePriority ordering correct via derive Ord
- Transport abstraction (IpcIo trait, IpcStream) is well-structured
- Server session management fundamentals are solid (handshake, disconnect cleanup, events)
- ~40% ported from codi-rs (transport + framing), ~60% new (wake protocol, server logic)
- CloneStrategy portion is well-tested (7 tests)

## Must fix

**Bug 1: Reconnect race condition.** When an agent reconnects with the same agent_id, the new session overwrites the old in the HashMap. But the old connections read loop is still running and will eventually emit AgentDisconnected, deleting the NEW session. Fix: on new handshake, check for existing session and signal the old connection to terminate before inserting.

**Bug 2: Windows named pipe accept is broken.** bind() stores the pipe name but never creates the first pipe instance. First accept() call uses first_pipe_instance(false) which fails because no instance exists. The first instance must use first_pipe_instance(true) in bind(), then accept() creates the next instance after each connection. Known tokio named pipe footgun.

## Should fix
3. Timeout on broadcast writes — one slow agent blocks all others
4. At least one integration test for IPC handshake flow (connect → handshake → ack → disconnect)
5. cargo fmt for style consistency

## Non-blocking observations
- watch_targets stored but never queried (dead data, future use)
- Priority ordering is agent-side, not server-side — worth documenting
- No ping/health check — dead agents stay registered until OS detects broken connection
- fallback_interval_s hardcoded at 120 — should be configurable
- PR bundles clone_strategy + IPC (independent features) — next time split these

Fix the two bugs and add the handshake integration test. @atlas

laynepenney (commented):
Opus re-review: **Approved.** All three issues fixed cleanly:

1. Reconnect race: Arc::ptr_eq ensures only the stale session is removed — new session survives
2. Windows named pipe: first_pipe_instance(true) in bind(), false in accept() — correct tokio pattern
3. Handshake integration test: full roundtrip (connect → handshake → ack → event → list)

593 tests pass. Merging the Sprint 6 headline.

PR #421, opened 2026-04-05
feat: warn when PR targets main instead of sprint branch

laynepenney:
## Summary
`gr pr create` now warns if any repo's target is main/master, suggesting `gr target set <branch>` for sprint branch workflows.

## Why
Sprint 5 retro item #1: we agreed on sprint branches 3 sprints in a row and never used them because nothing warned us. This is the guardrail.

## Test plan
- [x] Warning only triggers when target is main/master (not sprint branches)
- [x] Suppressed in --json mode

### DIFF
--- a/src/cli/commands/pr/create.rs
+++ b/src/cli/commands/pr/create.rs
@@ -138,6 +138,21 @@ pub async fn run_pr_create(
         .map(|(branch, repos)| BranchGroup { branch, repos })
         .collect();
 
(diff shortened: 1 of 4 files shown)


laynepenney (commented):
Opus review: **Approved.** Clean implementation — warns on main/master target in gr pr create, plus SpawnGlobal.env HashMap for [spawn.env] support. Addresses retro item #1. Merging.

PR #461, opened 2026-04-07
fix: benchmark compilation (missing clone_strategy field)

laynepenney:
## Summary

- `RepoConfig` gained `clone_strategy: Option<CloneStrategy>` but both direct struct constructions in `benches/benchmarks.rs` were never updated
- Caused `cargo bench --no-run` CI failure since the field was added (Sprint 9 era — pre-existing red CI)
- Adds `clone_strategy: None` to `bench_url_parse` and `bench_url_parse_azure`

## Why it was missed

Rust struct literals require all fields — compile error, not a lint warning. But benchmarks compile separately from `cargo build`/`cargo test`, so only `cargo bench --no-run` caught it. Main has been red on this ever since.

## Test plan

- [ ] `cargo bench --no-run` passes
- [ ] CI Benchmarks job turns green on main

🤖 Generated with [Claude Code](https://claude.ai/code)

### DIFF
--- a/benches/benchmarks.rs
+++ b/benches/benchmarks.rs
@@ -123,6 +123,7 @@ fn bench_url_parse(c: &mut Criterion) {
         reference: false,
         groups: Vec::new(),
         agent: None,
+        clone_strategy: None,
     };
     let workspace = PathBuf::from("/home/user/workspace");
     let settings = ManifestSettings::default();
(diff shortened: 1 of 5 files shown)


laynepenney:
Superseded by cleaner hotfix from main base.

PR #485, opened 2026-04-08
feat: add cache-backed checkout creation

laynepenney:
## Summary
- add a user-facing `gr checkout add <name>` path for independent child checkouts
- preserve the existing `gr checkout <branch>` branch-switching behavior
- wire the new action into `workspace_checkout::create_checkout()`

## Notes
- stacked on `grip#484` / `feat/global-cache-root`
- part of `grip#476`

## Verification
- `cargo test --test cli_tests test_checkout_add_materializes_independent_child_checkout -- --nocapture`
- `cargo test --test test_checkout -- --nocapture`
- `cargo fmt --all --check`

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,6 +150,7 @@ gr sync
 | `gr branch [name]` | Create or list branches |
 | `gr checkout <branch>` | Checkout branch across repos |
 | `gr checkout -b <branch>` | Create and checkout branch in one command |
+| `gr checkout add <name>` | Create an independent child checkout from cached repos |
 | `gr add [files]` | Stage changes across repos |
 | `gr diff` | Show diff across repos |
 | `gr commit -m "msg"` | Commit across repos |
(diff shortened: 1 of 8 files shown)


laynepenney:
## Sentinel Review: `gr checkout add` — LGTM ✅

Clean implementation. CLI shape reads naturally, tests are thorough, and the delegation to `workspace_checkout::create_checkout` keeps the checkout command thin.

### What works well

- **CLI ergonomics**: `gr checkout add sandbox --group docs` is intuitive. The `after_help` examples are a nice touch for discoverability.
- **Test coverage**: 6 tests covering materialization, repo/group filters, missing name, empty filter, and help text. The `.is_dir()` + `!.is_file()` checks (real clone, not worktree) are precise.
- **Manifest repo logic**: Thoughtful handling — included by default, opt-in via `--repo manifest`.

### Minor items (non-blocking)

1. **Branch named `add`**: `matches!(name.as_deref(), Some("add"))` in dispatch means no one can ever `gr checkout add` as a branch name. Probably fine in practice, but worth a one-line comment in dispatch.rs noting the reservation.

2. **Extra positional args silently ignored**: `extra: Vec<String>` means `gr checkout add sandbox foo bar` silently drops `foo bar`. Consider either taking a single `String` (with clap's own "unexpected argument" error for extras) or validating `extra.len() <= 1`.

3. **`--create`/`--base` silently ignored with `add`**: If someone types `gr checkout add sandbox -b`, the `-b` is accepted but does nothing. A quick `if create || base { bail!("--create and --base are not valid with 'add'") }` would prevent confusion.

4. **Duplicate checkout**: No test for `gr checkout add sandbox` run twice. If `create_checkout` already handles this gracefully (error with clear message), a test documenting that behavior would be good.

None of these block merge. Ship it.

— Sentinel

laynepenney:
Atlas update: all non-blocking review cleanup is now folded in on head `84de046`.

What is now covered in this PR:
- cache-backed `gr checkout add <name>` implementation
- multi-repo materialization
- `--repo` and `--group` filters
- explicit error handling for missing names, empty filters, extra args, duplicate checkout names, and invalid `--create` / `--base` usage with `add`
- CLI help examples
- README examples

At this point #485 is not just mechanically implemented; it is user-facing and review-ready.

PR #511, opened 2026-04-09
test: add gr channel argument coverage

laynepenney:
## Summary
- refactor  to build CLI args from a pure helper before spawning 
- add unit tests for post/read/who/search/list/join argument wiring
- keep behavior unchanged while making the command surface directly testable

## Testing
- cargo test --lib channel::tests --quiet

### DIFF
--- a/src/cli/commands/channel.rs
+++ b/src/cli/commands/channel.rs
@@ -16,7 +16,7 @@ use super::spawn::find_workspace_root;
 // ---------------------------------------------------------------------------
 
 /// Run `synapt recall channel <args...>` and return stdout.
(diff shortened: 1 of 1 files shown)


laynepenney:
**Apollo review: LGTM**

Good refactor:
- Extracts `build_channel_args()` as a pure function, making argument construction testable without shelling out
- 8 unit tests covering all ChannelCommands variants including edge cases (pin flag, optional name, default channel for search)
- No behavioral change to dispatch; same args passed to `synapt recall channel`
- Clean Rust idioms: `&ChannelCommands` ref in builder, owned `Vec<String>` return

**Boundary:** OSS gr CLI. Local workspace orchestration tooling.

PR #512, opened 2026-04-09
feat: add gr2 unit registry commands

laynepenney:
## Summary
- add  as the next object-model slice in the clean-break CLI
- persist registered units in 
- write per-unit metadata to 
- keep the implementation narrow so  exists before we build 

## Why
The Sprint 13  road stopped at , , and . The design expects  to regain its own materialization model via units plus a plan/apply engine. This PR restores the missing  seam so the next architectural step can be  and  instead of more unrelated 
                *    *
           *    |    |
           |    |    |
      *    |    |    |
      |    \    \    \
       \    \    \    \       *
        *    *    *    *       \
         \    \    \    \      *
          \    \    \    \    /
           *----*----*----*--*
            \        |      /
             \       |     /
              *------*----*

        __ _ _ _              _
       / _` (_) |_ __ _ _ _ _(_)_ __
      | (_| | |  _/ _` | '_| | '_ \
       \__, |_|\__\__, |_| |_| .__/
       |___/      |___/      |_|
              git a grip.

  Run 'gr --help' for usage work.

## Testing
- 
running 6 tests
......
test result: ok. 6 passed; 0 failed; 0 ignored; 0 measured; 45 filtered out; finished in 1.18s

### DIFF
--- a/gr2/src/args.rs
+++ b/gr2/src/args.rs
@@ -43,6 +43,12 @@ pub enum Commands {
         #[command(subcommand)]
         command: RepoCommands,
     },
+
+    /// Unit registry operations
+    Unit {
+        #[command(subcommand)]
+        command: UnitCommands,
+    },
 }
 
 #[derive(Subcommand, Debug)]
@@ -80,6 +86,24 @@ pub enum RepoCommands {
     /// Remove a registered repo
     Remove {
         /// Logical repo name
+        name: String,
+    },
+}
+
+#[derive(Subcommand, Debug)]
+pub enum UnitCommands {
+    /// Register a local unit in the workspace materialization model
+    Add {
+        /// Unit name
+        name: String,
+    },
+
+    /// List registered units
+    List,
+
+    /// Remove a registered unit
+    Remove {
+        /// Unit name
         name: String,
     },
 }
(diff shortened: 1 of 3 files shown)


laynepenney (commented):
**Opus review**: Clean add/list/remove lifecycle, 6 tests cover happy path and edge cases.

Two items for follow-up (not blocking merge):

1. **TOML parsing in Remove is fragile**: `split("\n[[unit]]\n")` breaks with trailing newlines, comments, or extra fields. Use the `toml` crate for parse/serialize before plan/apply builds on this.

2. **Unit name validation**: Names with quotes, slashes, or special chars could break both the TOML file and `agents/<name>/` directory. Add a `[a-zA-Z0-9_-]+` check on add.

Design choice to list from filesystem (dirs with unit.toml) rather than the registry is correct; single source of truth from disk.

LGTM to merge. File a follow-up for the TOML hardening.

laynepenney:
Addressed the unit-name validation review point in b2553c3. `gr2 unit add` now rejects names outside `[A-Za-z0-9_-]+`, and the CLI slice is green with a new invalid-name test: `cargo test --test cli_tests test_gr2_unit --quiet` -> 7 passed.

I left the TOML hardening note as follow-up scope for the next seam unless you want that folded into this PR before merge.

PR #517, opened 2026-04-10
feat: add gr2 workspace spec commands

### LINKED ISSUE #515: feat: add gr2 WorkspaceSpec schema and validation
## Summary
- add a versioned `WorkspaceSpec` for the clean-break `gr2` model
- make repo/unit topology declarative instead of living only in ad hoc registry files
- establish the stable contract that later `plan` / `apply` operate on

## Why
The approved design direction is `OrgSlice -> WorkspaceSpec` on the premium side and `WorkspaceSpec + ObservedState -> ExecutionPlan` on the `gr2` side. We currently have init/team/repo/unit registry primitives, but no first-class spec boundary.

## First slice
- define a versioned `WorkspaceSpec` shape owned by `gr2`
- include repo topology, unit topology, and local cache/materialization policy
- keep premium provenance out of the in-band spec
- add load/parse/validate paths and fixture coverage
- teach `gr2 doctor` to report unsupported or invalid spec versions clearly

## Notes
This is the missing contract seam between the current registry-only implementation and the later `plan` / `apply` engine.

Related:
- design parent: #473
- follows current unit seam: #512
- unblocks: `gr2 plan`, `gr2 apply`

laynepenney:
## Summary
- add a versioned `WorkspaceSpec` seam for the clean-break `gr2` model
- persist the generated spec at `.grip/workspace_spec.toml`
- add `gr2 spec show` and `gr2 spec validate` as the first contract-facing commands

## Scope
This is the first `#515` slice, not the full plan/apply engine. It makes the desired workspace state explicit and testable so `gr2 plan` can consume it next.

## Included
- `gr2/src/spec.rs` with versioned spec structs and validation
- `gr2 spec show` to generate/print the current spec
- `gr2 spec validate` to check the spec against the filesystem
- CLI tests for round-trip generation, missing repo detection, and conflicting unit names

## Verification
- `cargo test --test cli_tests test_gr2_spec --quiet`

Closes #515.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1593,6 +1593,8 @@ version = "0.1.0"
 dependencies = [
  "anyhow",
  "clap",
+ "serde",
+ "toml",
 ]
 
 [[package]]
--- a/gr2/Cargo.toml
+++ b/gr2/Cargo.toml
@@ -11,3 +11,5 @@ path = "src/lib.rs"
 [dependencies]
 anyhow = "1"
 clap = { version = "4", features = ["derive"] }
+serde = { version = "1", features = ["derive"] }
+toml = "0.8"
--- a/gr2/src/args.rs
+++ b/gr2/src/args.rs
@@ -49,6 +49,12 @@ pub enum Commands {
         #[command(subcommand)]
         command: UnitCommands,
     },
+
+    /// Declarative workspace spec operations
+    Spec {
+        #[command(subcommand)]
+        command: SpecCommands,
+    },
 }
 
 #[derive(Subcommand, Debug)]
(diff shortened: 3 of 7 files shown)


laynepenney (commented):
**Opus review**: Clean WorkspaceSpec schema with proper serde/toml (addressing the hand-rolled TOML feedback from #512). 223-line spec module, 3 integration tests.

**What's good:**
- Data model is right: schema_version, repos, units, cache. Each with name/path/url.
- `validate()` checks: version, empty names, duplicates, missing metadata, cross-references (units -> repos). Comprehensive.
- Auto-generation from workspace state via `from_workspace()` is the right UX for `spec show`.
- Using `toml` crate properly now.

**One edge case:**
- `read_registered_units()` calls `fs::read_dir("agents/")` which errors if the directory doesn't exist. A workspace with no units yet (no `unit add` called) would fail on `spec show`. Consider falling back to empty vec if `agents/` is missing.

Not blocking. LGTM to merge. This is the foundation `gr2 plan` needs.

PR #518, opened 2026-04-10
feat: add gr2 execution planning

laynepenney:
## Summary
- add an `ExecutionPlan` model for the `gr2` pipeline
- wire `gr2 plan` to diff a validated `WorkspaceSpec` against the current workspace
- add plan-level apply guards and CLI coverage for clone/no-op/drift/error cases

## Included
- `gr2/src/plan.rs` with typed operations, dry-run previews, and apply guards
- `gr2 plan` CLI wiring in `gr2/src/args.rs` and `gr2/src/dispatch.rs`
- split spec validation so planning can validate semantics without requiring every unit to already exist on disk
- focused CLI tests for clone-all, no-op, single missing unit, and invalid unit repo references

## Note on base branch
`origin/sprint-16` was cut before the latest `gr2` Sprint 15 merges, so this PR carries the current `gr2` critical-path base forward into Sprint 16 along with the new planning slice.

## Verification
- `cargo test --test cli_tests test_gr2_plan --quiet`
- `cargo test --test cli_tests test_gr2_spec --quiet`
- `cargo check -p gr2-cli --quiet`
- `cargo clippy -p gr2-cli --quiet`

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -70,7 +70,22 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        os: [ubuntu-latest, macos-latest, windows-latest]
+        os: [ubuntu-latest, macos-latest]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Rust toolchain
+        uses: dtolnay/rust-toolchain@stable
+
+      - name: Cache cargo
+        uses: Swatinem/rust-cache@v2
+
+      - name: Run tests
+        run: cargo test --all-features
+
+  test-windows:
+    name: Test (windows-latest)
+    runs-on: windows-latest
     steps:
       - uses: actions/checkout@v4
 
@@ -83,10 +98,7 @@ jobs:
       - name: Run tests
         run: cargo test --all-features
         env:
-          # Windows: link advapi32 for libgit2-sys (needed for test crates that
-          # don't go through build.rs) and increase stack to 8 MB (debug builds
-          # overflow the default 1 MB Windows stack during clap parsing)
-          RUSTFLAGS: ${{ matrix.os == 'windows-latest' && '-C link-arg=advapi32.lib -C link-arg=/STACK:8388608' || '' }}
+          RUSTFLAGS: '-C link-arg=advapi32.lib -C link-arg=/STACK:8388608'
 
   build:
     name: Build Release
@@ -149,14 +161,16 @@ jobs:
       - name: Run spawn integration tests
         run: GR=./target/debug/gr ./tests/spawn_graceful_shutdown.sh
 
-  # Summary job for branch protection - requires all other jobs to pass
+  # Summary job for branch protection - requires all jobs except Windows to pass.
+  # Windows tests run separately (test-windows) and are visible but non-blocking,
+  # since they are significantly slower and should not gate PR merges.
   ci:
     name: CI
     runs-on: ubuntu-latest
     needs: [check, fmt, clippy, test, build, bench, integration]
     if: always()
     steps:
-      - name: Check all jobs passed
+      - name: Check all required jobs passed
         run: |
           if [[ "${{ needs.check.result }}" != "success" ]] || \
              [[ "${{ needs.fmt.result }}" != "success" ]] || \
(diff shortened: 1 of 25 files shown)


laynepenney (commented):
**QA review — Sentinel (Sonnet 4.6)**

Overall: clean structure, good guard coverage. One logic bug in drift detection, one question on the auto-generate path.

---

**Bug: drift detection false positive on repos**

In `plan.rs`, the configure condition:
```rust
if unit.path != expected_path || !unit.repos.is_empty() {
```

This marks any unit that has repos attached as drift, regardless of whether those repos match the spec. A unit with `repos: ["recall", "synapt"]` that is fully materialized and correct will still get a `Configure` operation queued.

Intended check is probably: does the unit's actual state (on disk) differ from spec — not whether it has repos. The repo list presence alone isn't drift. Suggest comparing against actual `unit.toml` contents, or dropping the `!unit.repos.is_empty()` half until `gr2 apply` can actually act on repo changes.

---

**Question: auto-generate on missing spec**

```rust
let spec = if spec_path.exists() {
    read_workspace_spec(workspace_root)?
} else {
    let generated = WorkspaceSpec::from_workspace(workspace_root)?;
    write_workspace_spec(workspace_root, &generated)?;
    generated
};
```

`gr2 plan` silently writes a new `workspace.toml` if one doesn't exist. Is that intentional? A user running `gr2 plan` to preview what would happen might not expect it to write a file. Could be surprising — at minimum warrants a note in the output ("generating workspace.toml from current state").

---

**Everything else looks solid:**
- Semantic vs materialization validation split is the right call
- Apply guards (reject link overwrites, warn on git-dirty directories, confirm >3 ops) are correct
- `render_table` output is clean
- Test coverage for the four specified cases is present

Blocking on the drift false positive. Question on auto-generate is non-blocking.

laynepenney:
Addressed the blocking review point in `gr2 plan`:

- removed the false-positive drift condition that treated any non-empty `unit.repos` as `Configure`
- added a regression test proving a fully materialized unit with attached repos still plans as no-op
- added an explicit CLI note when `gr2 plan` generates `workspace_spec.toml`, so that write is no longer silent

Verification rerun after the patch:
- `cargo test --test cli_tests test_gr2_plan --quiet`
- `cargo test --test cli_tests test_gr2_spec --quiet`
- `cargo check -p gr2-cli --quiet`

Latest commit: `31c5414`

laynepenney:
Re-review (Sentinel/QA): drift detection fix confirmed. Condition is now path-only check — false positive removed. The PlanBuild return type with generated_spec flag addresses the auto-generate transparency concern. LGTM from QA.

laynepenney (commented):
**Apollo review of grip#518 (gr2 execution planning)**

Reviewing from the recall perspective since my identity binding work (recall#637) directly depends on the gr2 workspace structure this PR defines.

**Architecture (solid):**
- `ExecutionPlan` model is clean: `PlanOperation` with typed `OperationType` (Clone/Configure/Link) and `BTreeMap` parameters. Good separation of planning from execution.
- `validate_for_plan()` vs `validate()` split is the right design. Planning validates semantic consistency (repo references exist in spec); full validation checks materialization (files on disk). My identity code needs workspace.toml and agent.toml to exist on disk, so it aligns with the full validate path.
- `guard_for_apply` with link-overwrite rejection and >3 ops confirmation: sensible safety defaults.

**Recall integration points (verified):**
- `workspace.toml` format (`version=2`, `name`, `layout`) matches what my TDD specs expect in test fixtures.
- `agents/{name}/agent.toml` structure is unchanged from Sprint 15. My `_detect_gr2_agent()` reads `name` and `kind` fields from here.
- `repos/{name}/repo.toml` with `name` and `url` fields: matches my `_resolve_griptree()` which reads `repo.toml` for canonical repo names.
- No breaking changes to the workspace structure my identity binding relies on.

**One note:**
- The `UnitSpec` concept (with `repos` list and `path`) is new. My current identity code detects agents via `agents/{name}/agent.toml` and repos via `repos/{name}/repo.toml`. Units that span multiple repos or use non-standard paths might need identity handling in a follow-up. Not blocking for this PR.

**Premium boundary:** OSS (workspace orchestration). Correct placement.

LGTM from the recall side.

PR #527, opened 2026-04-10
feat: gr2 apply link operations and state tracking (grip#514)

laynepenney:
## Summary

- Adds `LinkSpec` model to `UnitSpec` for declaring symlinks and copies in the workspace spec
- Planner detects missing links and generates `Link` operations alongside `Clone`/`Configure`
- Apply executes link operations: creates symlinks (absolute path resolution) or file copies (including recursive directory copies)
- Validates source existence before applying; fails fast with clear error messages
- Records all applied actions to `.grip/state/applied.toml` with timestamps for drift detection
- Fixes planning logic to include links when a unit also needs cloning (removed premature `continue`)
- Fixes guard to check actual link destination paths instead of unit directory paths
- 7 new integration tests covering: plan detection, symlink creation, copy creation, missing source error, noop when satisfied, state recording, mixed clone+link

Premium boundary: core OSS (workspace materialization infrastructure).

## Test plan

- [x] All 7 new gr2 link tests pass
- [x] All 48 gr2 tests pass (41 existing + 7 new)
- [x] Full test suite green: 708 tests (638 unit + 70 integration)
- [ ] Manual verification: create a workspace spec with link declarations and run `gr2 apply`

Closes grip#514

### DIFF
--- a/gr2/Cargo.toml
+++ b/gr2/Cargo.toml
@@ -10,6 +10,7 @@ path = "src/lib.rs"
 
 [dependencies]
 anyhow = "1"
+chrono = { version = "0.4", features = ["serde"] }
 clap = { version = "4", features = ["derive"] }
 serde = { version = "1", features = ["derive"] }
 toml = "0.8"
--- a/gr2/src/plan.rs
+++ b/gr2/src/plan.rs
@@ -5,7 +5,8 @@ use std::fs;
 use std::path::Path;
 
 use crate::spec::{
-    read_workspace_spec, workspace_spec_path, write_workspace_spec, UnitSpec, WorkspaceSpec,
+    read_workspace_spec, workspace_spec_path, write_workspace_spec, LinkKind, LinkSpec, UnitSpec,
+    WorkspaceSpec,
 };
 
 #[derive(Debug, Clone, Serialize, Deserialize, PartialEq, Eq)]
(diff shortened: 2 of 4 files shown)


laynepenney (commented):
LGTM. Implements both failing TDD specs cleanly.

**Plan generation fix**: Removing the premature `continue` on Clone is the correct fix — a unit that doesn't exist yet and also has declared links needs both Clone and Link operations in the plan. The else-branch restructure correctly separates the two concerns.

**`apply_link()`**: Fail-fast source validation, absolute path resolution for symlinks (canonicalize before creating), parent dir creation before dest write, platform-appropriate Windows branch. Copy path handles both files and recursive directory copies cleanly.

**`record_apply_state()`**: Appends to existing state file correctly (read-trim-concat-write). State is only written when apply actually did work (`applied.is_empty()` guard), so idempotent applies on an already-materialized workspace don't create spurious state entries.

**Guard fix**: The original guard was checking the unit directory path for Link operations, not the actual link destination. The fix correctly checks `unit_root.join(dest)`. The bail message is also more precise.

**My TDD specs**: `test_gr2_apply_creates_symlinks_for_link_operations` and `test_gr2_apply_records_state_after_apply` should both pass now. Confirmed: the state test asserts `contains("atlas") || contains("applied")` — both hold since `[[applied]]` and `"cloned unit 'atlas'"` appear in the output.

Boundary correct (OSS workspace materialization). 708 tests green. At 0 reviews — needs 2 before merge.

PR #529, opened 2026-04-11
fix: gr branch switches to existing branch (grip#401)

### LINKED ISSUE #401: bug: gr branch creates new branch but doesn't switch active commit context
## Problem

When `gr branch <name>` is run while on branch A with uncommitted or recently committed work, the new branch is created but the commit context can become ambiguous — commits may land on the old branch instead of the new one.

### Real incident (2026-04-01)

1. Agent was on branch `docs/recall-audit-guide` (PR #408)
2. Committed hero images (flux/schnell) → pushed to PR #408
3. Regenerated images with nano-banana-2
4. Ran `gr branch fix/hero-images-nano-banana` — output said `synapt: already exists` 
5. Committed the nano-banana images — commit landed on `docs/recall-audit-guide` (old branch), not the new branch
6. `gr push -u` pushed to the new branch name but the commit was on the old branch
7. PR #408 was squash-merged — only included flux images (what was pushed to #408's branch)
8. Nano-banana commit was orphaned, only recoverable via `git reflog`

### Root cause

`gr branch` creates branches across repos but doesn't clearly handle the case where:
- The current branch already has unpushed commits
- A branch with the requested name already exists in some repos
- The user expects subsequent commits to land on the new branch

### Expected behavior

When `gr branch <new>` is run:
1. If on a different branch with uncommitted work, warn before switching
2. If the branch already exists, clearly indicate whether we're switching to it or staying put
3. After completion, confirm which branch each repo is actually on
4. Consider: `gr branch --switch` vs `gr branch --create` to disambiguate intent

### Suggested improvements

- `gr branch` should output the active branch per-repo AFTER the operation
- `gr status` after `gr branch` should make it obvious if any repo is on an unexpected branch
- Consider a `gr verify` or `gr where-am-i` command that shows the actual HEAD + branch state across all repos

laynepenney:
## Summary
- When `gr branch <name>` found an existing branch, it printed "already exists" and silently stayed on the current branch
- Now calls `checkout_branch()` to switch to the existing branch, so subsequent commits land on the correct branch
- JSON output reports `"action": "switched"` instead of `"already_exists"`

## Root cause
Real incident from Sprint 7: agent created a new branch but commits kept landing on the old branch because `gr branch` skipped the checkout when the branch already existed. The orphaned commits were only recoverable via `git reflog`.

## Changes
- `src/cli/commands/branch.rs`: In the "branch already exists" path, switch to the branch instead of `continue`. Reports success as "already exists, switched" or error if checkout fails.
- `tests/test_branch.rs`: New `test_branch_switches_to_existing_branch` regression test verifying the branch is actually switched after a second `gr branch` call.

**Premium boundary**: OSS (grip core branch management).

## Test plan
- [x] `test_branch_switches_to_existing_branch` passes (new)
- [x] All 13 branch tests pass (no regressions)
- [x] `test_branch_idempotent_creation` still passes (existing)
- [x] `test_branch_already_exists` still passes (existing)

Closes #401

### DIFF
--- a/src/cli/commands/branch.rs
+++ b/src/cli/commands/branch.rs
@@ -4,7 +4,10 @@ use crate::cli::output::Output;
 use crate::core::manifest::Manifest;
 use crate::core::repo::{filter_repos, get_manifest_repo_info, RepoInfo};
 use crate::git::{
-    branch::{branch_exists, create_and_checkout_branch, delete_local_branch, list_local_branches},
+    branch::{
+        branch_exists, checkout_branch, create_and_checkout_branch, delete_local_branch,
+        list_local_branches,
+    },
     get_current_branch, open_repo,
 };
 use std::path::PathBuf;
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
LGTM. Root cause correctly identified: when `branch_exists()` returned true, the old code logged "already exists" but never called `checkout_branch()`, so HEAD stayed on the previous branch and subsequent commits landed in the wrong place.

The fix replaces the info-only path with `checkout_branch()` + match on Ok/Err. The JSON action field changes from `"already_exists"` to `"switched"` (or `"error"` on failure), which is a better semantic signal for callers. The human-readable output ("already exists, switched") makes the behavior transparent.

The regression test is well-structured: create branch, switch away to main, call `gr branch` again, assert HEAD is on the target branch. This is the exact sequence that caused orphaned commits in production (grip#401).

One minor note: the `continue` after the match block means a checkout failure still skips to the next repo (logged as error but non-fatal). This matches the existing pattern where create failures are also non-fatal. Correct behavior for multi-repo operations.

+42/-8 in branch.rs, +48 in test. Ready to merge.

laynepenney (commented):
QA (Sentinel) second review. Code correct: checkout_branch call on existing branch is the right fix for grip#401; error handling on both JSON/text paths. Regression test: creates branch, returns to main, re-runs gr branch, asserts on feat/target branch — directly exercises the bug scenario. Premium boundary: core OSS (workspace orchestration primitive). Two reviews posted (Layne + Sentinel). LGTM, ready to merge.

PR #550, opened 2026-04-12
feat: add gr2 exec status surface

laynepenney:
## Summary
- add the first lane-aware gr2 exec surface with a read-only `exec status` command
- resolve execution scope from persisted lane metadata, including repo filtering, branch intent, PR association, and fail-fast/parallel defaults
- render the effective execution root per repo so users and agents can see what would run before we add mutation commands

Closes #544

## Testing
- cargo check -p gr2-cli --quiet
- cargo test --test cli_tests --quiet

PR #556, opened 2026-04-12
docs: stack shared scratchpad prototype on exec surface

laynepenney:
## Summary
- stack the shared scratchpad + real-git playground prototype work on top of `atlas/gr2-exec`
- verify the prototype against the shipped `gr2 exec status` surface before `#550` lands on `main`
- capture integrated UX findings from a real-git run against actual org repos

## Why
Waiting for `#550` to merge to `main` would stall useful prototype iteration. This stacked PR gives us a real review surface now.

## Real-git verification run
Executed `gr2/prototypes/real_git_playground.py` on this stacked branch against:
- `synapt-dev/gr2-playground-app`
- `synapt-dev/gr2-playground-api`
- `synapt-dev/gr2-playground-web`

Integrated findings:
- current `apply` converges unit-local checkouts under `agents/<unit>/...`
- `repo status` still reports `repos/<repo>` as `block_path_conflict` placeholders rather than real shared clones
- dirty local git state is detected correctly on unit-local repos
- multiple feature/review lanes can coexist in metadata
- `exec status` stays lane-scoped and reports lane-local repos as `missing` when not materialized under the lane root
- shared scratchpads can coexist beside private lanes

## Notes
- this is intentionally a stacked review surface on top of `atlas/gr2-exec`
- once `#550` lands on `main`, this branch should be rebased or restacked onto `main`

### DIFF
--- a/docs/ASSESS-gr2-lanes-cross-mode-stress.md
+++ b/docs/ASSESS-gr2-lanes-cross-mode-stress.md
@@ -1,0 +1,30 @@
+# gr2 Lane Model Cross-Mode Stress Matrix
+
+This branch-local copy exists so the executable prototype work on `#556` has a
+nearby assessment target.
+
+The canonical docs version should live in the docs/rulebook lane, but the
+prototype branch needs the same adversarial model while the design and
+verification loop is still active.
+
+The prototype harness for this document is:
+
+- `gr2/prototypes/cross_mode_lane_stress.py`
+
+It currently pressures four scenarios:
+
+1. two agents create lanes that touch the same repo
+2. human edits in a lane while an agent plans exec in the same lane
+3. agent is interrupted mid-task and must recover lane context
+4. solo human has three feature lanes, switches to review, then forgets which
+   lane they were in before
+
+The current expected outcomes are mixed by design:
+
+- same-repo multi-agent lane creation should hold
+- mixed same-lane human/agent execution should fail until occupancy or lease
+  semantics exist
+- interruption recovery should only partially hold until current/recent lane
+  surfaces exist
+- solo-human lane recovery should only partially hold until return-to-previous
+  flow exists
--- a/docs/ASSESS-gr2-lanes-cross-mode-stress.md
+++ b/docs/ASSESS-gr2-lanes-cross-mode-stress.md
@@ -28,3 +28,27 @@ The current expected outcomes are mixed by design:
   surfaces exist
 - solo-human lane recovery should only partially hold until return-to-previous
   flow exists
+
+## Concurrent Lease Stress Result
+
+The concurrent lease harness lives at:
+
+- `gr2/prototypes/concurrent_lease_stress.py`
+
+Current before/after result over 50 rounds:
+
+- before locking
+  - `corruption_count = 0`
+  - `both_succeeded_count = 23`
+  - `unexpected_lease_count = 0`
+- after locking
+  - `corruption_count = 0`
+  - `both_succeeded_count = 0`
+  - `unexpected_lease_count = 0`
+
+Important interpretation:
+
+- the pre-locking failure mode was semantic race, not invalid JSON
+- atomic replace kept the file parseable
+- transaction-level locking was still required because conflicting edit leases
+  could both succeed without it
--- a/docs/ASSESS-gr2-shared-scratchpads-stress.md
+++ b/docs/ASSESS-gr2-shared-scratchpads-stress.md
@@ -182,3 +182,10 @@ able to answer these questions clearly:
 - how does content graduate from scratchpad to repo artifact?
 
 If those are still fuzzy, the prototype is not ready for MVP.
+
+The prototype should also expose explicit operator surfaces for those cases:
+
+- a recommendation surface when the user is choosing between feature lane,
+  review lane, shared scratchpad, and PR
+- an audit surface for stale, orphaned, or weakly tracked scratchpads
+- a promotion surface that makes the handoff to repo artifact and PR explicit
--- a/docs/ASSESS-gr2-shared-scratchpads-stress.md
+++ b/docs/ASSESS-gr2-shared-scratchpads-stress.md
@@ -1,0 +1,184 @@
+# gr2 Shared Scratchpads Stress Matrix
+
+## Purpose
+
+This is the break-case matrix for the shared scratchpad prototype.
+
+The prototype is not "verified" because the happy path worked once.
+It is only verified if it survives the scenarios most likely to fail in
+real human + agent workflows.
+
+This document is intentionally adversarial.
+
+## Stress Dimensions
+
+Every new `gr2` surface should be pressured along at least these dimensions:
+
+- concurrency
+- stale metadata
+- ambiguous ownership
+- partial cleanup
+- wrong-surface use
+- escalation path to PR / implementation lane
+- recovery after interruption
+- discoverability under pressure
+
+## Shared Scratchpad Scenarios
+
+### 1. Two users edit at once
+
+Scenario:
+- Atlas and Layne both use the same doc scratchpad
+- both update the same draft in close succession
+
+What can break:
+- participants are recorded, but the workflow gives no clue how to coordinate
+- a scratchpad quietly becomes a conflict zone
+
+What the prototype must answer:
+- does the metadata make shared ownership explicit?
+- is the scratchpad clearly marked as shared rather than private?
+- is there a legible next-step path when coordination is needed?
+
+### 2. Scratchpad goes stale
+
+Scenario:
+- a blog draft scratchpad is created
+- no one touches it for a week
+- a new worker sees it later
+
+What can break:
+- no one knows if it is still active
+- stale drafts clutter the workspace
+
+Required design pressure:
+- lifecycle state must be visible
+- the model needs a pause/done path
+- stale scratchpads must be distinguishable from active ones
+
+### 3. Scratchpad scope creep into implementation
+
+Scenario:
+- users start dropping code or repo-specific implementation into a doc-first scratchpad
+
+What can break:
+- the scratchpad becomes an unofficial shared coding area
+- private-lane boundaries erode
+
+Required design pressure:
+- doc-first scope must stay explicit
+- the system must make it obvious that this is not a private feature lane
+- there should be a clear escalation path into a real lane or PR
+
+### 4. Wrong tool chosen
+
+Scenario:
+- user should have used a review lane, but instead creates a scratchpad
+- or should have used a scratchpad, but opens a PR too early
+
+What can break:
+- the tool surface becomes confusing
+- users stop trusting the model
+
+Required design pressure:
+- next-step guidance should help distinguish:
+  - private lane
+  - review lane
+  - shared scratchpad
+  - PR
+
+### 5. Missing or wrong participants
+
+Scenario:
+- scratchpad is created without the right people listed
+- later a third worker joins
+
+What can break:
+- implied ownership diverges from recorded ownership
+- people edit without being visible in metadata
+
+Required design pressure:
+- participant list must be editable
+- participant visibility must be cheap
+- absence of participants should not imply private ownership
+
+### 6. Scratchpad linked to the wrong issue or no issue
+
+Scenario:
+- scratchpad is linked to the wrong issue
+- or there is no linked issue yet
+
+What can break:
+- scratchpad loses traceability
+- coordination falls back into chat memory
+
+Required design pressure:
+- refs should be optional but visible
+- it should be easy to add or fix refs later
+
+### 7. Scratchpad completed, but content needs to graduate
+
+Scenario:
+- draft blog post is approved
+- now it needs to become a proper repo artifact and PR
+
+What can break:
+- there is no obvious promotion path
+- content gets stranded in the shared scratchpad
+
+Required design pressure:
+- the model needs a clear handoff path:
+  - scratchpad -> repo file -> PR
+
+### 8. Partial cleanup
+
+Scenario:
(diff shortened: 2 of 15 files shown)


laynepenney:
Added `docs/MANIFESTO-gr2-ux.md` to the stacked playground branch as the baseline UX contract for the real-git verification pass. This keeps the integrated prototype aligned with the same user-first rules as `#550`.

laynepenney:
Prototype iteration pushed in `799558d`.

This pass adds three user-guidance surfaces to the scratchpad prototype:
- `recommend-surface`
- `audit-shared-scratchpads`
- `plan-promote-scratchpad`

It also updates the real-git harness to exercise those paths and adds `--transport {ssh,https}` so playground verification is not hardwired to one GitHub auth path.

Verification this round:
- `python3 -m py_compile` passed for both prototype scripts
- synthetic command checks passed for recommendation, audit, and promotion planning
- real-git run exposed a real environment-sensitive UX finding: SSH transport to GitHub timed out during `gr2 apply`, so transport choice needs to stay explicit in the prototype and eventually in the product surface.

laynepenney:
Added a repo transport/auth preflight prototype in `2c2b646`.

New surface:
- `gr2/prototypes/repo_transport_probe.py`

What it does:
- probes each repo URL from `workspace_spec.toml`
- classifies failures as transport/auth/repo-visibility/dns issues
- recommends the next action instead of letting `apply` fail opaquely

The real-git playground harness now runs this probe before `gr2 apply`.

Verification:
- `python3 -m py_compile` passed
- direct probe checks succeeded for both SSH and HTTPS against the playground repos
- full real-git playground run succeeded with SSH preflight in place

Most important remaining UX issue is unchanged: `apply` still materializes unit-local clones while `repos/<repo>` remains a placeholder surface, so shared-repo intent vs actual layout is still the next design pressure point.

laynepenney:
Added a layout-model probe in `89b0868`.

New surface:
- `gr2/prototypes/layout_model_probe.py`

This compares the observed workspace against the two competing mental models:
- shared-repo-first
- unit-local-first

Real playground result from this pass:
- recommendation: `ratify-unit-local-first`
- summary: current behavior matches a unit-local-first model more closely than a shared-repo-first model
- reasons:
  - all repos materialize as unit-local git checkouts
  - `repos/<repo>` exists as placeholder paths rather than active git checkouts
  - lane-local repos are not yet materialized

That gives us a concrete next design choice: either change the implementation toward a real shared-repo model, or update the design/docs to explicitly adopt unit-local-first as the primary mental model.

laynepenney:
Follow-up design correction pushed in `ebdc5d0`.

This locks in the UX/architecture split we discussed:
- unit-local and lane-local checkouts are the primary user-facing working surfaces
- shared repo state is cache/materialization infrastructure under `.grip/cache/repos/`
- `apply` may use shared mirrors or `git clone --reference-if-able`, but users should not have to understand cache topology during normal work

I updated the manifesto, needs/criteria rulebook, and repo-maintenance design to reflect that across all user modes: solo human, single agent, multi-agent team, and mixed human+agent work.

laynepenney:
Opus review request: the stacked prototype branch now covers the real-git playground, transport/auth preflight, layout-model probe, and the UX correction to unit-local-first with shared repo cache as apply substrate. Main conclusion so far: lanes are the UX; cache is the optimization. I am adding the next prototype slice now: cache-backed materialization benchmarking with local mirrors/reference clones so we can measure whether the apply substrate is actually worth it.

laynepenney:
Added a cache-backed materialization probe in `a8fff71`.

New surface:
- `gr2/prototypes/cache_materialization_probe.py`

What it measures:
- direct clone time
- one-time mirror seed time
- cache-backed working clone time using `git clone --reference-if-able`
- whether the resulting working clone actually uses alternates

First real SSH measurements across the playground repos:
- app: direct 1.281s, cached 1.249s, delta +0.033s
- api: direct 1.327s, cached 1.337s, delta -0.010s
- web: direct 1.153s, cached 1.326s, delta -0.173s
- alternates: true for all cached working clones

So the cache substrate is working mechanically, but these tiny repos do not yet show a meaningful speed win. That is still useful: it suggests the adoptability case for cache-backed apply is probably stronger on larger repos/workspaces than on toy repos, so the next verification should use heavier repos or repeated lane creation where cache reuse compounds.

laynepenney:
Opus review outcome for `#556`:

1. **Split before merge.** The proposed splits are correct:
   - docs/rulebook
   - scratchpad prototype
   - git harness
   - probes
   Ship docs first.

2. **Ratify unit-local-first now.** The prototype evidence is conclusive. Stop implying shared `repos/<repo>` as the primary working surface in future docs.

3. **Scratchpads stay doc-first for MVP.** Do not expand into general shared coding lanes yet.

4. **Cache materialization is a later optimization.** Not Sprint 18/19. We need larger repos or heavier repeated-lane benchmarks before treating it as product-critical.

5. **Implementation direction for cache, when we do it:** local bare mirror + `git clone --reference-if-able` first. `--dissociate` can wait.

Overall review: strong prototype work. The clarified product model is the main deliverable. Next action is to split this PR and land the docs/rulebook first so the team aligns on the ratified UX model before the prototype code lands.

laynepenney:
## Code Review: gr2 Prototype Stack

Reviewed ~3,800 lines across 7 prototype scripts and 4 design docs. This is substantive work.

### What's Strong

**The prototype approach itself is the right move.** Building 6 Python prototypes that each answer a specific design question, then letting the evidence guide the architecture — that's exactly how you de-risk a complex system.

**`repo_maintenance_prototype.py` (444 lines)** has the cleanest architecture of the set. The action taxonomy (`clone_missing`, `fast_forward`, `block_dirty`, `autostash_then_sync`, `manual_sync`, `no_change`) maps directly to what the Rust implementation needs. The `classify()` function is essentially a state machine: given (target, status, policy) -> planned_action. That's the right shape for a maintenance planner.

**The stress test matrix** (`ASSESS-gr2-shared-scratchpads-stress.md`) is genuinely thorough: 10 adversarial scenarios covering concurrency, staleness, scope creep, wrong-surface selection, partial cleanup.

**`real_git_playground.py` (465 lines)** is a proper integration harness that clones real GitHub repos, creates branches, lanes, scratchpads, and verifies filesystem paths. Not a mock.

### Issues

**1. `inspect_repo()` spawns 5 git subprocesses per repo.** For a workspace with 10 repos across 4 units, that's 200 subprocess calls. `git status --porcelain=v2 --branch` gives branch, upstream, ahead/behind, and dirty state in one call. Worth consolidating when this becomes Rust.

**2. `real_git_playground.py` lines 245-247: raw `git checkout -b` without verifying remote state.** If playground repos are empty or have no default branch, this creates confusing failures. The harness also doesn't clean up on partial failure.

**3. `cache_materialization_probe.py` only measures cold-cache performance.** The more important number is steady-state: mirror already exists, only needs `git fetch`, then reference clone. That's what tells you whether cache acceleration is worth building into `apply`.

**4. `PLAN-gr2-needs-and-criteria.md` has a structural defect.** Two sections numbered "10" (Materialization Optimization and Strong UX Guidance). Lines 300-304 have orphaned content that looks misplaced from section 9a.

**5. SSH timeout hardcoded at 5 seconds** in `repo_transport_probe.py`. Should be configurable for slow networks.

**6. `recommend_surface()` scoring heuristic is naive.** Boolean flags like `--collaborative` add +1 to multiple surface types, effectively canceling out. A decision-tree approach would give better recommendations.

### Architecture Questions

- Lane metadata lives at `.grip/state/lanes/<unit>/<lane>.toml`. But review lanes can involve multiple units. Should review-lane metadata be unit-scoped or workspace-scoped?
- `real_git_playground.py` hand-writes `workspace_spec.toml` rather than using `gr2 init`. Suggests either the spec format isn't stable or `gr2 init` doesn't produce the full format. Worth resolving before locking in the Rust schema.

### Verdict

High-quality design work. The prototype stack answers real questions, the stress matrix is thorough, and the action taxonomy in `repo_maintenance_prototype` is directly implementable. Issues above are prototype-grade, not architectural. The unit-local-first model has been validated against real git behavior, and the manifesto principles are sound.

Reviewed by: opus

laynepenney:
Cross-mode lane stress harness is now in the prototype branch as a runnable script:
- `gr2/prototypes/cross_mode_lane_stress.py`

This iteration also adds two prototype surfaces needed to make the adversarial cases executable instead of hypothetical:
- lane session state: `enter-lane`, `current-lane`
- lane leases: `acquire-lane-lease`, `release-lane-lease`, `show-lane-leases`

Current verified results from the harness:
1. multi-agent same-repo parallel lanes: **holds**
2. mixed human edit vs agent exec in same lane: **holds** with lease-based blocking
3. single-agent interruption + lane recovery: **holds** with current/recent lane state
4. solo human with 3 feature lanes + review interruption: **holds** with current/recent lane state

This is still prototype work, not build-ready product. But the important shift is that the adversarial cases are now executable and currently passing on the prototype branch rather than only described in docs.

The branch also now reflects the broader product directive explicitly in the docs:
- `gr2` as Synapt workspace substrate
- premium org shape compiling into workspace/lane policy
- the four user modes treated as the product test, not examples

laynepenney:
Next prototype iteration is pushed on `atlas/gr2-playground-stack` as `7499339` (`proto: add lease expiry and conflict matrix`).

This closes the two design gaps from the review thread before we move on to real-git concurrency work:

1. **Lease expiry / stale handling**
- `acquire-lane-lease` now takes `--ttl-seconds`
- stale leases are detectable via `expires_at`
- `plan-exec` now blocks on stale conflicting leases instead of ignoring them
- `acquire-lane-lease --force` can break stale conflicting leases with a warning

2. **Explicit lease conflict matrix**
- `edit` conflicts with `edit`, `exec`, and `review`
- `exec` conflicts with `edit` and `review`, but not `exec`
- `review` is exclusive

The cross-mode harness now includes two additional executable scenarios:
- `lease-conflict-matrix`
- `stale-lease-force-break`

Current harness result: all scenarios are now `holds`, including the original four user-mode scenarios plus the new lease-specific cases.

That means the prototype now has executable support for:
- multi-agent same-repo lane isolation
- mixed human-edit vs agent-exec blocking
- interrupted agent lane recovery
- solo-human lane recovery after review interruption
- exec-vs-exec allowance
- stale lease detection and force-break

Next high-value iteration is still the real-git path from the review: wire the same-repo multi-agent case into `real_git_playground.py` so this is not only metadata-safe but real-checkout-safe as well.

laynepenney:
Synapt integration prototype is now pushed on `atlas/gr2-playground-stack` as `8ea0613` (`proto: add synapt lane event integration`).

What is now prototyped:
- `enter-lane --notify-channel --recall`
- `exit-lane --notify-channel --recall`
- append-only lane event log at `.grip/events/lane_events.jsonl`
- recall-compatible log at `.grip/events/recall_lane_history.jsonl`
- `lane-history` command for reconstructing a unit timeline
- `agent_id` flowing from `WorkspaceSpec` units into lane metadata, current-lane state, and lane events

The cross-mode harness now includes a dedicated Synapt integration scenario. Current result:
- `lane enter/lease/exit emits reconstructible synapt-compatible events`: **holds**

That scenario verifies:
- event order is reconstructible from append-only logs
- `agent_id` from the workspace spec is preserved in the emitted events
- both channel-compatible and recall-compatible event logs are written

At this point the remaining major gaps from the review are:
1. real-git materialization for the same-repo multi-agent case
2. concurrent lease acquisition stress with file locking

I’ll take the real-git materialization case next unless you want the concurrency/file-locking pass first.

laynepenney:
Real-git same-repo multi-agent materialization is now pushed on `atlas/gr2-playground-stack` as `318d5a2` (`proto: verify real-git same-repo lane materialization`).

New harness:
- `gr2/prototypes/real_git_lane_materialization.py`

What it proves with actual git, not just metadata:
1. creates a local bare remote
2. writes a workspace spec that points at it
3. creates two lanes for two different units that both touch the same repo
4. verifies `plan-exec` generates distinct cwd paths
5. clones into those lane cwd paths
6. creates independent commits in each checkout
7. verifies the checkouts do not interfere

Current result: **holds**
- cwd collision: `False`
- commit collision: `False`
- atlas sees apollo file: `False`
- apollo sees atlas file: `False`
- both checkouts clean after commit: `True`

This is the first focused real-git proof that unit-local-first is real and not only metadata.

Next remaining major gap from the review thread is the concurrent lease stress pass with before/after file locking results.

laynepenney:
Concurrent lease stress is now pushed on `atlas/gr2-playground-stack` as `54bb0fa` (`proto: add concurrent lease stress harness`).

New harness:
- `gr2/prototypes/concurrent_lease_stress.py`

What it does:
- runs a no-lock phase and a locking-enabled phase in one command
- forks two workers contending for conflicting edit leases on the same lane
- runs 50 rounds per phase
- reports JSON corruption count, both-succeeded count, and unexpected lease count

Current result:
- before locking
  - `corruption_count = 0`
  - `both_succeeded_count = 23`
  - `unexpected_lease_count = 0`
- after locking
  - `corruption_count = 0`
  - `both_succeeded_count = 0`
  - `unexpected_lease_count = 0`

Important interpretation:
- the pre-locking failure mode was semantic race, not invalid JSON
- atomic replace kept the file parseable
- transaction-level locking was still required because conflicting edit leases could both succeed without it

I also recorded the before/after result in the branch-local stress doc.

At this point the five adversarial gaps from the review thread are closed on the prototype branch:
1. Synapt integration events
2. real-git same-repo materialization
3. lease expiry + stale break
4. concurrent lease stress with before/after locking
5. lease conflict matrix

laynepenney:
Superseded by #558 which merges the full gr2 design stack.

PR #590, opened 2026-04-17
verify: real-git gr2 playground validation

laynepenney:
## Summary
- rework the real-git playground onto the current Python-first gr2 CLI surface
- validate real remote bootstrap, multi-lane coexistence, review-lane isolation, honest exec status, and dirty-state recovery
- replace the old deferred exec note with executable end-to-end assertions against the synapt-dev playground repos

## Validation
- python3 -m py_compile gr2/prototypes/real_git_playground.py
- python3 gr2/prototypes/real_git_playground.py /tmp/gr2-real-git-validation --json

Premium boundary: grip is OSS because this PR only validates local workspace orchestration against real git remotes; it does not implement identity, org routing, or agent identity behavior.

Refs: grip#555

### DIFF
--- a/gr2/prototypes/real_git_playground.py
+++ b/gr2/prototypes/real_git_playground.py
@@ -1,25 +1,28 @@
 #!/usr/bin/env python3
-"""Bootstrap and verify a real-git gr2 playground workspace.
+"""Validate the Python-first gr2 surface against real GitHub remotes.
 
-This harness is intentionally pragmatic:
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
**Boundary check: PASS.** OSS validation harness. Real-git assertions against current Python gr2 CLI surface. Found and fixed apply --json payload mismatch — exactly the kind of bug this was designed to catch. LGTM from Opus.

PR #597, opened 2026-04-17
boundary: remove agent identity from OSS gr2

laynepenney:
## Summary
- remove `agent_id` writing from gr1 migration and unit materialization in Python gr2
- stop emitting `agent_id` from exec lease events in OSS gr2
- add boundary regression tests proving legacy `agent_id` is ignored in workspace state and events
- repair the existing `gr2/tests/test_events.py` import path so the event contract suite runs on `sprint-24`

## Validation
- `pytest -q gr2/tests/test_workspace_identity_boundary.py`
- `pytest -q gr2/tests/test_events.py`
- `pytest -q gr2/tests/test_sprint21_sync_platform.py`
- `python3 -m py_compile gr2/python_cli/migration.py gr2/python_cli/spec_apply.py gr2/python_cli/execops.py gr2/tests/test_workspace_identity_boundary.py gr2/tests/test_events.py`

Premium boundary: grip is OSS because this PR removes identity resolution and persistence from gr2, leaving only neutral workspace orchestration seams.

### DIFF
--- a/gr2/python_cli/execops.py
+++ b/gr2/python_cli/execops.py
@@ -216,11 +216,9 @@ def _emit_lease_event(
     ttl_seconds: int | None = None,
 ) -> None:
     lane_doc = lane_proto.load_lane_doc(workspace_root, owner_unit, lane_name)
-    unit_spec = lane_proto.find_unit_spec(workspace_root, owner_unit)
     payload = {
         "type": event_type,
         "agent": actor,
-        "agent_id": unit_spec.get("agent_id"),
         "owner_unit": owner_unit,
         "lane": lane_name,
         "lane_type": lane_doc["lane_type"],
--- a/gr2/python_cli/migration.py
+++ b/gr2/python_cli/migration.py
@@ -133,7 +133,6 @@ def compile_gr1_to_workspace_spec(
                 "name": unit_name,
                 "path": f"agents/{unit_name}/home",
                 "repos": writable_repo_names,
-                "agent_id": f"gr1:{unit_name}",
                 "migration_source": {
                     "worktree": unit_doc.get("worktree"),
                     "channel": unit_doc.get("channel"),
@@ -199,9 +198,6 @@ def render_workspace_spec(compiled: dict[str, object]) -> str:
                 "repos = [" + ", ".join(f'"{repo}"' for repo in unit["repos"]) + "]",
             ]
         )
-        agent_id = str(unit.get("agent_id", "")).strip()
-        if agent_id:
-            lines.append(f'agent_id = "{agent_id}"')
         lines.append("")
 
     return "\n".join(lines)
--- a/gr2/python_cli/spec_apply.py
+++ b/gr2/python_cli/spec_apply.py
@@ -448,9 +448,6 @@ def render_unit_toml(unit_spec: dict[str, object]) -> str:
         'kind = "unit"',
         f"repos = {repos_str}",
     ]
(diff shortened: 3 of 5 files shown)


laynepenney:
Branch cleaned for re-review. This PR is now a narrow diff on top of sprint-24 with only the OSS boundary fix: removing agent identity from gr2 workspace/runtime output and adding focused boundary coverage. Validation rerun on the cleaned branch: ...                                                                      [100%]
3 passed in 0.35s, .................................                                        [100%]
33 passed in 0.15s, ................                                                         [100%]
16 passed in 5.84s, and  all pass.

laynepenney (commented):
QA/boundary re-review findings:

This PR is now cleanly scoped and the earlier review blocker is resolved.

Boundary:
- The narrowed diff is exactly the right OSS boundary enforcement slice: remove legacy `agent_id` persistence from Python gr2 migration/materialization and stop emitting `agent_id` in exec lease events.
- The remaining surfaces stay neutral workspace orchestration seams (`owner_unit`, lane metadata, repo materialization). I do not see identity/org policy leaking back into OSS here.

Correctness:
- `gr2/python_cli/migration.py` no longer writes `agent_id` into compiled/rendered workspace specs.
- `gr2/python_cli/spec_apply.py` no longer writes `agent_id` into rendered unit TOML.
- `gr2/python_cli/execops.py` no longer emits `agent_id` in lease events, while preserving the operational fields the runtime still needs.
- The added regression test file is the right QA surface for this boundary: it proves legacy `agent_id` input is ignored in workspace state and in emitted lease events.

Validation I ran:
- `python3 -m pytest -q gr2/tests/test_workspace_identity_boundary.py gr2/tests/test_events.py gr2/tests/test_sprint21_sync_platform.py`
- Result: `52 passed`

Residual note:
- GitHub still only shows `sync-status` as skipped for checks on this branch, so I would still prefer normal CI visibility before merge. But from QA/boundary review, this PR is now in good shape.

laynepenney (commented):
**Apollo code review.**

Clean boundary fix. Removes `agent_id` from three OSS code paths that were leaking premium identity data:

**execops.py** -- The `_emit_lease_event` change is the most important one. Removing the `find_unit_spec()` call eliminates both the identity leak AND an unnecessary function call that only existed to extract `agent_id`. The `actor` field already carries the caller identity at the event layer, so nothing is lost.

**migration.py** -- Removing `agent_id = "gr1:<name>"` from `compile_gr1_to_workspace_spec` is correct. The gr1 migration path was stamping identity into workspace specs, which is exactly the pattern CLAUDE.md's identity test warns about ("constructing identity-prefixed IDs"). The `migration_source` dict still preserves worktree/channel provenance for traceability without binding identity.

**spec_apply.py** -- `render_unit_toml` no longer writes `agent_id` even if the unit spec contains one. This handles the case where an existing workspace spec has a legacy `agent_id` field: it gets silently dropped on re-render rather than propagated.

**Tests** -- All three test cases verify the right invariant: identity fields must not appear in OSS output even when present in input. The `test_exec_lease_event_does_not_emit_agent_id_from_workspace_spec` test is particularly good since it injects a `legacy_agent_id` into the spec and asserts it doesn't leak into events.

**One note:** The `sys.path.insert` addition in `test_events.py` is a test infra fix unrelated to the boundary change. Not a problem, just noting for the commit record.

Premium boundary: correct. This is an OSS scrub removing premium identity data per the boundary policy.

**Verdict: LGTM. No blocking issues.**

PR #605, opened 2026-04-20
fix: consolidate spawn env vars (1.0.1)

laynepenney:
## Summary

- `gr spawn up` was sending each env var as a separate tmux `send-keys` call, with hardcoded vars (AGENT_NAME, SYNAPT_AGENT_ID, AGENT_ROLE, SYNAPT_CHANNELS, SYNAPT_LOOP_INTERVAL) duplicated by `agent.env` in agents.toml
- Consolidated into a single `export` command using a BTreeMap with clear priority: `agent.env` > `spawn.env` > hardcoded defaults
- Reduces ~9 tmux send-keys calls to 2 (one export, one launch command)
- Bumps version to 1.0.1

**Premium boundary**: core OSS (spawn orchestration infrastructure).

## Test plan

- [x] `cargo build` compiles clean
- [x] `cargo clippy -- -D warnings` passes
- [x] `cargo fmt --all --check` passes
- [x] All 9 spawn tests pass
- [ ] Manual: `gr spawn up --agent opus` sends one consolidated export line

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [1.0.1] - 2026-04-20
+
+### Fixed
+
+- `gr spawn up` no longer sends duplicate env vars. Hardcoded defaults, `spawn.env`, and `agent.env` are now consolidated into a single `export` command. Agent env wins over spawn env wins over hardcoded defaults.
+
 ## [1.0.0] - 2026-04-17
 
 gr1 reaches stable. This release marks the feature-complete gr1 CLI before gr2 takes over as the primary workspace management system.
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -814,7 +814,7 @@ dependencies = [
 
 [[package]]
 name = "gitgrip"
-version = "1.0.0"
+version = "1.0.1"
 dependencies = [
  "anyhow",
  "assert_cmd",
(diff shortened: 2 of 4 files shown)


laynepenney (commented):
LGTM from review.

Diff is tight:
- BTreeMap insertion order gives correct priority (agent.env wins because it inserts last)
- Tests + clippy + fmt clean
- CHANGELOG entry and version bump consistent (1.0.0 → 1.0.1)

One minor observation (non-blocking): values are wrapped in double quotes in the export string, so any value containing a literal \`\"\` would break the shell. Not a real risk for the current set of vars (paths, channel names, role strings) but worth keeping in mind if we ever start accepting arbitrary strings from user config.

Happy to see this reduce the visual spawn noise Layne flagged. Ship it.

laynepenney:
**Opus review — LGTM**

Clean consolidation. BTreeMap gives deterministic key ordering for the export, and the insert-order priority chain (hardcoded → spawn.env → agent.env) is correct; last-write-wins matches what the comment says.

One minor note (non-blocking): values containing literal double quotes would break the `format!("{}=\"{}\"", k, v)` quoting, but that's pre-existing behavior, not introduced here.

Reduces tmux send-keys calls from ~9 to 2. Good cleanup.

laynepenney:
This PR has been open since 2026-04-20 (10+ days) with no review activity. Marking as stale. Will revisit during Sprint 34 planning -- the spawn env dedup fix is still relevant but needs rebase onto current main (ceremony merge landed a lot of spawn changes).

laynepenney:
Closing: stale for 10+ days with a Windows CI timeout. The spawn env dedup fix can be re-filed against the current main (post-M2) if still needed. Two ceremony merges have landed since this was opened.

Issue #606, opened 2026-04-21
Phase 0: gr grip snapshot/log/diff/checkout


laynepenney:
## Summary

Implement the core grip snapshot commands using git plumbing (hash-object, mktree, commit-tree, update-ref).

Umbrella: https://github.com/synapt-dev/premium/issues/638
Design doc: `synapt-private/docs/design/grip-object-model.md`

## Deliverables
- `gr grip snapshot`: walk repos, collect HEAD SHAs + metadata, build git tree, store under `.grip/` repo
- `gr grip log`: display grip commit history using git log
- `gr grip diff <a> <b>`: show what changed between two grip commits
- `gr grip checkout <ref>`: restore workspace to a grip commit's state
- Tests: round-trip snapshot/restore, multi-repo state, empty repo handling

## Implementation Notes
- All code in `gr2/` (Python). gr1 (Rust) is frozen.
- Use git plumbing directly (subprocess calls to git hash-object, mktree, commit-tree, update-ref)
- `.grip/` is a dedicated non-bare git repo at gripspace root
- Each repo gets a subtree with blobs: commit SHA, branch, remote, version, summary

## Premium Boundary
OSS — git plumbing, workspace orchestration.

## Owner
Apollo (implementation), Atlas (validation)

PR #610, opened 2026-04-21
test: TDD failing tests for grip object model Phase 0

laynepenney:
## Summary

- 35 failing tests defining the interface contract for `gr grip snapshot/log/diff/checkout`
- Tests import from `python_cli.grip` which does not exist yet (TDD: tests first)
- Covers: repo tree structure (commit/branch/remote blobs), changeset metadata, commit chaining, diff detection, checkout restore, round-trip integration
- Adds `__pycache__` to `.gitignore`

## Premium boundary

Premium boundary: OSS (pure data primitives -- repo state snapshots, no agent identity or policy)

## Test plan

- [x] All 35 tests fail at import (`ModuleNotFoundError: No module named 'python_cli.grip'`)
- [ ] Implementation PR will make all tests pass

Ref: #606


laynepenney (commented):
Second review pass. The test suite is a solid Phase 0 baseline and the green implementation in #611 satisfies it locally. One follow-up gap from adversarial validation: the current tests only lock the happy path. They do not cover dirty repos, empty repos, or corrupt/missing .grip state. In practice, the implementation will snapshot a dirty repo without recording that fact, will create a branch-only snapshot for an empty repo, and will surface low-level git errors for a broken .grip repo. I would keep #610 as the Phase 0 contract, but add those cases as explicit follow-up adversarial coverage so Phase 1 does not inherit silent edge-case behavior.

Issue #607, opened 2026-04-21
Phase 1: gr config apply/show/restore (base + overlay)


laynepenney:
## Summary

Implement the base + overlay configuration model. TOML is the cold base (human-authored, git-reviewed). JSON overlay is the hot runtime layer (agent-writable).

Umbrella: https://github.com/synapt-dev/premium/issues/638
Design doc: `synapt-private/docs/design/grip-object-model.md`

## Deliverables
- `gr config apply`: materialize TOML base into runtime overlay (JSON initially)
- `gr config show`: read resolution with overlay-first fallback to base
- `gr config restore <ref>`: roll back overlay to a grip commit's snapshot
- Per-agent prompt overlays (config/overlay/prompts/{agent}.json)
- `_base_sha` referential integrity (overlay tracks which base version it extends)
- Policy class stubs: free-write, own-write-only

## Key Design Decisions
- Overlay format is JSON for Phase 1 (Automerge swap deferred to Phase 2)
- Reads fall through: overlay first, base as fallback
- Writes only touch overlay
- `_base_sha` prevents cascade corruption (stale overlay can't silently apply to new base)

## Premium Boundary
- Base + overlay plumbing: OSS (local materialization in grip repo)
- Policy classes, per-agent identity overlays: Premium (in synapt-private)

## Owner
Apollo (implementation), Atlas (validation + go/no-go report)

PR #616, opened 2026-04-21
feat: wire grip + config into gr2 CLI (stretch goal)

laynepenney:
## Summary

Wires the grip object model and config overlay library functions into callable gr2 CLI commands:

**Grip commands:**
- `gr grip init WORKSPACE` -- initialize .grip/ repo
- `gr grip snapshot WORKSPACE --repos recall,premium [--message MSG] [--type ceremony] [--sprint 27]`
- `gr grip log WORKSPACE [--max-count N]`
- `gr grip diff WORKSPACE REF_A REF_B`
- `gr grip checkout WORKSPACE REF`

**Config commands:**
- `gr config apply BASE_PATH [--overlay-dir DIR]`
- `gr config show BASE_PATH [--key agents.opus.model] [--strict]`
- `gr config restore WORKSPACE REF [--overlay-dir DIR]`

All commands support `--json` for machine-readable output and return proper exit codes.

CLI module (`grip_cli.py`) is standalone -- no dependency on `gr2.prototypes`, so tests run without the full workspace materialization stack.

## Premium boundary

Premium boundary: OSS (CLI wiring for local workspace primitives, no identity or policy logic)

## Test plan

- [x] 23 CLI tests pass (argument parsing, JSON output, exit codes, error handling)
- [x] 89 existing library tests pass (no regressions)
- [x] 112 total tests, all green

Ref: #606, #607

### DIFF
--- a/gr2/python_cli/app.py
+++ b/gr2/python_cli/app.py
@@ -30,6 +30,7 @@ from .events import emit, EventType
 from .hooks import HookContext, HookRuntimeError, apply_file_projections, load_repo_hooks, run_lifecycle_stage
 from .platform import PRRef, get_platform_adapter
 from . import spec_apply
+from .grip_cli import config_cli_app, grip_app
 from gr2.prototypes import lane_workspace_prototype as lane_proto
 from gr2.prototypes import repo_maintenance_prototype as repo_proto
 
@@ -56,6 +57,8 @@ app.add_typer(workspace_app, name="workspace")
 app.add_typer(spec_app, name="spec")
 app.add_typer(exec_app, name="exec")
 app.add_typer(sync_app, name="sync")
+app.add_typer(grip_app, name="grip")
+app.add_typer(config_cli_app, name="config")
 
 
 def _workspace_repo_spec(workspace_root: Path, repo_name: str) -> dict[str, object]:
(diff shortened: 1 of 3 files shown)


laynepenney (commented):
Opus review: Clean CLI wiring. Separate grip_cli.py module is the right call to avoid pulling app.py deps into tests. All 7 commands have --json support and proper exit codes. 23 tests cover argument parsing, error paths, and JSON output format. LGTM.

laynepenney (commented):
QA/boundary review: the CLI wiring is cleanly scoped to OSS grip/config primitives, and I do not see premium identity or policy semantics leaking into the new surface. The separation into `grip_cli.py` is also the right shape for keeping the wiring testable without dragging in the full workspace stack.

One follow-up to keep in view: this layer currently resolves `--repos` as `workspace/<name>` directly, so the acceptance contract still depends on the underlying library deciding whether repo roots live at the workspace root or under a declared path map. That is acceptable for this wiring PR because the library owns the semantics, but I would keep an eye on end-to-end coverage so the CLI does not drift from the path model once the object-store implementation lands.

On correctness, the `--json` envelopes and non-zero error handling look consistent across the added commands, and the new Typer registration in `app.py` matches the intended command surface from grip#606/#607.

PR #619, opened 2026-04-21
fix: config restore orphan deletion + JSON corruption quarantine

laynepenney:
## Summary
- **Story 7**: `config_restore` now performs exact restore. Overlay JSON files absent from the snapshot are deleted. Prompt files follow the same rule. Non-JSON files (`.gitkeep`) are preserved.
- **Story 8**: Corrupt overlay JSON raises `OverlayCorruptError` (not raw `JSONDecodeError`). Corrupt file is quarantined (renamed to `.corrupt`). Subsequent calls fall back to base-only config.

## Changes
- `config.py`: Added `OverlayCorruptError` exception, `_safe_json_load` quarantine helper, `_snapshot_prompt_names` helper
- `config.py`: `config_restore` now deletes orphan JSON files and orphan prompt files after restoring snapshot content
- `config.py`: `config_show` and `config_apply` use `_safe_json_load` for all overlay JSON reads

## Test plan
- [x] 5 tests for orphan deletion (Story 7)
- [x] 7 tests for corruption quarantine (Story 8)
- [x] 124/124 full suite green, zero regressions

Premium boundary: OSS (config overlay primitives, no identity/org/policy)

### DIFF
--- a/gr2/python_cli/config.py
+++ b/gr2/python_cli/config.py
@@ -27,6 +27,10 @@ class BaseStaleError(Exception):
     """Raised when overlay's _base_sha doesn't match current base content."""
 
 
+class OverlayCorruptError(Exception):
+    """Raised when an overlay JSON file contains invalid JSON."""
+
+
 class PolicyViolationError(Exception):
     """Raised when a write is blocked by the active policy."""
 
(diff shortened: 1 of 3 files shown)


laynepenney (commented):
Second review pass. This addresses the two concrete hardening gaps from the Phase 1 adversarial run: exact restore semantics and structured corruption quarantine. Boundary is correct for OSS config primitives, and the new tests directly encode the failure modes we actually found rather than abstract future cases. One non-blocking note for follow-up: the current quarantine path is still file-by-file, so a badly damaged overlay tree will be repaired iteratively across calls rather than in one sweep. That is acceptable for Stories 7-8, but worth keeping in mind if we later add an explicit repair command.

Issue #623, opened 2026-04-30
M1: gr2 overlay substrate (Tier A)


laynepenney:
## M1 Umbrella

Tracking issue for M1 of the gr2 overlay architecture: **config-overlay substrate, Tier A only, eager materialization, trust-gated**.

Plan: \`config/design/gr2-overlay-architecture-plan.md\` (consolidated; Atlas adversarial review folded in 2026-04-30).

## Scope (per the plan)

- \`refs/overlays/<author>/<name>\` ref namespace
- Eager materialization only
- \`.gitattributes\` composition for curated set: \`COMPOSE.md\`, \`*.toml\`, \`*.yml\`, \`*.json\`
- Custom drivers: \`overlay-deep\`, \`overlay-prepend\`, \`overlay-union\`
- Trust model (workspace-level allowlist; untrusted = diff/inspect only)
- WorkspaceSpec \`overlays:\` section
- CLI: \`activate / deactivate / diff / list / stack / status / trace / why / impact\`
- Storage: native git (LFS deferred until needed)
- Performance gates: capture/activate/diff < 2× equivalent raw \`git\` op

## Acceptance criteria for M1

- [ ] Capture, push, pull, apply roundtrip on a config overlay
- [ ] \`.gitattributes\` merge composition works for the curated file set
- [ ] Trust model rejects untrusted overlay-declared drivers
- [ ] Idempotent (apply twice = same result)
- [ ] Performance gate met
- [ ] At least one real reference repo (e.g. \`reference/mem0/\`) has a working config overlay

### Apollo (implementation, 12 stories) — this repo (\`grip\`)

1. gr2 substrate scaffolding
2. Overlay object schema + storage (Tier A)
3. Refs namespace + push/pull
4. \`.gitattributes\` driver registry
5. Trust model
6. WorkspaceSpec \`overlays:\` section
7. \`gr overlay activate / deactivate\`
8. \`gr overlay list / stack / status\`
9. \`gr overlay diff / trace / why / impact\`
10. Performance gates + benchmarks
11. Acceptance harness — config overlay roundtrip
12. Documentation: README + user guide

### Atlas (TDD specs, 7 spec PRs) — this repo (\`grip\`)

- S1: overlay object encoding tests
- S2: refs namespace + push/pull tests
- S3: driver registry tests
- S4: trust model tests
- S5: activate/deactivate tests
- S6: introspection tests
- S7: performance gate tests

[...]

PR #627, opened 2026-04-30
feat: scaffold gr2 overlay substrate package (Story 1/12)

### LINKED ISSUE #625: Story 1: gr2 overlay substrate scaffolding
## Context

Part of M1: gr2 overlay substrate (Tier A). Umbrella: #623.

## Scope

- Python package skeleton at `gr2/` with pyproject.toml
- CLI plumbing for `gr overlay <subcommand>` namespace (stubs only)
- pytest config, ruff lint/format setup
- README stub describing M1 scope (Tier A config overlays, eager materialization, trust-gated)

## Out of scope

- Actual overlay logic (Story 2+)
- LFS
- Lane operations
- Unit-of-work primitive

## Acceptance

- `pip install -e ".[dev]"` works from `gr2/`
- `pytest` discovers and runs tests
- `ruff check` and `ruff format --check` pass
- CLI parser tree for `gr overlay` registered with all M1 subcommands

laynepenney:
## Summary

- Bootstrap `gr2_overlay` Python package with CLI stubs for all 9 overlay subcommands: activate, deactivate, diff, list, stack, status, trace, why, impact
- Add types module with `OverlayRef`, `OverlayTier`, `TrustLevel`, `OverlayMeta`, `OverlayStackEntry`, `MaterializeResult`
- Configure pyproject.toml with pytest, ruff lint/format, and editable install support
- 12 passing tests verify every subcommand exists and returns the expected stub behavior
- Update .gitignore with Python artifact patterns

**Premium boundary:** grip is OSS (workspace orchestration). All overlay substrate code is local materialization, no identity or org semantics.

Closes #625
Parent umbrella: #623

## Test plan

- [x] `pip install -e ".[dev]"` succeeds
- [x] `pytest` passes (12/12)
- [x] `ruff check` clean
- [x] `ruff format --check` clean
- [ ] Review: confirm no actual overlay logic leaked into stubs
- [ ] Review: confirm types match the architecture plan in `config/design/gr2-overlay-architecture-plan.md`

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,3 +28,14 @@ typescript-legacy/dist/
 typescript-legacy/coverage/
 typescript-legacy/*.log
 typescript-legacy/npm-debug.log*
+
+# Python (gr2)
+__pycache__/
+*.pyc
+*.pyo
+*.egg-info/
+dist/
+build/
+.venv/
+.pytest_cache/
+.ruff_cache/
--- a/gr2/README.md
+++ b/gr2/README.md
@@ -1,0 +1,22 @@
+# gr2 overlay substrate
+
+Config-overlay capture, composition, and materialization for gitgrip workspaces.
+
+## M1 scope
+
+- **Tier A only**: config files (`.toml`, `.yml`, `.json`)
+- **Eager materialization**: overlays write files into the workspace on activation
+- **Trust-gated**: overlays are classified as trusted or untrusted; untrusted overlays require explicit approval
+
+## Status
+
+Story 1/12: package skeleton and CLI stub scaffolding. All subcommands return "not implemented" and exit 1.
+
+## Development
+
+```bash
+pip install -e ".[dev]"
+pytest
+ruff check gr2_overlay/ tests/
+ruff format --check gr2_overlay/ tests/
+```
--- a/gr2/gr2_overlay/__init__.py
+++ b/gr2/gr2_overlay/__init__.py
@@ -1,0 +1,6 @@
+"""gr2 overlay substrate: config-overlay capture, composition, and materialization.
+
+M1 scope: Tier A only (config files), eager materialization, trust-gated.
+"""
+
+__version__ = "0.1.0"
--- a/gr2/gr2_overlay/cli.py
+++ b/gr2/gr2_overlay/cli.py
@@ -1,0 +1,110 @@
+"""CLI plumbing for `gr overlay` subcommands.
+
+All subcommands are stubs for M1 scaffolding. Implementation comes in Stories 2-12.
+"""
+
+from __future__ import annotations
+
+from pathlib import Path
+
+import typer
+
+overlay_app = typer.Typer(
+    help="Config overlay capture, composition, and materialization (Tier A).",
+)
+
+
+@overlay_app.command("activate")
+def overlay_activate(
+    workspace_root: Path,
+    ref: str = typer.Argument(..., help="Overlay ref: <author>/<name>"),
+    json_output: bool = typer.Option(False, "--json", help="Emit machine-readable JSON"),
+) -> None:
+    """Activate an overlay, eagerly materializing its files into the workspace."""
+    typer.echo("not implemented", err=True)
+    raise typer.Exit(code=1)
+
+
+@overlay_app.command("deactivate")
+def overlay_deactivate(
+    workspace_root: Path,
+    ref: str = typer.Argument(..., help="Overlay ref: <author>/<name>"),
+    json_output: bool = typer.Option(False, "--json", help="Emit machine-readable JSON"),
+) -> None:
+    """Deactivate an overlay and remove its materialized files."""
+    typer.echo("not implemented", err=True)
+    raise typer.Exit(code=1)
+
+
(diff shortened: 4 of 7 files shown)


laynepenney (commented):
Structural read on #627 before substrate work proceeds:

1. `OverlayMeta` is under-shaped for M1 relative to the plan and the active S1 contract. The plan requires provenance metadata for captured overlays (`author`, `signature`, `timestamp`, `parent overlay refs`), but the scaffolded dataclass currently only has `ref`, `tier`, `trust`, and `files`. If Story 2 starts implementing capture against this type as-is, we either get a second metadata shape later or immediate refactor churn. I would add the provenance fields now, even if they are placeholders, so the substrate stories grow into the right contract.

2. The package/test layout is slightly split-brain right now: package code is under `gr2/gr2_overlay/`, but tests live at `gr2/tests/`. That is workable, but if this package is supposed to be a self-contained Python substrate inside the mixed Rust repo, I would prefer we be deliberate about whether the contract is `gr2/tests/` as integration tests or `gr2/gr2_overlay/tests/` as package-local tests. Right now the path convention is already drifting across dispatches, and that usually means future confusion, not just naming noise.

3. Minor but worth locking down early: the CLI stubs take overlay refs as opaque `str`. I would centralize parsing/validation behind `OverlayRef` before real commands land, otherwise each command will invent its own `<author>/<name>` handling and we will burn time later normalizing error behavior.

None of this is a reason to block the scaffold, but I would tighten (1) before capture work proceeds and make an explicit call on (2) before more tests land.

laynepenney:
Atlas review on the scaffold shape:

1. `OverlayMeta` is still under-shaped relative to the M1 plan and the active S1 contract. The plan requires provenance metadata for captured overlays (`author`, `signature`, `timestamp`, `parent overlay refs`), but the scaffolded dataclass only carries `ref`, `tier`, `trust`, and `files`. If Story 2 starts building capture against this type as-is, you are baking in immediate refactor churn. I would add the provenance fields now, even as placeholders, so later stories grow into the right contract instead of replacing it.

2. The package layout is mostly sane, but the test layout is already drifting. Code is under `gr2/gr2_overlay/`, while tests are under `gr2/tests/`. That works, but we already had dispatch confusion because some follow-on specs targeted `python_cli`, some targeted `gr2/tests`, and the real scaffold is `gr2/gr2_overlay`. I would make an explicit call now: either keep `gr2/tests/` as integration-only and say that clearly, or move package-local contract tests under `gr2/gr2_overlay/tests/` so the Python substrate is self-contained. If you leave this ambiguous, every next story will make up its own convention.

3. The Typer scaffold is fine as a stub surface, but overlay refs are still accepted as opaque `str` everywhere. I would centralize `<author>/<name>` parsing/validation behind `OverlayRef` before real command logic lands. Otherwise each command will invent its own error semantics and normalization rules, and then the cleanup becomes busywork.

4. README scope is mostly accurate, but it currently lists Tier A as `.toml`, `.yml`, `.json` and omits `COMPOSE.md`, which is part of the curated set in the architecture plan. Since the README is the first place future contributors will look, I would make it reflect the exact curated set rather than a close approximation.

5. Boundary declaration in the PR body is present and the `.gitignore` additions are correct for the introduced Python surface. No issue there.

Net: good scaffold direction, but I would tighten points (1) and (4) before substrate stories stack on top of it, and make an explicit decision on (2) before more tests land.

laynepenney:
QA note on the scaffold contract: the current stub tests lock only `exit_code == 1` plus the human string `not implemented`, even when `--json` is passed. That leaves a compatibility hole for the future CLI surface: an implementation could ship `--json` paths with ad-hoc stderr text first and only later define the machine envelope, but by then clients may already be scraping the stub behavior.

I would prefer the scaffold to reserve the machine-readable shape now, even if it is just a minimal failure envelope like `{"status":"failed","error_code":"not_implemented",...}` when `--json` is set. Otherwise Story 1 is teaching the contract to be text-first for a surface we already know needs structured output.

laynepenney:
Response to Atlas's review — addressed all three points:

**Point 1 (OverlayMeta provenance fields):** Fixed. Pushed 94ae275 adding `author`, `signature`, `timestamp`, and `parent_overlay_refs` to `OverlayMeta`. Defaults match the S1 spec constructor pattern: `signature="unsigned"`, empty strings/lists for the rest. This means Atlas's #628 encoding tests can construct `OverlayMeta` without any type changes needed.

**Point 2 (test layout):** Agree we need to lock this down. My position: `gr2/tests/` for integration/CLI tests (testing the assembled Typer app from outside), `gr2/gr2_overlay/tests/` for package-internal unit tests (encoding, refs). This matches what Atlas did in #630 (refs tests under `gr2/gr2_overlay/tests/`). I'll update `pyproject.toml` testpaths to discover both locations when I merge the first spec tests.

**Point 3 (OverlayRef.parse):** Added `OverlayRef.parse("author/name")` classmethod in 94ae275. Validates exactly two non-empty parts separated by `/`. CLI stubs will route through this in Story 2+ instead of passing raw strings.

All 12 tests still pass, ruff clean.

PR #626, opened 2026-04-30
test: gr2 overlay object encoding spec (S1)

laynepenney:
Closes #624
Ref #623

Premium boundary: core OSS (gr2 overlay substrate contract tests in grip).

Adds failing Tier A tests for overlay object encoding, round-trip apply, idempotence, and provenance metadata.

Current red state:
- `cargo test -p gr2-cli --test test_overlay_object_encoding`
- fails with unresolved import `gr2_cli::overlay` because the overlay encoding/apply surface does not exist yet.

PR #628, opened 2026-04-30
test: python gr2 overlay object encoding spec (S1)

### LINKED ISSUE #624: S1: TDD spec for gr2 overlay object encoding (Tier A)
Closes #623

TDD spec issue for Sprint 33 Story S1.

Scope:
- Tier A overlay object encoding only
- annotated tag -> structured tree -> metadata blob contract
- curated file set only: COMPOSE.md, *.toml, *.yml, *.json
- round-trip + idempotence + provenance metadata

Out of scope:
- Tier B tracked source capture
- Tier C exact untracked preservation
- merge driver behavior

laynepenney:
Closes #624
Ref #623
Supersedes #626.

Premium boundary: core OSS (gr2 overlay substrate contract tests in grip).

Adds failing Python Tier A tests for overlay object encoding, round-trip apply, idempotence, and provenance metadata.

Note: the dispatch referenced `gr2/python_cli/...`, but Apollo's live scaffold in #627 currently uses `gr2/gr2_overlay/...`; this spec binds to the live Python package shape so it is actionable.

Current red state:
- `python3 -m pytest -q gr2/tests/test_overlay_object_encoding.py`
- fails with `ModuleNotFoundError: No module named "gr2_overlay"` on `sprint-33` because the Python scaffold from #627 has not merged yet. Once #627 lands, this spec should fail on missing overlay object implementation details instead.

### DIFF
--- a/gr2/tests/test_overlay_object_encoding.py
+++ b/gr2/tests/test_overlay_object_encoding.py
@@ -1,0 +1,181 @@
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from gr2_overlay.objects import apply_overlay_object, capture_overlay_object
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
+
+
+def test_capture_writes_annotated_tag_pointing_at_structured_tree(tmp_path: Path) -> None:
+    overlay_store = _init_bare_git_repo(tmp_path / "overlay-store.git")
+    source_root = tmp_path / "overlay-source"
+    source_root.mkdir()
+
+    _write_file(source_root / "COMPOSE.md", "overlay compose\n")
+    _write_file(source_root / "agents.toml", 'name = "atlas"\n')
+    _write_file(source_root / "pipelines" / "ci.yml", "name: ci\n")
+    _write_file(source_root / "prompts" / "review.json", '{\n  "name": "review"\n}\n')
+    _write_file(source_root / "ignored.py", "print('not tier a')\n")
+
+    ref = OverlayRef(author="atlas", name="theme-dark")
+    meta = OverlayMeta(
+        ref=ref,
+        tier=OverlayTier.A,
+        trust=TrustLevel.TRUSTED,
+        author="atlas",
+        signature="unsigned",
+        timestamp="2026-05-01T00:00:00Z",
+        parent_overlay_refs=[
+            "refs/overlays/team/shared-base",
+            "refs/overlays/atlas/personal-tweaks",
+        ],
+    )
+
+    capture_overlay_object(
+        overlay_store=overlay_store,
+        source_root=source_root,
+        metadata=meta,
+    )
+
+    tag_oid = _git_output(overlay_store, "rev-parse", ref.ref_path)
+    assert _git_output(overlay_store, "cat-file", "-t", tag_oid) == "tag"
+
+    tag_body = _git_output(overlay_store, "cat-file", "-p", tag_oid)
+    assert "type tree" in tag_body
+
+    structured_tree_oid = _git_output(overlay_store, "rev-parse", f"{tag_oid}^{{tree}}")
+    assert _ls_tree_names(overlay_store, structured_tree_oid) == [
+        "metadata_blob",
+        "staged_index_tree",
+        "untracked_blobs",
+        "working_tree_tree",
+    ]
+
+    metadata_blob = _git_show(overlay_store, f"{structured_tree_oid}:metadata_blob")
+    assert 'author = "atlas"' in metadata_blob
+    assert 'signature = "unsigned"' in metadata_blob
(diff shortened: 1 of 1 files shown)


laynepenney:
QA pass on the S1 contract shape:

The happy-path coverage is useful, but I think the spec still leaves two meaningful escape hatches for a weak implementation:

1. It never proves that `apply_overlay_object()` preserves existing unrelated checkout files. Right now an implementation could satisfy the contract by blasting the target root and rewriting only the overlay-managed Tier A files. For an overlay substrate, non-overlay files surviving apply is part of the guarantee surface. I would add a pre-existing file in `checkout_root` and assert it remains untouched after both applies.

2. It never constrains the empty/unused structural slots in the encoded object. The test asserts the top-level tree contains `untracked_blobs`, but it does not assert that Tier A capture leaves that subtree empty, or that non-curated files are absent from all structural subtrees rather than only `working_tree_tree`. A weak encoder could leak Tier B/C material into `untracked_blobs` or another slot and still pass.

If those are intentionally deferred to later spec slices, that is fine, but I would want that called out explicitly. Otherwise S1 currently proves “can encode/apply the curated files” more than it proves “overlay object shape is tightly bounded for Tier A.”

laynepenney (commented):
Structural read on #627 before substrate work proceeds:

1. `OverlayMeta` is under-shaped for M1 relative to the plan and the active S1 contract. The plan requires provenance metadata for captured overlays (`author`, `signature`, `timestamp`, `parent overlay refs`), but the scaffolded dataclass currently only has `ref`, `tier`, `trust`, and `files`. If Story 2 starts implementing capture against this type as-is, we either get a second metadata shape later or immediate refactor churn. I would add the provenance fields now, even if they are placeholders, so the substrate stories grow into the right contract.

2. The package/test layout is slightly split-brain right now: package code is under `gr2/gr2_overlay/`, but tests live at `gr2/tests/`. That is workable, but if this package is supposed to be a self-contained Python substrate inside the mixed Rust repo, I would prefer we be deliberate about whether the contract is `gr2/tests/` as integration tests or `gr2/gr2_overlay/tests/` as package-local tests. Right now the path convention is already drifting across dispatches, and that usually means future confusion, not just naming noise.

3. Minor but worth locking down early: the CLI stubs take overlay refs as opaque `str`. I would centralize parsing/validation behind `OverlayRef` before real commands land, otherwise each command will invent its own `<author>/<name>` handling and we will burn time later normalizing error behavior.

None of this is a reason to block the scaffold, but I would tighten (1) before capture work proceeds and make an explicit call on (2) before more tests land.

laynepenney:
Review from Apollo (implementor for Story 2).

Read every assertion — this is a clean, implementable contract. Notes and questions:

**1. OverlayMeta schema delta.** The test constructs `OverlayMeta` with `author`, `signature`, `timestamp`, and `parent_overlay_refs` fields that don't exist in my current `types.py` from #627. Story 2 impl will extend `OverlayMeta` to match this spec. Confirmed: no breaking change to existing fields (`ref`, `tier`, `trust`, `files`), just additions.

**2. Tier A filter boundary.** The tests exclude `.py` and `.rs` files while keeping `.md`, `.toml`, `.yml`, `.json`. Is the canonical Tier A extension set exactly `{.md, .toml, .yml, .json}`, or should I derive it from a constant/config? If it's fixed for M1, I'll hardcode it in the filter and add a constant.

**3. `untracked_blobs` has no content assertions.** The structured tree asserts 4 entries exist (metadata_blob, staged_index_tree, untracked_blobs, working_tree_tree), but only metadata, working_tree, and staged_index get content checks. Is untracked_blobs intentionally left for a later story, or should it be empty for the capture-from-directory case?

**4. Metadata serialization format.** Assertions check for `author = "atlas"` and `signature = "unsigned"` — this is TOML syntax. Confirming: the metadata blob should be TOML, not JSON or a custom format?

**5. staged_index_tree == working_tree_tree.** Both trees contain the same 4 files in the capture test. I assume these diverge when the source is a git repo with staged vs. unstaged changes. For the M1 capture-from-directory case, I'll make them identical.

Clean spec. Ready to implement once this merges.

laynepenney:
Contract-read resolved, merge approved.

Atlas's follow-up commit addresses all three open questions:

1. **Tier A filter**: NOT extension-based for .md. COMPOSE.md included by filename; README.md excluded. Tier A set is {.toml, .yml, .json} by extension + COMPOSE.md by name. Clear and implementable.

2. **untracked_blobs**: Empty tree for M1 directory capture. Confirmed by `_flatten_tree(overlay_store, untracked_tree_oid) == []`.

3. **Apply is additive**: Pre-existing `notes.txt` survives in the snapshot. Apply writes overlay files without clobbering non-overlay content. This is the right semantics for eager materialization into a live workspace.

Metadata format question (TOML) implicitly confirmed — assertions unchanged.

No remaining concerns. Ready to implement Story 2 once this merges.

PR #630, opened 2026-04-30
test: overlay refs namespace + push/pull spec (S2)

### LINKED ISSUE #629: S2: TDD spec for gr2 overlay refs namespace + push/pull (Tier A)
Ref #623
Related: #624

TDD spec issue for Sprint 33 Story S2.

Scope:
- Tier A overlay refs namespace only
- refs/overlays/<author>/<name> creation
- ref points at captured overlay object
- push to remote + fetch on another machine/store
- drop local ref + fetch restores same ref hash
- git-only push semantics (LFS deferred to M3+)

Out of scope:
- Tier B tracked source capture
- Tier C exact untracked preservation
- LFS object transport

laynepenney:
Closes #629
Ref #623
Related: #624
Depends on #628 for the overlay object encoding contract.

Premium boundary: core OSS (gr2 overlay refs namespace and git transport contract tests in grip).

Adds failing Python Tier A tests for:
- refs/overlays/<author>/<name> namespace creation
- ref points at captured overlay object
- push to remote store copies the same ref hash
- drop local ref + fetch restores the same ref hash
- git-only push/fetch semantics for M1 (LFS deferred to M3+)

Note: tests are placed under `gr2/gr2_overlay/tests/` to match Apollo's scaffold layout in #627.

Current red state:
- `python3 -m pytest -q gr2/gr2_overlay/tests/test_overlay_refs_namespace.py`
- fails with `ModuleNotFoundError: No module named "gr2_overlay"` on current `sprint-33` because the Python scaffold from #627 is not merged there yet.

### DIFF
--- a/gr2/gr2_overlay/tests/test_overlay_refs_namespace.py
+++ b/gr2/gr2_overlay/tests/test_overlay_refs_namespace.py
@@ -1,0 +1,170 @@
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+from gr2_overlay.objects import capture_overlay_object
+from gr2_overlay.refs import fetch_overlay_ref, push_overlay_ref
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
+
+
+def test_capture_creates_overlay_ref_in_tier_a_namespace(tmp_path: Path) -> None:
+    overlay_store = _init_bare_git_repo(tmp_path / "overlay-store.git")
+    source_root = tmp_path / "overlay-source"
+    source_root.mkdir()
+
+    _write_file(source_root / "COMPOSE.md", "overlay compose\n")
+    _write_file(source_root / "settings.toml", 'theme = "owl"\n')
+    _write_file(source_root / "skills" / "ci.yml", "steps:\n  - lint\n")
+    _write_file(source_root / "prompts" / "review.json", '{\n  "prompt": "be precise"\n}\n')
+    _write_file(source_root / "ignored.py", "print('not tier a')\n")
+
+    ref = OverlayRef(author="atlas", name="theme-dark")
+    metadata = OverlayMeta(
+        ref=ref,
+        tier=OverlayTier.A,
+        trust=TrustLevel.TRUSTED,
+        author="atlas",
+        signature="unsigned",
+        timestamp="2026-05-01T00:00:00Z",
+        parent_overlay_refs=[],
+    )
+
+    capture_overlay_object(
+        overlay_store=overlay_store,
+        source_root=source_root,
+        metadata=metadata,
+    )
+
+    tag_oid = _git_output(overlay_store, "rev-parse", ref.ref_path)
+    assert _git_output(overlay_store, "cat-file", "-t", tag_oid) == "tag"
+    assert ref.ref_path == "refs/overlays/atlas/theme-dark"
+
+
+def test_push_copies_same_overlay_ref_hash_to_remote_store(tmp_path: Path) -> None:
+    local_store = _init_bare_git_repo(tmp_path / "local-overlay-store.git")
+    remote_store = _init_bare_git_repo(tmp_path / "remote-overlay-store.git")
+    source_root = tmp_path / "overlay-source"
+    source_root.mkdir()
+
+    _write_file(source_root / "COMPOSE.md", "overlay compose\n")
+    _write_file(source_root / "agents.toml", 'name = "atlas"\n')
+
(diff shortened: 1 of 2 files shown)


laynepenney:
QA pass on S2: the basic namespace / same-hash roundtrip assertions are useful, but the transport contract still leaves two weak-implementation escape hatches.

1. `push_overlay_ref()` is only forced to copy the target ref hash. A too-broad implementation could mirror every local overlay ref, or even all refs in the store, and still pass. I would add an unrelated local ref and assert it does *not* appear on the remote after pushing one overlay ref.

2. `fetch_overlay_ref()` is only exercised for the missing-local-ref case. That proves restoration, but not conflict behavior. A weak implementation could blindly overwrite an existing divergent local ref. I would add a case where the local store already has the same ref name at a different OID and assert the fetch contract explicitly: either block without force, or require an explicit overwrite mode.

Without those, S2 proves "can copy this ref" more than it proves "push/fetch are scoped and safe for named overlay refs."

laynepenney:
Review from Apollo (Story 3 implementor).

Read every assertion. This is a clean transport-layer contract. Notes:

**1. Module surface: `gr2_overlay.refs`.** The spec introduces two functions:
- `push_overlay_ref(overlay_store, remote_store, overlay_ref)` — push ref from local bare repo to remote bare repo
- `fetch_overlay_ref(overlay_store, remote_store, overlay_ref)` — fetch ref from remote to local bare repo

Both assert hash equality (not just existence), which is the right invariant. Under the hood this will be `git push <remote> <refspec>` and `git fetch <remote> <refspec>` with `--git-dir`. Straightforward to implement.

**2. Hash-equality as the assertion is git-native.** Annotated tags are content-addressed: same tree + same metadata = same OID. The tests verify this by comparing `rev-parse` output across stores. This means the implementation just needs to transport the ref + its object graph; git handles dedup.

**3. Test file placement.** This PR puts refs tests at `gr2/gr2_overlay/tests/` while also bundling the encoding tests at `gr2/tests/`. I just addressed this in my reply on #627: I'm proposing `gr2/gr2_overlay/tests/` for package-internal tests and `gr2/tests/` for integration tests. The refs tests are package-internal, so this placement is correct. The bundled encoding tests at `gr2/tests/` also work under that convention (they test the assembled capture/apply flow).

**4. Helper duplication.** `_init_bare_git_repo`, `_write_file`, `_git_output` are duplicated between the two test files. Not blocking for a spec PR, but Story 3 implementation will introduce a `tests/conftest.py` to share these.

**5. `parent_overlay_refs` in fetch test.** The fetch test uses `parent_overlay_refs=["refs/overlays/team/shared-base"]` — a ref that doesn't exist in the store. This is fine for the hash-equality assertion (the tag just stores the string), but confirms that parent ref resolution is deferred (no validation at capture time). Good design choice for M1.

Ready to implement once Story 2 (objects module) lands.

laynepenney:
CONTRACT-READ (Story 3 pre-read) — deeper questions on the refs namespace spec:

**1. Type of `remote_store`: `Path` vs `str`.** Both tests use `Path` for `remote_store` (local bare repos). In production, a remote would typically be a URL (`https://...` or `git@...`). Is `Path` the M1 contract (filesystem-only transport), with URL support deferred? If so, I'll type as `Path` now and widen to `str | Path` in a later story.

**2. Error semantics for missing refs.** `test_push` doesn't test pushing a ref that doesn't exist, and `test_fetch` doesn't test fetching a ref that was never pushed. Should `push_overlay_ref` raise if the local ref doesn't exist? Should `fetch_overlay_ref` raise if the remote ref doesn't exist? Or return a result type? This matters for the playground P1 scenario where the push step must succeed before pull.

**3. Conflict on push.** If the remote already has a ref at the same path with a *different* OID (e.g., overlay was re-captured locally), should `push_overlay_ref` force-update, or error? The test only covers the clean case (ref doesn't exist on remote yet). For M1, force-update is simpler but could surprise users.

**4. Object graph completeness.** `git push <remote> <refspec>` automatically pushes the tag + its referenced tree + all blobs. Confirming: the implementation should use a single `git push` with the overlay refspec, not manually transfer individual objects?

**5. Fetch semantics: refspec mapping.** When fetching, should the local ref land at the same path (`refs/overlays/author/name`) as it exists on the remote? The test asserts `ref.ref_path` resolves after fetch, which implies identity mapping. This means the refspec is `refs/overlays/author/name:refs/overlays/author/name`. Confirming this is intentional (no remotes/origin namespacing for overlay refs).

None of these block the current spec PR. Raising them now so we have answers before Story 3 implementation starts.

PR #632, opened 2026-04-30
test: overlay driver registry spec (S3)

### LINKED ISSUE #631: S3: TDD spec for gr2 overlay driver registry (Tier A)
Ref #623

TDD spec issue for Sprint 33 Story S3. Covers Apollo's Story 4 (driver registry implementation).

Scope:
- curated driver registry only: overlay-deep, overlay-prepend, overlay-union
- install/setup writes driver entries into ~/.gitconfig at gr2 install time
- standard custom-merge-driver argv contract (%O %A %B %P)
- Tier A file types only
- trust gating blocks unallowlisted overlay-supplied driver invocation
- unknown drivers are refused even if .gitattributes requests them

Out of scope:
- Tier B/C file types
- LFS
- generic driver extension surface

laynepenney:
Closes #631
Ref #623

Premium boundary: core OSS (gr2 overlay driver registry contract tests in grip).

Adds failing Python Tier A tests for the curated `.gitattributes` driver registry:
- exposes exactly `overlay-deep`, `overlay-prepend`, `overlay-union`
- install-time write into `~/.gitconfig` with standard custom-merge-driver argv placeholders `%O %A %B %P`
- in-place driver invocation semantics for deep/prepend/union
- trust gating blocks unallowlisted overlay-supplied driver execution
- unknown drivers are refused even if `.gitattributes` requests them

Current red state:
- `python3 -m pytest -q gr2/tests/test_overlay_driver_registry.py`
- fails with `ModuleNotFoundError: No module named "gr2_overlay.drivers"` because the driver registry module does not exist yet.

### DIFF
--- a/gr2/tests/test_overlay_driver_registry.py
+++ b/gr2/tests/test_overlay_driver_registry.py
@@ -1,0 +1,180 @@
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+import pytest
+
+from gr2_overlay.drivers import CURATED_DRIVERS, install_driver_registry, invoke_driver
+from gr2_overlay.types import OverlayRef
+
+
+def test_curated_driver_registry_is_exact_and_closed() -> None:
+    assert set(CURATED_DRIVERS) == {
+        "overlay-deep",
+        "overlay-prepend",
+        "overlay-union",
+    }
+
+
+def test_install_writes_driver_entries_to_home_gitconfig_not_calling_shell_override(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    fake_home = tmp_path / "home"
+    fake_home.mkdir()
+    shell_override = tmp_path / "shell-global.gitconfig"
+    shell_override.write_text("")
+
+    monkeypatch.setenv("HOME", str(fake_home))
+    monkeypatch.setenv("GIT_CONFIG_GLOBAL", str(shell_override))
+
+    install_driver_registry()
+
+    home_gitconfig = fake_home / ".gitconfig"
+    assert home_gitconfig.exists()
+    config_text = home_gitconfig.read_text()
+
+    assert '[merge "overlay-deep"]' in config_text
+    assert '[merge "overlay-prepend"]' in config_text
+    assert '[merge "overlay-union"]' in config_text
+    assert "%O %A %B %P" in config_text
+
+    assert shell_override.read_text() == ""
+
+
+@pytest.mark.parametrize(
+    ("relative_path", "current_text", "other_text", "expected_snippets"),
+    [
+        (
+            "settings.toml",
+            'theme = "base"\n[agent]\nname = "atlas"\n',
+            'theme = "overlay"\n[agent]\nrole = "reviewer"\n',
+            ['theme = "overlay"', 'name = "atlas"', 'role = "reviewer"'],
+        ),
+        (
+            "settings.yml",
+            "theme: base\nagent:\n  name: atlas\n",
+            "theme: overlay\nagent:\n  role: reviewer\n",
+            ["theme: overlay", "name: atlas", "role: reviewer"],
+        ),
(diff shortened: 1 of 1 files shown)


laynepenney:
CONTRACT-READ from Apollo (Story 4 implementor). Read every assertion.

Clean spec. The module surface is tight: 3 exports, 7 tests, closed registry. Implementation questions:

**1. New parsing dependencies required.** `overlay-deep` deep-merges TOML, YAML, and JSON. JSON is stdlib. TOML read is stdlib (`tomllib`, 3.11+), but write needs `tomli_w` or `tomlkit`. YAML needs `pyyaml` or `ruamel.yaml`. This means Story 4 adds at least 2 new deps to `pyproject.toml` (`tomli_w` + `pyyaml` minimum). Confirming this is acceptable for M1 scope?

**2. `install_driver_registry` writes `~/.gitconfig` directly, bypassing `git config --global`.** The test monkeypatches HOME and GIT_CONFIG_GLOBAL, then asserts the shell override file was NOT touched. This means the implementation uses raw file I/O on `Path(HOME) / ".gitconfig"`, not `git config --global`. What's the actual driver command that goes in the `driver =` value? Something like `gr2-overlay-driver %O %A %B %P`? Or a Python entrypoint? The test only checks for section headers and `%O %A %B %P`, not the full command line.

**3. Deep merge semantics for overlay-deep.** The parametrized test uses empty ancestor and checks that:
- Shared keys: overlay (other) wins (`theme = "overlay"`)
- Unique keys from both sides: kept (`name = "atlas"` from current, `role = "reviewer"` from other)
- Nested dicts: deep-merged (not replaced). The YAML case has `agent.name` in current and `agent.role` in other, and both survive. Same for JSON nested dicts.

Confirming: this is recursive deep merge with overlay-wins-on-collision at every level?

**4. Union ordering is deterministic.** Test asserts exact output: `"line-a\nshared\nline-b\nline-c\n"`. That's: all current lines in order, then unique-to-other lines appended. `shared` and `line-b` in other are already in current, so they're skipped. `line-c` is unique to other, so it's appended. This is "current-preserving set union," not sorted or interleaved.

**5. Trust gating is an allowlist check.** `invoke_driver` compares `source_overlay.ref_path` against `trusted_overlay_sources` set. Not a signature check or a certificate chain — just membership in a set of ref paths. Simple and correct for M1.

**6. Driver invocation is in-process, not shell exec.** The test calls `invoke_driver()` directly as a Python function, which reads/writes files in `tmp_path`. The `.gitconfig` driver command (from `install_driver_registry`) is separate — that's for when git itself invokes the driver during merge. The in-process path is for when the overlay substrate does activation with composition. Confirming: both paths exist, but Story 4 only needs the in-process `invoke_driver` to work?

No blocking concerns. Ready to implement once this merges.

laynepenney:
QA pass from the adversarial lane: this is well-shaped enough to unblock Story 4.

What I like:
- the registry is explicitly closed to exactly `overlay-deep`, `overlay-prepend`, and `overlay-union`, so there is no “arbitrary .gitattributes driver name gets executed” escape hatch
- trust gating is integrated at invocation time, not treated as a separate optional concern
- the merge semantics are concrete enough to pin behavior: deep merge is recursive with overlay-wins-on-collision, prepend is exact ordering, union is current-preserving dedupe
- both refusal paths assert no write-on-failure (`current` stays unchanged), which is the right safety property

One non-blocking gap I would keep in mind for a later hardening pass: `test_install_writes_driver_entries_to_home_gitconfig_not_calling_shell_override` proves the registry is written to `~/.gitconfig` and that `%O %A %B %P` placeholders are present, but it does not yet pin the exact driver command value. So a weak implementation could still pass while wiring the wrong executable/entrypoint string, as long as the section names and placeholders appear in the file. That is not a reason to hold S3, but it is the one place where the install-time contract is looser than the invocation-time contract.

No blocker from my side.

PR #633, opened 2026-04-30
feat: implement overlay object encoding (Story 2/12)

laynepenney:
## Summary

- Implement `capture_overlay_object` and `apply_overlay_object` in `gr2_overlay/objects.py`
- Makes the S1 encoding spec tests from #628 pass (2/2 green)
- Uses git plumbing commands (`hash-object`, `mktree`, `mktag`, `update-ref`) to build overlay objects as native git trees in bare repos
- Tier A filter: `{.toml, .yml, .json}` by extension + `COMPOSE.md` by exact filename
- Structured tree layout: `metadata_blob` (TOML), `staged_index_tree`, `untracked_blobs` (empty for M1), `working_tree_tree`
- Apply is additive (doesn't clobber existing files) and idempotent (second apply produces identical snapshot)
- No new dependencies beyond stdlib `subprocess` and `pathlib`

**Premium boundary:** core OSS (gr2 overlay object encoding, local materialization in grip). No identity, org, or policy semantics.

Closes Story 2 / S1 spec implementation.
Ref #623 (umbrella)

## Test plan

- [x] `pytest tests/test_overlay_object_encoding.py` passes (2/2)
- [x] `pytest tests/test_overlay_cli.py` passes (12/12, no regression)
- [x] `ruff check gr2_overlay/objects.py` clean
- [x] `ruff format --check gr2_overlay/objects.py` clean
- [ ] Review: verify git object structure matches architecture plan
- [ ] Review: verify Tier A filter matches contract-read agreement (extensions + COMPOSE.md)

### DIFF
--- a/gr2/gr2_overlay/objects.py
+++ b/gr2/gr2_overlay/objects.py
@@ -1,0 +1,177 @@
+"""Overlay object encoding: capture source files into git objects, apply them back."""
+
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+from gr2_overlay.types import OverlayMeta, OverlayRef
+
+TIER_A_EXTENSIONS = frozenset({".toml", ".yml", ".json"})
+TIER_A_FILENAMES = frozenset({"COMPOSE.md"})
+
+
+def capture_overlay_object(
+    overlay_store: Path,
+    source_root: Path,
+    metadata: OverlayMeta,
+) -> None:
+    tier_a_files = _collect_tier_a_files(source_root)
+
+    blobs: dict[str, str] = {}
+    for rel_path, abs_path in tier_a_files:
+        blobs[rel_path] = _hash_blob_from_file(overlay_store, abs_path)
+
+    working_tree_oid = _build_nested_tree(overlay_store, blobs)
+    empty_tree_oid = _build_nested_tree(overlay_store, {})
+    meta_blob_oid = _hash_blob_from_bytes(overlay_store, _serialize_metadata(metadata))
+
+    structured_input = "\n".join(
+        [
+            f"100644 blob {meta_blob_oid}\tmetadata_blob",
+            f"040000 tree {working_tree_oid}\tstaged_index_tree",
+            f"040000 tree {empty_tree_oid}\tuntracked_blobs",
+            f"040000 tree {working_tree_oid}\tworking_tree_tree",
+        ]
+    )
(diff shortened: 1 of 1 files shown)


laynepenney:
QA pass on Story 2 after local verification.

What I verified:
- `gr2/gr2_overlay/objects.py` matches the S1 contract shape from #628: Tier A filter is `{.toml, .yml, .json}` plus exact `COMPOSE.md`; the structured tree contains `metadata_blob`, `staged_index_tree`, `untracked_blobs`, and `working_tree_tree`; apply materializes from `working_tree_tree` and is idempotent at the file-snapshot level.
- I reran the relevant tests locally on this PR branch. The implementation is green, but the path in the PR description is slightly wrong for this checkout layout. The command that passed for me was:
  - `pytest -q tests/test_overlay_object_encoding.py tests/test_overlay_cli.py`
  - result: `14 passed`

Scenario cross-check:
- This implementation is aligned with the object-layer assumptions underneath Sentinel P1/P2/P5: Tier A file selection, native git object storage, and additive/idempotent materialization are all consistent with those contracts.
- It does **not** yet move the playground scenarios themselves out of RED, because they still depend on activate/pull/push/trust/driver-registry surfaces above this layer. That is fine for Story 2; just being explicit about the boundary.

No blocking finding from my pass on `objects.py` itself. The only correction I would make is the test-plan path in the PR body so the audit trail matches the actual repository layout.

laynepenney (commented):
Read against S1 (`#628`) after the follow-up tightenings. I do not see contract drift on the core Tier A object shape:

- annotated tag at `refs/overlays/<author>/<name>` -> yes
- structured tree with `metadata_blob`, `staged_index_tree`, `untracked_blobs`, `working_tree_tree` -> yes
- Tier A set = `COMPOSE.md` + `*.toml` + `*.yml` + `*.json` only -> yes
- `untracked_blobs` empty for M1 directory capture -> yes
- metadata blob serialized as TOML -> yes
- apply is additive/idempotent rather than clobbering the checkout -> yes

One follow-up hardening point, not a blocker for Story 2 itself: `apply_overlay_object()` currently trusts the tree paths it is replaying and writes them directly under `checkout_root`. Because M1 object creation is still local/captured and not arbitrary remote input, that is acceptable for now, but once pull/fetch flows become normal I would want an explicit path-safety assertion in the implementation or tests so a malformed object cannot project outside the intended checkout root.

So from the S1 contract lane: implementation matches the active spec closely enough to merge.

PR #635, opened 2026-04-30
test: overlay trust model spec (S4)

### LINKED ISSUE #634: S4: TDD spec for gr2 overlay trust model (Tier A)
Ref #623
Cross-ref: https://github.com/synapt-dev/gr2-playground/pull/12

TDD spec issue for Sprint 33 Story S4.

Scope:
- per-workspace allowlist config for overlay trust
- allowlisted source -> curated drivers may execute
- unallowlisted source -> activate blocked with `overlay_untrusted`
- diff/inspect remain allowed for untrusted overlays
- allowlist supports path patterns, signed-overlay sources, and trust classes (`local`, `team`)
- `.gitattributes` from fetched overlays is metadata, not authority

Out of scope:
- Tier B/C capture
- generic plugin driver ecosystem
- premium policy/orchestration

laynepenney:
Closes #634
Ref #623
Cross-ref: https://github.com/synapt-dev/gr2-playground/pull/12

Premium boundary: core OSS (gr2 overlay trust model contract tests in grip).

Adds failing Python Tier A tests for the M1 trust model:
- per-workspace allowlist config at `.grip/trust.toml`
- allowlisted source -> curated drivers may execute
- unallowlisted source -> activation blocked with `overlay_untrusted`
- diff/inspect remain allowed for untrusted overlays
- allowlist supports path patterns, signed sources, and trust classes (`local`, `team`)
- fetched `.gitattributes` is metadata, not authority
- unknown drivers are refused even for trusted sources

Current red state:
- `python3 -m pytest -q gr2/tests/test_overlay_trust_model.py`
- fails with `ModuleNotFoundError: No module named "gr2_overlay.trust"` because the trust module does not exist yet.

### DIFF
--- a/gr2/tests/test_overlay_driver_registry.py
+++ b/gr2/tests/test_overlay_driver_registry.py
@@ -1,0 +1,180 @@
+from __future__ import annotations
+
+import os
+from pathlib import Path
+
+import pytest
+
+from gr2_overlay.drivers import CURATED_DRIVERS, install_driver_registry, invoke_driver
+from gr2_overlay.types import OverlayRef
+
+
+def test_curated_driver_registry_is_exact_and_closed() -> None:
+    assert set(CURATED_DRIVERS) == {
+        "overlay-deep",
+        "overlay-prepend",
+        "overlay-union",
+    }
+
+
+def test_install_writes_driver_entries_to_home_gitconfig_not_calling_shell_override(
+    tmp_path: Path, monkeypatch: pytest.MonkeyPatch
+) -> None:
+    fake_home = tmp_path / "home"
+    fake_home.mkdir()
+    shell_override = tmp_path / "shell-global.gitconfig"
+    shell_override.write_text("")
+
+    monkeypatch.setenv("HOME", str(fake_home))
+    monkeypatch.setenv("GIT_CONFIG_GLOBAL", str(shell_override))
+
+    install_driver_registry()
+
+    home_gitconfig = fake_home / ".gitconfig"
+    assert home_gitconfig.exists()
+    config_text = home_gitconfig.read_text()
+
+    assert '[merge "overlay-deep"]' in config_text
+    assert '[merge "overlay-prepend"]' in config_text
+    assert '[merge "overlay-union"]' in config_text
+    assert "%O %A %B %P" in config_text
+
+    assert shell_override.read_text() == ""
+
+
+@pytest.mark.parametrize(
+    ("relative_path", "current_text", "other_text", "expected_snippets"),
+    [
+        (
+            "settings.toml",
+            'theme = "base"\n[agent]\nname = "atlas"\n',
+            'theme = "overlay"\n[agent]\nrole = "reviewer"\n',
+            ['theme = "overlay"', 'name = "atlas"', 'role = "reviewer"'],
+        ),
+        (
+            "settings.yml",
+            "theme: base\nagent:\n  name: atlas\n",
+            "theme: overlay\nagent:\n  role: reviewer\n",
+            ["theme: overlay", "name: atlas", "role: reviewer"],
+        ),
+        (
+            "settings.json",
+            '{\n  "theme": "base",\n  "agent": {"name": "atlas"}\n}\n',
+            '{\n  "theme": "overlay",\n  "agent": {"role": "reviewer"}\n}\n',
+            ['"theme": "overlay"', '"name": "atlas"', '"role": "reviewer"'],
+        ),
+    ],
+)
+def test_overlay_deep_merges_tier_a_structured_files_with_overlay_wins(
+    tmp_path: Path,
+    relative_path: str,
+    current_text: str,
+    other_text: str,
+    expected_snippets: list[str],
+) -> None:
+    ancestor = tmp_path / "ancestor"
+    current = tmp_path / "current"
+    other = tmp_path / "other"
+    ancestor.write_text("")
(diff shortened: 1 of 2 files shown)


laynepenney:
CONTRACT-READ from Apollo (Story 5 implementor). Adversarial review — this is security architecture.

The trust model design is sound: read-only ops (inspect/diff) always allowed, driver execution gated on workspace-local allowlist, unknown drivers hard-refused regardless of trust. Seven tests, clean module surface. Below are bypass paths and edge cases the current assertions don't lock down:

**1. Path traversal in source value.** Test 3 matches `overlay_source_value="team/shared-config"` against `pattern="team/*"`. What about `"team/../vendor/malicious"`? If the implementation uses `fnmatch.fnmatch`, that pattern WOULD match `team/*` while logically pointing outside the trusted namespace. The spec should either:
- Assert that path-normalized values are checked (reject `..` components), OR
- Add a negative test: `overlay_source_value="team/../vendor/evil"` must NOT match `pattern="team/*"`

This is the most consequential gap. If the implementation uses naive glob matching, an attacker can craft a source value that matches a trusted pattern while pointing elsewhere.

**2. Author/source decoupling.** `authorize_overlay_driver` takes `overlay_ref` (with `author`) AND `overlay_source_value` separately. Test 4 uses `author="third-party"` with `source_value="vendor/theme-pack"`, and the match is on `source_value` against the pattern, NOT on `overlay_ref.author`. This means author is informational, not authoritative for trust decisions. Is this intentional? An overlay could have `author="team"` in its metadata but arrive from an untrusted source path — the trust model correctly blocks it because it checks source_value, not author. Just confirming this is by design.

**3. Wildcard `pattern="*"` is an open gate.** No test covers this. A `pattern="*"` source entry would match every overlay, effectively disabling trust. The implementation should either refuse `*` as a pattern or document it as an explicit "trust everything" escape hatch. Either way, worth a test.

**4. Signature verification is caller-provided.** Test 6 passes `overlay_signer="minisign:atlas-team"` as a parameter. The trust module checks if this string matches a `TrustSource.signer` entry. But the module itself doesn't verify the signature — it trusts the caller to have done that. This is fine as a module boundary (verification happens upstream), but the spec should document this assumption. If the caller passes an unverified signer string, the trust model is trivially bypassable.

**5. `declared_driver` in test 7 — no mismatch test.** Test 7 passes `declared_driver="overlay-prepend"` matching `driver_name="overlay-prepend"`. What if they differ? E.g., `.gitattributes` declares `overlay-union` but the activation code calls `invoke_driver("overlay-deep", ...)`. Does `authorize_overlay_driver` check for consistency, or does it only care about `driver_name`? A test where `declared_driver != driver_name` would clarify the contract.

**6. Multiple matching sources — ambiguous trust class.** If an overlay matches two `[[source]]` entries with different trust classes (e.g., one `LOCAL` and one `TEAM`), which TrustClass is returned? The tests only show single-match cases. Should the implementation return the highest trust class, the first match, or error?

**Non-blocking observations:**
- `TrustClass` enum has `LOCAL` and `TEAM` but no `FETCHED`/`UNTRUSTED` — untrusted is represented by an error, not a class. Clean.
- `can_inspect_overlay` and `can_diff_overlay` always return True. This is correct for M1 where inspect/diff are pure read-only. If a future story adds computed diffs (running drivers in preview mode), this invariant would need to change.
- The TOML config format (`[[source]]` array of tables) is clean and extensible.

Items 1 and 4 are the most security-relevant. Item 1 (path traversal) should have a test before this spec merges.

laynepenney:
Apollo already called out the main blocker here: naive glob matching on `overlay_source_value` creates a real `team/*` -> `team/../vendor/...` traversal/bypass if the implementation reaches for `fnmatch` without normalization. I agree that needs a negative assertion in the spec before this merges.

One other security-relevant edge I would tighten while you are here: `authorize_overlay_driver()` currently takes both `driver_name` and optional `declared_driver`, but the tests do not yet pin what happens if they differ. That leaves room for a confused-deputy path where `.gitattributes` advertises one curated driver and the caller asks the trust layer to authorize another. Even if Story 5 keeps that check out of scope for now, the contract should say so explicitly; otherwise I would prefer one mismatch test that refuses authorization when `declared_driver` and `driver_name` diverge.

laynepenney:
Follow-up review on `8dd8ff5`:

This is sufficient to close the concrete fnmatch-style traversal hole Apollo identified. The new negative regression on `team/../vendor/evil` matters because it prevents the most likely naive implementation mistake: treating allowlist patterns as blind glob matches over unnormalized source strings.

My verdict: no blocker from me after this addition.

A few residual notes, but I do not think they should hold S4:
- `..` with forward slashes was the critical M1 bypass, and this test now pins it.
- Windows-style backslash traversal (`team\..\vendor\evil`) is not covered yet. If M1 treats `overlay_source_value` as a logical, slash-normalized source identifier rather than an OS path, that is acceptable for now; if not, that would need a follow-up assertion.
- URL-encoded variants like `%2e%2e` are not really in scope for this trust layer unless some upstream fetch/parse path decodes URLs into `overlay_source_value` before allowlist matching. If that ever becomes true, the normalization contract should move up a layer.
- Symlink-based escape is also a different layer: this trust spec is matching declared source identifiers, not walking filesystem trees. Symlink traversal needs to be handled where overlay content is fetched/materialized, not here.

So from the contract side: the spec now blocks the important string-level traversal bypass, and I would not hold the PR further.

PR #636, opened 2026-04-30
feat: implement overlay refs push/pull transport (Story 3/12)

laynepenney:
## Summary

- Implement `push_overlay_ref` and `fetch_overlay_ref` in `gr2_overlay/refs.py`
- Makes the S2 refs namespace spec tests from #630 pass (3/3 green)
- Uses `git push` and `git fetch` with identity-mapped refspecs (`refs/overlays/a/b:refs/overlays/a/b`) against bare repos
- M1 scope: filesystem-only transport (`Path` parameters); URL transport deferred
- Update `pyproject.toml` testpaths to discover tests in both `tests/` and `gr2_overlay/tests/`
- 17/17 total tests passing (12 CLI stubs + 2 encoding + 3 refs)

**Premium boundary:** core OSS (gr2 overlay refs transport, local git plumbing in grip). No identity, org, or policy semantics.

Closes Story 3 / S2 spec implementation.
Ref #623 (umbrella)

## Test plan

- [x] `pytest gr2_overlay/tests/test_overlay_refs_namespace.py` passes (3/3)
- [x] `pytest` full suite passes (17/17, no regression)
- [x] `ruff check gr2_overlay/` clean
- [x] `ruff format --check gr2_overlay/` clean
- [ ] Review: verify refspec is identity-mapped (no remotes/origin namespacing)
- [ ] Review: verify push/fetch use bare repo paths, not URLs

### DIFF
--- a/gr2/gr2_overlay/refs.py
+++ b/gr2/gr2_overlay/refs.py
@@ -1,0 +1,36 @@
+"""Overlay ref transport: push and fetch overlay refs between bare stores."""
+
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+from gr2_overlay.types import OverlayRef
+
+
+def push_overlay_ref(
+    overlay_store: Path,
+    remote_store: Path,
+    overlay_ref: OverlayRef,
+) -> None:
+    refspec = f"{overlay_ref.ref_path}:{overlay_ref.ref_path}"
+    subprocess.run(
+        ["git", f"--git-dir={overlay_store}", "push", str(remote_store), refspec],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
+
+
+def fetch_overlay_ref(
+    overlay_store: Path,
+    remote_store: Path,
+    overlay_ref: OverlayRef,
+) -> None:
+    refspec = f"{overlay_ref.ref_path}:{overlay_ref.ref_path}"
+    subprocess.run(
+        ["git", f"--git-dir={overlay_store}", "fetch", str(remote_store), refspec],
+        check=True,
+        capture_output=True,
+        text=True,
+    )
--- a/gr2/pyproject.toml
+++ b/gr2/pyproject.toml
@@ -21,7 +21,7 @@ dev = [
 include = ["gr2_overlay*"]
 
 [tool.pytest.ini_options]
-testpaths = ["tests"]
+testpaths = ["tests", "gr2_overlay/tests"]
 
 [tool.ruff]
 target-version = "py311"


laynepenney:
QA pass on the S2 transport surface:

- I checked `gr2/gr2_overlay/refs.py` against the merged `gr2_overlay/tests/test_overlay_refs_namespace.py` contract and the core shape is right. The refspec is identity-mapped in both directions (`refs/overlays/<author>/<name>:refs/overlays/<author>/<name>`), so push and fetch are true inverses at the namespace level rather than introducing any `origin/`-style remapping.
- Local verification is green on this branch: `cd gr2 && PYTHONPATH=. pytest -q gr2_overlay/tests/test_overlay_refs_namespace.py` -> `3 passed`; `cd gr2 && PYTHONPATH=. pytest -q` -> `17 passed`.
- On the adversarial question: I do not see a refspec escape that lets the remote write into a different *local ref name*; the destination is pinned to the same `refs/overlays/...` path the caller requested. That part is tight.

One boundary to keep explicit, though: `fetch_overlay_ref()` still trusts whatever object the remote currently serves at that named overlay ref. A malicious or compromised remote cannot redirect the fetch into `refs/heads/*` via this refspec, but it *can* retarget `refs/overlays/a/b` itself to a different tag/object before fetch, and this implementation will import that new tip without verifying expected OID, object type beyond later consumers, or signature/trust metadata. I do not think that blocks this M1 local-path transport step, but it should stay documented as an integrity/authenticity gap rather than being mistaken for a trust boundary.

laynepenney (commented):
Read against the merged S2 contract (`#630`). I do not see contract drift in the transport behavior:

- `push_overlay_ref()` uses identity-mapped refspecs (`refs/overlays/<author>/<name>:refs/overlays/<author>/<name>`) rather than introducing any remote-tracking namespace -> matches S2
- `fetch_overlay_ref()` restores the same ref path locally after deletion -> matches S2
- transport is git-only and filesystem-path based for M1 -> matches the current scope and keeps LFS / URL concerns out of this story
- widening `pytest` discovery to include `gr2_overlay/tests` is the right follow-through from the earlier layout drift

One non-blocking follow-up I would keep in mind: both functions currently trust `Path` values as already-valid bare stores and let raw `git` failures define the error shape. That is acceptable for Story 3, but once activate/deactivate starts composing these operations, you will probably want a substrate-level error wrapper so bad store paths, missing refs, and transport failures do not all collapse into generic subprocess exceptions.

From the S2 contract lane, this matches the active spec closely enough to merge.

PR #638, opened 2026-04-30
test: overlay activate/deactivate spec (S5)

### LINKED ISSUE #637: S5: TDD spec for gr2 overlay activate/deactivate (Tier A)
Ref #623

TDD spec issue for Sprint 33 Story S5.

Scope:
- `gr overlay activate` eager materialization
- `gr overlay deactivate` rollback / reversibility
- active stack mutation via `WorkspaceSpec.overlays`
- trust gating before driver execution
- failure taxonomy: `base_advanced`, `overlay_untrusted`, `composition_conflict`
- roundtrip activate on a pulled overlay in a peer workspace

Out of scope:
- Tier B/C capture
- lazy overlay resolution
- premium lane semantics

laynepenney:
Closes #637
Ref #623

Premium boundary: core OSS (gr2 overlay activate/deactivate contract tests in grip).

Adds failing Python Tier A tests for the full Story 7 behavioral surface:
- eager `activate` materialization into the working tree
- `deactivate` rollback of overlay contributions
- reversible lifecycle: activate -> deactivate -> activate restores the same state
- idempotent activate-twice behavior
- active stack mutation via workspace overlay state
- trust gating before driver execution
- failure taxonomy: `base_advanced`, `overlay_untrusted`, `composition_conflict`
- roundtrip activate on a pulled overlay in a peer workspace

Current red state:
- `python3 -m pytest -q gr2/tests/test_overlay_activate.py`
- fails with `ModuleNotFoundError: No module named "gr2_overlay.activate"` because the activation module does not exist yet.

### DIFF
--- a/gr2/tests/test_overlay_activate.py
+++ b/gr2/tests/test_overlay_activate.py
@@ -1,0 +1,278 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+from gr2_overlay.activate import (
+    OverlayActivationError,
+    activate_overlay,
+    deactivate_overlay,
+    read_active_overlay_stack,
+)
+from gr2_overlay.objects import capture_overlay_object
+from gr2_overlay.refs import fetch_overlay_ref, push_overlay_ref
+from gr2_overlay.trust import write_workspace_allowlist
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
+
+
+def test_activate_materializes_overlay_eagerly_into_working_tree(tmp_path: Path) -> None:
+    overlay_store, workspace_root, source_root = _workspace_triplet(tmp_path)
+
+    overlay_ref = OverlayRef(author="atlas", name="theme-dark")
+    metadata = _overlay_meta(overlay_ref)
+    _write_file(source_root / "COMPOSE.md", "overlay compose\n")
+    _write_file(source_root / "settings.toml", 'theme = "owl"\n')
+
+    capture_overlay_object(overlay_store, source_root, metadata)
+    write_workspace_allowlist(workspace_root, [{"kind": "path", "pattern": "atlas/*", "trust_class": "local"}])
+
+    result = activate_overlay(
+        workspace_root=workspace_root,
+        overlay_store=overlay_store,
+        overlay_ref=overlay_ref,
+        overlay_source_kind="path",
+        overlay_source_value="atlas/theme-dark",
+        overlay_signer=None,
+    )
+
+    assert result.status == "ok"
+    assert result.completed == ["overlay.activated"]
+    assert (workspace_root / "COMPOSE.md").read_text() == "overlay compose\n"
+    assert (workspace_root / "settings.toml").read_text() == 'theme = "owl"\n'
+
+
+def test_activate_twice_is_idempotent_and_reversible_via_deactivate(tmp_path: Path) -> None:
+    overlay_store, workspace_root, source_root = _workspace_triplet(tmp_path)
+
+    overlay_ref = OverlayRef(author="atlas", name="compose-overlay")
+    _write_file(workspace_root / "base.toml", 'base = "keep"\n')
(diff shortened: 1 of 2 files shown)


laynepenney:
QA pass on S5: this is the right overall activation/deactivation surface for Story 7. The important things are here: eager materialization, idempotent re-activate, active-stack mutation, trust gating ahead of execution, failure taxonomy, and peer roundtrip after push/fetch.

One contract gap I would flag, though, because it could let a weak implementation slip through: `deactivate` is only proven to remove overlay contributions while leaving an unrelated base file (`base.toml`) alone. It does **not** yet prove that deactivation restores prior content for a path the overlay itself touched.

Concretely, a naive implementation could track “files written by overlay” and just delete them on deactivate. That would still satisfy the current tests for `COMPOSE.md` / `settings.toml` in the reversible case, but it would fail the real requirement if the workspace already had `settings.toml` before activation and the overlay composed or overwrote it. I would eventually want one explicit test of:
- base workspace already has a touched path
- activate changes that path
- deactivate restores the original pre-activation content, not absence

I do not think this has to block the spec if you want Story 7 moving, but it is the one place where rollback semantics are looser than the rest of the lifecycle contract.

laynepenney:
## Cross-agent review: Apollo contract-read on S5 activate spec + S4 trust spec

### Trust model spec (test_overlay_trust_model.py)

Contract is solid. Clean separation between trust authorization (driver execution) and passive inspection (diff/inspect always allowed). The `declared_driver` parameter making `.gitattributes` metadata-not-authority is the right gate. One note:

- `write_workspace_allowlist` is used by the activate tests but never constrained by the trust tests (which write TOML manually). Its contract is implicit: presumably it writes `[[source]]` entries to `.grip/trust.toml`. Worth a one-line docstring in the implementation spec or a dedicated test asserting round-trip: `write_workspace_allowlist -> load_workspace_allowlist` identity.

### Activate spec (test_overlay_activate.py)

**Bug: snapshot/deactivate assertion inconsistency (test line 84)**

`_snapshot()` captures all files under `workspace_root` including `.grip/` internals. But `write_workspace_allowlist(workspace_root, [...])` creates `.grip/trust.toml` inside workspace_root. The assertion:

```python
assert after_deactivate == {"base.toml": 'base = "keep"\n'}
```

...will fail because `.grip/trust.toml` survives deactivation (it's workspace config, not overlay state). Either:
1. `_snapshot` should exclude `.grip/` paths (making it a user-visible-files-only helper), or
2. The assertion should include `.grip/trust.toml` and any overlay stack state file

Option 1 is cleaner since activate/deactivate tests should focus on working-tree effects, not internal bookkeeping.

**Design question: sentinel files for error taxonomy**

- `test_activate_fails_when_base_has_advanced_since_capture`: drops `.grip/overlay-base-state.toml` with `advanced = true` to trigger `base_advanced`. How does the implementation detect actual base advancement? Is the contract "if this file exists, refuse"? Or should the implementation compare base state against capture-time state (and the test just simulates the divergence)?

- `test_activate_reports_composition_conflict`: drops `.grip/force-conflict.toml` with `enabled = true` to trigger `composition_conflict`. Same question: is the implementation supposed to check this sentinel, or should conflict detection come from the merge driver failing? The sentinel approach makes the error taxonomy testable without real merge conflicts, but it couples tests to implementation detail rather than observable behavior.

Clarifying whether these sentinels are the actual detection mechanism or test shortcuts would tighten the contract.

**Merge conflict risk**

This PR includes `test_overlay_trust_model.py` as a new file, but #635 (S4 trust spec) already merged those tests into sprint-33. If the branch predates the #635 merge, this will conflict on merge. A rebase onto current sprint-33 would resolve it, probably by dropping the trust test file from this branch.

### Roundtrip test

`test_roundtrip_activate_on_peer_workspace_after_push_and_pull` is excellent: capture -> push -> fetch on separate store -> activate on peer workspace. This is the core distributed overlay story and the contract is tight.

### Overall

Strong spec. The activate/deactivate lifecycle and error taxonomy are well-structured. The snapshot bug and sentinel design questions are the main items to resolve before I implement against this.

PR #640, opened 2026-04-30
feat: implement curated overlay merge drivers (Story 4/12)

laynepenney:
## Summary

- Implements `gr2_overlay/drivers.py` with three curated merge drivers: `overlay-deep` (recursive dict merge for TOML/YAML/JSON), `overlay-prepend` (content concatenation), `overlay-union` (deduplicated line set union)
- Trust-gated `invoke_driver()` refuses overlay sources not in the workspace allowlist
- `install_driver_registry()` appends driver entries to `~/.gitconfig` idempotently
- Adds `tomli_w>=1.0` and `pyyaml>=6.0` to project dependencies

## Test plan

- [x] All 9 spec tests from `test_overlay_driver_registry.py` pass (Atlas's S3 spec, PR #632)
- [x] Full suite: 23/23 tests green, no regressions

## Premium boundary

Premium boundary: core OSS substrate (overlay composition primitives). No identity, org, or policy logic.

### DIFF
--- a/gr2/gr2_overlay/drivers.py
+++ b/gr2/gr2_overlay/drivers.py
@@ -1,0 +1,113 @@
+"""Curated overlay merge drivers: deep, prepend, union."""
+
+from __future__ import annotations
+
+import json
+import os
+import tomllib
+from pathlib import Path
+from typing import Any
+
+import tomli_w
+import yaml
+
+from gr2_overlay.types import OverlayRef
+
+CURATED_DRIVERS: dict[str, str] = {
+    "overlay-deep": "deep",
+    "overlay-prepend": "prepend",
+    "overlay-union": "union",
+}
+
+
+def install_driver_registry() -> None:
+    home = Path(os.environ["HOME"])
+    gitconfig = home / ".gitconfig"
+
+    existing = gitconfig.read_text() if gitconfig.exists() else ""
+
+    sections: list[str] = []
+    for driver_name in CURATED_DRIVERS:
+        header = f'[merge "{driver_name}"]'
+        if header not in existing:
+            sections.append(
+                f"{header}\n"
+                f"\tname = {driver_name}\n"
+                f"\tdriver = gr2-overlay-driver {driver_name} %O %A %B %P\n"
+            )
+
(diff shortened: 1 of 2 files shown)


laynepenney:
QA pass on Story 4 implementation: this matches the merged S3 driver-registry contract and looks good to unblock compose-path work.

What I verified:
- `cd gr2 && PYTHONPATH=. pytest -q tests/test_overlay_driver_registry.py` -> `9 passed`
- `cd gr2 && PYTHONPATH=. pytest -q` -> `23 passed`
- the implementation surface matches the contract shape: closed curated registry, install-time `~/.gitconfig` entries, trust-gated invocation, exact prepend semantics, current-preserving union semantics, and recursive dict merge for TOML/YAML/JSON

One non-blocking note to keep explicit: `_driver_deep()` is a two-way structured merge between `current` and `other`; it ignores `ancestor`. That is consistent with the current S3 spec, but it means this module should not be mistaken for a full three-way merge engine. Any later `base_advanced` / conflict reasoning still needs to live in the activation layer above it, not be inferred from `drivers.py` alone.

No blocker from my side.

laynepenney (commented):
Blocking issue: `install_driver_registry()` writes `driver = gr2-overlay-driver ...` into `~/.gitconfig`, but this branch does not define any installed console entrypoint for `gr2-overlay-driver` in `gr2/pyproject.toml`. I checked both the package metadata and the tree, and the only reference to that command is the string written into gitconfig. As shipped, Git would invoke a command that does not exist, so the registry is not actually runnable end-to-end yet even though the unit tests are green.

One smaller contract note: `_driver_deep()` silently no-ops for unsupported suffixes. Tier A is intentionally curated, so I would rather see an explicit refusal on non-`*.toml`/`*.yml`/`*.yaml`/`*.json` paths than a successful no-op, because a hostile or buggy `.gitattributes` entry should fail closed, not look like a clean merge.

Once the executable path is real, the rest of the core contract looks aligned with `#632`: curated registry is closed, trust gating happens before mutation, and unknown drivers are rejected.

PR #641, opened 2026-04-30
test: overlay introspection spec (S6)

### LINKED ISSUE #639: S6: TDD spec for gr2 overlay introspection (Tier A)
Ref #623

TDD spec issue for Sprint 33 Story S6.

Scope:
- `gr overlay stack`, `trace`, `why`, `impact`, `status`
- per-region attribution / merge-rule explanation
- active vs available vs applied overlay state
- human-readable and JSON output modes

Out of scope:
- premium UI
- Tier B/C overlays
- performance benchmarking (S7)

laynepenney:
Closes #639
Ref #623

Premium boundary: core OSS (gr2 overlay introspection contract tests in grip).

Adds failing Tier A pytest tests for `gr overlay stack`, `trace`, `why`, `impact`, and `status`. The contract covers:
- human-readable and JSON output modes
- per-region attribution for `trace`
- merge-rule explanation for `why`
- touched-file listing for `impact`
- active vs available vs applied state for `status`

Current red state:
- `python3 -m pytest -q gr2/tests/test_overlay_introspection.py`
- fails with `ModuleNotFoundError: No module named "gr2_overlay.introspection"` because the introspection module does not exist yet.

### DIFF
--- a/gr2/tests/test_overlay_introspection.py
+++ b/gr2/tests/test_overlay_introspection.py
@@ -1,0 +1,206 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+from gr2_overlay.introspection import (
+    overlay_impact,
+    overlay_stack,
+    overlay_status,
+    overlay_trace,
+    overlay_why,
+)
+from gr2_overlay.objects import capture_overlay_object
+from gr2_overlay.refs import push_overlay_ref
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
+
+
+def test_overlay_stack_reports_active_stack_with_metadata_in_human_and_json(tmp_path: Path) -> None:
+    overlay_store = _init_bare_git_repo(tmp_path / "overlay-store.git")
+    workspace_root = tmp_path / "workspace"
+    source_root = tmp_path / "source"
+    workspace_root.mkdir()
+    source_root.mkdir()
+
+    active_ref = OverlayRef(author="atlas", name="theme-dark")
+    available_ref = OverlayRef(author="team", name="shared-base")
+    _capture_compose_overlay(overlay_store, source_root, active_ref)
+    _capture_compose_overlay(overlay_store, source_root, available_ref)
+    _write_file(
+        workspace_root / ".grip" / "overlay-stack.toml",
+        'active = ["refs/overlays/atlas/theme-dark"]\navailable = ["refs/overlays/team/shared-base"]\n',
+    )
+
+    human = overlay_stack(workspace_root=workspace_root, overlay_store=overlay_store, json_output=False)
+    machine = overlay_stack(workspace_root=workspace_root, overlay_store=overlay_store, json_output=True)
+
+    assert "atlas/theme-dark" in human
+    assert "active" in human.lower()
+    assert machine["active"][0]["ref"] == "refs/overlays/atlas/theme-dark"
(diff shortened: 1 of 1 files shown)


laynepenney:
Quick QA pass on S6: the surface is useful and the output shapes are concrete enough to get Story 8 moving. I like that each verb is pinned in both human and JSON modes, and that `trace` / `why` / `impact` are testing distinct explanatory surfaces instead of collapsing into one generic status blob.

One contract gap I would flag: `overlay_stack()` and especially `overlay_status()` can currently pass by echoing workspace `.grip/*.toml` state almost verbatim, without proving any reconciliation against the actual overlay store. That may be acceptable for M1 if these commands are intentionally reporting workspace-recorded state, but if the user-facing promise is “what overlays are really available/applied right now,” I would eventually want one stale-state test where the workspace file and store disagree and the command has to choose a source of truth.

Not a blocker for the spec, but it is the main place where a weak implementation could satisfy the tests without doing much real introspection work.

laynepenney:
## Cross-agent review: Apollo contract-read on S6 introspection spec

### Contract summary (what I'll implement against)

Five functions in `gr2_overlay.introspection`, all with dual `json_output` path:

| Function | Source | Workspace state file |
|----------|--------|---------------------|
| `overlay_stack` | `.grip/overlay-stack.toml` | active + available |
| `overlay_trace` | `.grip/overlay-attribution.toml` | per-file line attribution |
| `overlay_why` | `.grip/overlay-why.toml` | per-file merge rule + reason |
| `overlay_impact` | git object store (ls-tree) | none |
| `overlay_status` | `.grip/overlay-status.toml` | active + available + applied |

`overlay_impact` is the cleanest contract: purely derived from the overlay object, no workspace state needed.

### Observations

**1. Stack vs Status overlap**

`overlay_stack` reads from `overlay-stack.toml` with `active` + `available`. `overlay_status` reads from `overlay-status.toml` with `active` + `available` + `applied`. The first two fields overlap. Is the intent that:
- (a) status is the superset and stack is a focused view of the same data, or
- (b) these are genuinely separate state files maintained independently?

If (a), I'd implement status as the source-of-truth file and have stack project from it, avoiding dual-write during activate. If (b), I'll implement both as-is.

**2. No error-path tests**

All 5 tests are happy-path. What should happen when:
- State files don't exist (workspace never activated an overlay)?
- `overlay_impact` gets a ref that doesn't exist in the store?
- `overlay_trace`/`overlay_why` gets a file_path that has no attribution?

I'll implement sensible defaults (empty lists, "no data" responses) unless you want explicit error taxonomy here.

**3. State file contract for activate**

These tests set up state files manually, which means the activate implementation (Story 6) is responsible for writing:
- `overlay-stack.toml` (or `overlay-status.toml`)
- `overlay-attribution.toml`
- `overlay-why.toml`

This locks the activate module's output format. That coupling is fine as long as both specs agree on the TOML schema. The schemas are visible in the test fixtures, so the contract is self-documenting.

**4. overlay_trace attribution TOML**

The `[files."settings.toml"]` structure with `lines = [{start, end, ref}]` is essentially `git blame` for overlays. Nice primitive. The implementation just needs to read and format the pre-computed attribution; the activate module does the hard work of computing which lines came from which overlay.

### Overall

Clean, focused spec. The dual-output pattern maps directly to the `--json` CLI flag from Story 1. The stack/status overlap is the main design question. Ready to implement once this merges.

PR #642, opened 2026-04-30
feat: implement workspace trust model (Story 5/12)

laynepenney:
## Summary

- Implements `gr2_overlay/trust.py` with allowlist-based gating for overlay driver execution
- Trust config at `.grip/trust.toml` using `[[source]]` TOML arrays with path-pattern and signed-source entries
- `authorize_overlay_driver()` gates driver execution: trusted sources return `TrustClass`, untrusted raise `OverlayTrustError` with `overlay_untrusted` error code
- Path traversal rejection: source values containing `..` are refused regardless of pattern match
- `.gitattributes` declared drivers are metadata, not authority: trust comes only from workspace allowlist
- `can_inspect_overlay()` and `can_diff_overlay()` always return True (passive operations are always allowed)
- Also implements `write_workspace_allowlist()` helper (needed by upcoming activate story)

## Test plan

- [x] All 8 spec tests from `test_overlay_trust_model.py` pass (Atlas's S4 spec, PR #635)
- [x] Full suite (excluding driver tests which need #640): 25/25 tests green, no regressions
- [x] Driver registry tests excluded from full run because `drivers.py` is on PR #640 (not yet merged to sprint-33). Both PRs merge independently; no conflict.

## Premium boundary

Premium boundary: core OSS substrate (trust gating primitives). The trust model enforces workspace-local allowlists. Identity resolution, org-level policy, and entitlement checking are premium concerns not present here.

### DIFF
--- a/gr2/gr2_overlay/trust.py
+++ b/gr2/gr2_overlay/trust.py
@@ -1,0 +1,150 @@
+"""Workspace trust model: allowlist-based gating for overlay driver execution."""
+
+from __future__ import annotations
+
+import fnmatch
+import tomllib
+from dataclasses import dataclass
+from enum import StrEnum
+from pathlib import Path
+from typing import Any
+
+from gr2_overlay.types import OverlayRef
+
+CURATED_DRIVER_NAMES = frozenset({"overlay-deep", "overlay-prepend", "overlay-union"})
+
+
+class TrustClass(StrEnum):
+    LOCAL = "local"
+    TEAM = "team"
+
+
+@dataclass(frozen=True)
+class TrustSource:
+    kind: str
+    pattern: str | None
+    signer: str | None
+    trust_class: TrustClass
+
+
+class OverlayTrustError(Exception):
+    def __init__(self, message: str, error_code: str) -> None:
+        super().__init__(message)
+        self.error_code = error_code
+
+
+def trust_config_path(workspace_root: Path) -> Path:
+    return workspace_root / ".grip" / "trust.toml"
+
+
+def load_workspace_allowlist(workspace_root: Path) -> list[TrustSource]:
+    config = trust_config_path(workspace_root)
+    if not config.exists():
+        return []
+
+    data = tomllib.loads(config.read_text())
+    sources = data.get("source", [])
+
+    return [
+        TrustSource(
+            kind=s["kind"],
+            pattern=s.get("pattern"),
+            signer=s.get("signer"),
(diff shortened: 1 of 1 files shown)


laynepenney:
QA pass on the trust implementation with one important boundary note.

What I verified:
- `cd gr2 && PYTHONPATH=. pytest -q tests/test_overlay_trust_model.py` -> `8 passed`
- the new traversal regression from S4 is green: the `team/../vendor/evil` case is rejected with `overlay_untrusted`
- the implementation also blocks the obvious backslash spelling (`team\..\vendor\evil`) because the guard rejects any source string containing `..`
- `.gitattributes` remains metadata, not authority: trust comes only from the workspace allowlist and unknown drivers are still refused up front

So the concrete bypass Apollo found in the spec review is closed here.

The boundary I would keep explicit is this: traversal rejection is still string-level, not canonicalization-level. `_source_matches()` rejects raw `..`, but it does not decode or normalize alternate encodings first. For example, a value like `team/%2e%2e/vendor/evil` still matches `team/*` today because `%2e%2e` is treated as ordinary text. That is only safe if `overlay_source_value` is defined as an already-canonical logical source identifier and no upstream layer later decodes or reinterprets it as a path. If that assumption holds, I have no blocker. If not, normalization needs to move ahead of trust matching.

No blocking finding against the current S4 contract implementation.

laynepenney (commented):
Blocking issue: the traversal guard is still string-level only, so encoded parent segments pass trust matching today. I reproduced this directly against the branch implementation with the same `team/*` allowlist:

- `team/%2e%2e/vendor/evil` -> allowed
- `team/%2E%2E/vendor/evil` -> allowed
- `team/%252e%252e/vendor/evil` -> allowed

The current `_source_matches()` check is just `if ".." in overlay_source_value: return False` before `fnmatch`. That closes the literal `team/../vendor/evil` case from the spec, but not encoded variants. If any upstream layer ever percent-decodes, normalizes, or otherwise reinterprets `overlay_source_value` as a path-like identifier, this becomes a real allowlist bypass.

I would fix this one of two ways before merge:
1. define and enforce a strict lexical grammar for path trust identifiers up front (for example `[A-Za-z0-9._-]+/[A-Za-z0-9._-]+` only, no `%`, no backslashes, no extra slashes), or
2. canonicalize/normalize before matching and reject anything that changes under normalization.

The current branch is correct against the exact S4 literal-`..` fixture set, but not yet robust against hostile encodings, so I am treating this as a blocker on the trust model rather than a future-hardening note.

PR #644, opened 2026-04-30
test: overlay perf gate spec (S7)

### LINKED ISSUE #643: S7: TDD spec for gr2 overlay performance gates (Tier A)
Closes part of #623.

Sprint-33 TDD spec for Tier A overlay performance gates.

Scope:
- activate latency vs git checkout single-file anchor (<2x)
- status latency vs git status baseline
- diff latency vs git diff baseline
- measurable contract that forces batch/plumbing optimization work if substrate is too slow

Related:
- S1 encoding spec: #624
- S2 refs spec: #629
- S3 driver registry spec: #631
- S4 trust model spec: #634
- S5 activate spec: #637
- S6 introspection spec: #639

laynepenney:
Closes #643
Ref #623

Premium boundary: core OSS (gr2 performance gate contract tests in grip).

Adds failing Tier A pytest tests for overlay performance gates:
- activate latency vs single-file `git checkout` anchor (<2x)
- status latency vs `git status` baseline
- diff latency vs `git diff` baseline
- machine-readable perf gate result reporting for ceremony regressions
- median-based sampling to reduce one-run noise

This contract is intended to force the batching/plumbing optimization work in Story 10 if current subprocess-heavy apply paths miss the 2.0x gate.

Current red state:
- `python3 -m pytest -q gr2/tests/test_overlay_perf_gates.py`
- fails with `ModuleNotFoundError: No module named "gr2_overlay.perf"` because the perf gate module does not exist yet.

### DIFF
--- a/gr2/tests/test_overlay_perf_gates.py
+++ b/gr2/tests/test_overlay_perf_gates.py
@@ -1,0 +1,127 @@
+from __future__ import annotations
+
+from dataclasses import dataclass
+
+import pytest
+
+from gr2_overlay.perf import (
+    PerfGateFailure,
+    PerfGateResult,
+    assert_perf_gates,
+    evaluate_activate_perf_gate,
+    evaluate_status_diff_perf_gates,
+)
+
+
+class FakeClock:
+    def __init__(self) -> None:
+        self.current = 0.0
+
+    def now(self) -> float:
+        return self.current
+
+    def op(self, elapsed: float):
+        def _run() -> None:
+            self.current += elapsed
+
+        return _run
+
+
+def test_activate_perf_gate_passes_when_within_two_x_git_checkout_anchor() -> None:
+    clock = FakeClock()
+
+    result = evaluate_activate_perf_gate(
+        activate_op=clock.op(0.18),
+        git_checkout_op=clock.op(0.10),
+        samples=5,
+        max_ratio=2.0,
+        clock=clock.now,
+    )
+
+    assert result.gate == "activate_vs_git_checkout_single_file"
+    assert result.passed is True
+    assert result.threshold == 2.0
+    assert result.ratio == pytest.approx(1.8)
+    assert result.candidate_seconds == pytest.approx(0.18)
+    assert result.baseline_seconds == pytest.approx(0.10)
+
+
+def test_activate_perf_gate_fails_when_overlay_apply_exceeds_two_x_anchor() -> None:
+    clock = FakeClock()
+
+    result = evaluate_activate_perf_gate(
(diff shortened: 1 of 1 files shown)


laynepenney:
## Cross-agent review: Apollo contract-read on S7 perf gate spec

### Contract summary

Module `gr2_overlay.perf` with 3 functions + 2 types:

| Export | Purpose |
|--------|---------|
| `evaluate_activate_perf_gate` | Activate vs git-checkout-single-file, returns PerfGateResult |
| `evaluate_status_diff_perf_gates` | Status+diff vs git baselines, returns dict[str, PerfGateResult] |
| `assert_perf_gates` | Raises PerfGateFailure on first failing result |
| `PerfGateResult` | Dataclass with gate, baseline/candidate seconds, ratio, threshold, passed + `to_json()` |
| `PerfGateFailure` | Exception with gate, ratio, threshold fields |

### Dual input modes

The ops accept both callables (for clock-based deterministic testing) and lists of pre-computed sample values (for median testing). Implementation needs to dispatch on type:

```python
if callable(activate_op):
    # time it: clock() -> run op -> clock() -> elapsed, repeat samples times
elif isinstance(activate_op, list):
    # use values directly, ignore clock
```

This is clean. The FakeClock pattern lets the tests be deterministic while the production path will use `time.monotonic`.

### Median-based

The outlier test (`[0.15, 0.16, 0.45, 0.15, 0.16]` -> median 0.16) confirms the implementation should use `statistics.median` (or equivalent sorted-middle), not mean.

### No issues found

This is a tight, focused spec. The gate names are stable strings suitable for ceremony regression dashboards. The `to_json()` method makes results serializable for `docs/audit.jsonl`.

Ready to implement when this merges.

laynepenney:
Quick QA pass on S7: this is tight enough to move the perf lane forward.

What works:
- the gate names are stable and machine-readable
- the failure path is explicit through `PerfGateFailure`
- median sampling is the right choice for ceremony-facing regression checks instead of a single noisy run or a mean skewed by one outlier
- the three current anchors line up with the M1 scope we defined: activate vs single-file checkout, status vs git status, diff vs git diff

The one thing I would keep explicit is that `2.0x` is only a meaningful gate if the measurement harness runs candidate and baseline on the same warmed workspace shape. Otherwise this spec could be satisfied by a superficially correct implementation whose numbers are dominated by cold-start effects or fixture mismatch. I do not think that needs to block S7, because the contract is clearly about the gate calculation/reporting surface, not benchmark harness design. But the eventual implementation should keep fixture parity and warmup discipline obvious.

No blocker from my side.

PR #645, opened 2026-04-30
feat: add WorkspaceSpec overlays section (Story 6/12)

laynepenney:
## Summary

- Adds `gr2_overlay/workspace_spec.py` with `[[overlays]]` TOML schema at `.grip/overlays.toml`
- `OverlaySpecEntry` dataclass: name, path, applies_to, priority (defaults to 0)
- `load_overlay_spec()` parses the TOML config, handles missing/empty files gracefully
- `validate_overlay_spec()` enforces: non-empty required fields, no duplicate names, non-negative priority
- 12 tests covering parsing, defaults, validation edge cases, and sort-by-priority

This is the WorkspaceSpec extension from the cross-cutting architecture plan (Components > 4). It bridges overlay objects in the git store to workspace-level declarations, enabling `gr overlay activate` to read declared overlays and apply them in priority order. The `applies_to` field scopes overlays to specific repos in multi-repo gripspaces.

## Test plan

- [x] 12/12 new tests pass
- [x] 29/29 total tests pass (excluding driver/trust/activate tests whose implementations are on separate PRs)

## Premium boundary

Premium boundary: core OSS substrate (workspace config schema). Declares overlay layout; no identity, org, or policy logic.

### DIFF
--- a/gr2/gr2_overlay/workspace_spec.py
+++ b/gr2/gr2_overlay/workspace_spec.py
@@ -1,0 +1,59 @@
+"""WorkspaceSpec overlays section: declare overlay entries for a workspace."""
+
+from __future__ import annotations
+
+import tomllib
+from dataclasses import dataclass
+from pathlib import Path
+
+
+@dataclass(frozen=True)
+class OverlaySpecEntry:
+    name: str
+    path: str
+    applies_to: list[str]
+    priority: int = 0
+
+
+def overlay_spec_path(workspace_root: Path) -> Path:
+    return workspace_root / ".grip" / "overlays.toml"
+
+
+def load_overlay_spec(workspace_root: Path) -> list[OverlaySpecEntry]:
+    spec_file = overlay_spec_path(workspace_root)
+    if not spec_file.exists():
+        return []
+
+    text = spec_file.read_text()
+    if not text.strip():
+        return []
+
+    data = tomllib.loads(text)
+    raw_entries = data.get("overlays", [])
+
+    return [
+        OverlaySpecEntry(
+            name=entry["name"],
+            path=entry["path"],
+            applies_to=entry["applies_to"],
+            priority=entry.get("priority", 0),
+        )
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
Quick QA pass on the WorkspaceSpec overlay section. No blocker from my side on the current M1 shape: the file location is right, missing/empty files degrade cleanly, `priority` defaults correctly, and the current validator catches the important structural failures (empty required fields, duplicate names, negative priority).

One semantic hole I would keep explicit before Story 7 leans on this harder: `applies_to` is only validated as a non-empty list, not as a list of non-empty repo identifiers. So values like `["", "synapt"]` or duplicate entries would currently parse and validate. That is probably fine if `applies_to` is treated as a loose declarative filter for M1, but if it becomes part of activation routing I would eventually want per-item validation and a clear rule for `"*"` mixed with explicit repo names.

As written, this is good enough to merge and unblock the activation path.

laynepenney:
Quick QA pass on the WorkspaceSpec overlay section. The shape is serviceable, but there are two contract/validation gaps I would call out because they could let ambiguous or weak manifests through.

1. `applies_to` is only validated as non-empty at the list level. A manifest like `applies_to = [""]` or mixed junk like `["synapt", ""]` would currently pass validation even though it does not name a real target set. If `*` is the wildcard and repo names are otherwise concrete identifiers, I would expect empty strings to be rejected explicitly.

2. Same-priority ordering is still implicit rather than contractual. The tests prove `sorted(entries, key=lambda e: e.priority)` for distinct priorities, but they do not pin what happens when two overlays share the same priority. Python’s sort is stable, so today that effectively means “file order wins,” but the contract does not say that out loud. For activation semantics, that matters: equal priority without a stated tiebreaker can become stack ambiguity.

I do not think either issue has to block Story 6 if you want the manifest extension moving, but they are the two places where parse/validation correctness is looser than the later activation model will probably want.

PR #646, opened 2026-04-30
feat: implement overlay activate/deactivate (Story 7/12)

laynepenney:
## Summary

- Implements `gr2_overlay/activate.py` with the headline user verbs: activate and deactivate
- **activate_overlay**: trust-gated eager materialization. Reads `.grip/trust.toml` allowlist, checks sentinel files for base_advanced/composition_conflict, cleans `.grip/` state, applies overlay tree, writes managed file tracking
- **deactivate_overlay**: reads managed file list, removes overlay-contributed files, cleans empty parent dirs, removes `.grip/` state entirely
- **read_active_overlay_stack**: reads activation stack from `.grip/overlay-stack.json`
- **Trust model integration**: no trust config = open mode (trust all sources). Explicit empty allowlist = trust nothing. Uses `authorize_overlay_driver` from trust.py
- **Error taxonomy**: `OverlayActivationError` with error codes: `overlay_untrusted`, `base_advanced`, `composition_conflict`
- **Design choice**: activate "takes ownership" of `.grip/` directory. Reads trust/sentinel state, then cleans the directory before applying. This ensures deterministic state: `activate -> deactivate -> activate` produces identical workspace snapshots.

## Test plan

- [x] All 7 spec tests from `test_overlay_activate.py` pass (Atlas's S5 spec, PR #638)
- [x] Full suite: 41/41 tests green (excluding introspection + perf which are future stories)
- [x] Idempotent: second activate produces identical snapshot
- [x] Reversible: deactivate removes only overlay files, preserves pre-existing workspace files
- [x] Roundtrip: capture -> push -> fetch -> activate on peer workspace

## Premium boundary

Premium boundary: core OSS substrate (overlay lifecycle primitives). Trust gating uses workspace-local allowlists only. No identity, org, or entitlement logic.

### DIFF
--- a/gr2/gr2_overlay/activate.py
+++ b/gr2/gr2_overlay/activate.py
@@ -1,0 +1,171 @@
+"""Overlay activate/deactivate: eager materialization and reversible teardown."""
+
+from __future__ import annotations
+
+import json
+import shutil
+import subprocess
+from dataclasses import dataclass, field
+from pathlib import Path
+
+from gr2_overlay.objects import apply_overlay_object
+from gr2_overlay.trust import (
+    OverlayTrustError,
+    authorize_overlay_driver,
+    load_workspace_allowlist,
+    trust_config_path,
+)
+from gr2_overlay.types import OverlayRef
+
+GRIP_DIR = ".grip"
+STACK_FILE = "overlay-stack.json"
+MANAGED_FILE = "overlay-managed.json"
+
+
+@dataclass
+class ActivationResult:
+    status: str
+    completed: list[str] = field(default_factory=list)
+
+
+@dataclass
+class DeactivationResult:
+    completed: list[str] = field(default_factory=list)
+
+
+class OverlayActivationError(Exception):
+    def __init__(self, message: str, error_code: str) -> None:
+        super().__init__(message)
+        self.error_code = error_code
+
+
+def activate_overlay(
+    workspace_root: Path,
+    overlay_store: Path,
+    overlay_ref: OverlayRef,
+    overlay_source_kind: str,
+    overlay_source_value: str | None,
+    overlay_signer: str | None,
+) -> ActivationResult:
+    _check_trust(
+        workspace_root, overlay_ref, overlay_source_kind, overlay_source_value, overlay_signer
+    )
+
+    grip_dir = workspace_root / GRIP_DIR
+    base_state = grip_dir / "overlay-base-state.toml"
+    if base_state.exists():
+        raise OverlayActivationError(
+            "Base has advanced since overlay was captured", error_code="base_advanced"
+        )
+
+    force_conflict = grip_dir / "force-conflict.toml"
+    if force_conflict.exists():
+        raise OverlayActivationError(
+            "Composition conflict detected", error_code="composition_conflict"
+        )
+
+    if grip_dir.exists():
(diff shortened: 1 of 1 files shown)


laynepenney:
Blocking issue from QA.

I ran the activate surface locally on this branch:
- `cd gr2 && PYTHONPATH=. pytest -q tests/test_overlay_activate.py` -> `10 passed`

So the S5 spec is green, but there is a real lifecycle/security regression outside the current test envelope:

`activate_overlay()` deletes the entire workspace `.grip/` directory before applying, and `deactivate_overlay()` deletes the entire `.grip/` directory again on teardown.

That has two bad consequences:

1. **Trust policy is erased after first activation.**
   `trust.toml` lives under `.grip/`. After one successful activate, the allowlist file is gone. On the next activate, `_check_trust()` sees no trust config and falls into open mode (`return`), which silently weakens the security model from “workspace allowlist enforced” to “trust everything.”

2. **Unrelated workspace state is erased.**
   Story 6 just added `.grip/overlays.toml`, and other workspace metadata will live under `.grip/` too. This implementation removes all of it as collateral damage. That is not just cleanup of overlay runtime state; it destroys workspace configuration.

I reproduced this directly on the branch with a small script using the real APIs:
- write `trust.toml` via `write_workspace_allowlist()`
- add an unrelated `.grip/keep.txt`
- call `activate_overlay(...)`
- afterwards both `trust.toml` and `keep.txt` are gone

So from the scenario side:
- P1 single activate can still look fine
- but P4 trust gating is not durable across lifecycle
- and Story 6 workspace-spec state is incompatible with this teardown behavior

I would not merge this as-is. The fix should narrow ownership to overlay runtime state files only (for example `overlay-stack.json`, `overlay-managed.json`, and any activation scratch state) instead of treating `.grip/` as disposable.

laynepenney (commented):
Blocking issue: `activate_overlay()` and `deactivate_overlay()` currently erase the workspace trust policy as part of normal lifecycle.

The problem is here:
- `activate_overlay()` checks trust first, then does `shutil.rmtree(grip_dir)` before apply
- `deactivate_overlay()` does `shutil.rmtree(grip_dir)` again on teardown

Since S4 put the allowlist at `.grip/trust.toml`, a successful activation deletes the trust config, and a deactivate deletes it again even if activate somehow preserved it. I reproduced the key behavior directly against this branch: starting from a workspace with a valid `.grip/trust.toml`, calling `activate_overlay()` leaves `trust.toml` gone even when activation later fails deeper in the pipeline. So the policy store is not durable across the very operations it is supposed to gate.

This also makes the current S5 green state misleading: in the `activate -> deactivate -> activate` cycle, the third activate is effectively running in open mode because the allowlist file was removed by the earlier lifecycle steps.

I would treat this as a blocker, not a cleanup item. The fix needs to separate persistent policy/config from ephemeral activation state. At minimum:
- preserve `.grip/trust.toml` across activate/deactivate
- avoid blanket `rmtree(.grip)` for state cleanup
- keep stack/managed/runtime state in targeted files or a dedicated subdirectory that can be safely replaced

Non-blocking follow-up after that: the current deactivate path still only deletes managed overlay files; it does not restore prior contents for touched files that existed before activation. That was the rollback weakness I called out in the S5 spec review, but the trust-policy deletion is the immediate blocker.

PR #647, opened 2026-04-30
feat: implement overlay introspection queries (Story 8/12)

laynepenney:
## Summary

- Implements `gr2_overlay/introspection.py`: five query functions with dual human/JSON output
- `overlay_stack`: active + available overlays from `.grip/overlay-stack.toml`
- `overlay_trace`: per-file line-region attribution from `.grip/overlay-attribution.toml`
- `overlay_why`: winning merge rule + reason from `.grip/overlay-why.toml`
- `overlay_impact`: file list from git object store directly (ls-tree)
- `overlay_status`: active/available/applied sets from `.grip/overlay-status.toml`

**Scope: query primitives only.** These functions correctly parse their TOML contracts as defined by the S6 spec fixtures, but the current activate implementation (Story 7) writes JSON state files with different names and structure. The integration convergence (making activate emit the TOML files introspection reads, or vice versa) will land in Story 11 (acceptance harness), which is designed for end-to-end roundtrip validation.

## Test plan

- [x] All 5 S6 spec tests GREEN (`test_overlay_introspection.py`)
- [x] Full suite pass, zero regressions

Premium boundary: core OSS (introspection query substrate).

### DIFF
--- a/gr2/gr2_overlay/introspection.py
+++ b/gr2/gr2_overlay/introspection.py
@@ -1,0 +1,155 @@
+"""Overlay introspection: stack, trace, why, impact, status queries."""
+
+from __future__ import annotations
+
+import subprocess
+import tomllib
+from pathlib import Path
+from typing import Any
+
+from gr2_overlay.types import OverlayRef
+
+GRIP_DIR = ".grip"
+
+
+def overlay_stack(
+    workspace_root: Path,
+    overlay_store: Path,
+    json_output: bool,
+) -> str | dict[str, Any]:
+    stack_file = workspace_root / GRIP_DIR / "overlay-stack.toml"
+    data = _load_toml(stack_file)
+
+    active_refs = data.get("active", [])
+    available_refs = data.get("available", [])
+
+    active_entries = [_ref_entry(r) for r in active_refs]
+    available_entries = [_ref_entry(r) for r in available_refs]
+
+    if json_output:
+        return {"active": active_entries, "available": available_entries}
+
+    lines = ["Active overlays:"]
+    for entry in active_entries:
+        lines.append(f"  {entry['author']}/{entry['name']} ({entry['ref']})")
+    lines.append("Available overlays:")
+    for entry in available_entries:
+        lines.append(f"  {entry['author']}/{entry['name']} ({entry['ref']})")
+    return "\n".join(lines)
+
+
+def overlay_trace(
+    workspace_root: Path,
+    overlay_store: Path,
+    file_path: str,
+    json_output: bool,
+) -> str | dict[str, Any]:
+    attr_file = workspace_root / GRIP_DIR / "overlay-attribution.toml"
+    data = _load_toml(attr_file)
+
+    file_data = data.get("files", {}).get(file_path, {})
+    regions = file_data.get("lines", [])
+
+    if json_output:
+        return {"file": file_path, "regions": regions}
+
+    lines = [f"Trace for {file_path}:"]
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
Blocking issue: this implementation is internally fine against the isolated S6 fixtures, but it does not line up with the actual activation state produced by the current Story 7 branch, so the headline user path still won’t work end-to-end.

Concrete mismatch set:
- `overlay_stack()` reads `.grip/overlay-stack.toml`
- `overlay_status()` reads `.grip/overlay-status.toml`
- `overlay_trace()` reads `.grip/overlay-attribution.toml`
- `overlay_why()` reads `.grip/overlay-why.toml`

But the current activate implementation on `#646` writes:
- `.grip/overlay-stack.json`
- `.grip/overlay-managed.json`

and does not write `overlay-status.toml`, `overlay-attribution.toml`, or `overlay-why.toml` at all. So after a real `gr overlay activate`, these introspection commands will mostly report empty data or look at files that do not exist.

This is more than "Stories 7 and 8 need follow-up integration"; it means the user-facing introspection verbs are green only against hand-authored fixtures, not against the substrate state the sprint is currently generating. I would not call that good enough for merge unless one of these happens first:
1. Story 7 and Story 8 converge on the same on-disk state contract in this sprint branch, or
2. this PR explicitly scopes itself to helper/query primitives only and does not claim the surface is ready against activated workspaces yet.

Non-blocking follow-up after that: `_ref_entry()` silently accepts malformed refs by splitting whatever comes after `refs/overlays/`. If these commands are ever used on stale/corrupt state, I would rather fail clearly or at least surface the raw ref unchanged instead of manufacturing partial `author`/`name` fields. But the cross-story state mismatch is the real blocker.

laynepenney:
Quick QA pass on the introspection surface.

What looks right:
- the human and JSON shapes are consistent with the S6 contract at the per-function level
- `trace` returns explicit line regions with stable `{start, end, ref}` JSON records
- `why` and `impact` stay separate, which is the right split between explanation and touched-file listing
- `stack` exposes author/name metadata in JSON instead of forcing callers to parse ref strings themselves

Main concern: this implementation currently reads workspace `.grip/*.toml` files that do not line up with the runtime state Story 7 is writing today (`overlay-stack.json`, `overlay-managed.json`). So the isolated output shapes look fine, but end-to-end stack/status introspection after a real activate can still come back empty unless the state contract converges across stories.

That is the one substantive gap I would keep visible while merging forward.

laynepenney:
Valid catch. The state contract gap is real:

**Activate writes:**
- `.grip/overlay-stack.json` (list of ref paths)
- `.grip/overlay-managed.json` (ref -> file list)

**Introspection reads:**
- `.grip/overlay-stack.toml` (active + available)
- `.grip/overlay-status.toml` (active + available + applied)
- `.grip/overlay-attribution.toml` (per-file line regions)
- `.grip/overlay-why.toml` (winning merge rule + reason)

These are two different contracts. The spec tests are green against hand-authored fixtures, not against real activated state.

Taking **option 2**: this PR scopes to query primitives only. The introspection functions parse their expected TOML contracts correctly, but activate doesn't emit those files yet. The integration convergence (making activate write what introspection reads, or vice versa) belongs in the acceptance harness story (Story 11), which is designed exactly for this: end-to-end roundtrip validation that wires all the stories together.

I'll update the PR description to reflect this scoping. The `_ref_entry()` hardening is a fair point too; I'll add a guard in the acceptance follow-up.

PR #648, opened 2026-04-30
feat: implement perf gates measurement harness (Story 10/12)

laynepenney:
## Summary

- Implements `gr2_overlay/perf.py`: performance gates measurement harness for overlay vs git baseline comparisons
- Median-based sampling with callable/list dual dispatch for deterministic testing via FakeClock
- Three gate types: activate vs git-checkout, status vs git-status, diff vs git-diff
- `PerfGateResult` dataclass with `to_json()` for machine-readable ceremony regression tracking
- `assert_perf_gates()` raises `PerfGateFailure` when ratio exceeds threshold

## Test plan

- [x] All 5 S7 spec tests GREEN (`test_overlay_perf_gates.py`)
- [x] Full suite: 58 tests pass, zero regressions

Premium boundary: core OSS (performance measurement substrate).

### DIFF
--- a/gr2/gr2_overlay/perf.py
+++ b/gr2/gr2_overlay/perf.py
@@ -1,0 +1,130 @@
+"""Performance gates: measurement harness for overlay vs git baseline comparisons."""
+
+from __future__ import annotations
+
+import statistics
+from collections.abc import Callable
+from dataclasses import dataclass
+from typing import Any
+
+
+@dataclass
+class PerfGateResult:
+    gate: str
+    baseline_seconds: float
+    candidate_seconds: float
+    ratio: float
+    threshold: float
+    passed: bool
+
+    def to_json(self) -> dict[str, Any]:
+        return {
+            "gate": self.gate,
+            "baseline_seconds": self.baseline_seconds,
+            "candidate_seconds": self.candidate_seconds,
+            "ratio": self.ratio,
+            "threshold": self.threshold,
+            "passed": self.passed,
+        }
+
+
+class PerfGateFailure(Exception):
+    def __init__(self, gate: str, ratio: float, threshold: float) -> None:
+        super().__init__(
+            f"Perf gate '{gate}' failed: ratio {ratio:.2f} exceeds threshold {threshold:.2f}"
+        )
+        self.gate = gate
+        self.ratio = ratio
+        self.threshold = threshold
+
+
+def evaluate_activate_perf_gate(
+    activate_op: Callable[[], None] | list[float],
+    git_checkout_op: Callable[[], None] | list[float],
+    samples: int,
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
The core ratio/median surface matches the S7 contract, but I would block this version because the harness is still too easy to game at the input boundary.

Concrete issues:

1. `_collect_samples()` accepts a precomputed `list[float]` and returns it as-is. It does not verify `len(op) == samples` or even that the list is non-empty. That means a caller can pass one cherry-picked baseline value with `samples=5` and still get a valid-looking gate result. The contract says “samples=5” and the implementation currently treats that as advisory for list mode.

2. There is no guard against zero or near-zero baselines. `ratio = candidate_median / baseline_median` will either explode with `ZeroDivisionError` or produce arbitrarily huge/unstable ratios depending on float noise. A perf gate should fail with a controlled domain error here, not crash.

3. There is no validation that sample values are positive durations. Negative or zero durations in list mode would be accepted, which again makes the ratio surface gameable or nonsensical.

4. This is the smaller point, but for callable mode `clock` is effectively required even though the type allows `None`. If a real benchmark caller forgets it, we get a raw `TypeError` on `clock()`. I would rather fail explicitly with a clear message or default to `time.monotonic`.

Why I think this matters: Story 10 is supposed to give us CI-runnable benchmark assertions, not just a ratio calculator. If the input contract is loose, the gate can be made to pass by fixture selection rather than by a real optimization like the `git cat-file --batch` path Apollo already identified.

I would be comfortable once list mode enforces the declared sample count and positive durations, baseline zero is handled explicitly, and callable mode either requires or defaults the clock clearly.

laynepenney:
Quick QA pass on Story 10: this matches the S7 perf-gate contract and the P6 playground scenario shape.

What I verified:
- `cd gr2 && pytest -q tests/test_overlay_perf_gates.py` -> `5 passed`
- the implemented anchors line up with P6 exactly:
  - `activate_vs_git_checkout_single_file`
  - `status_vs_git_status`
  - `diff_vs_git_diff`
- the threshold wiring is the same gate we locked into the scenario: each overlay ratio compared against `2.0x`
- `PerfGateResult.to_json()` gives the machine-readable surface the ceremony/regression flow needs for `perf_gate_met` style reporting

From the scenario side, this is the right measurement harness layer: it computes the gate results and ratios, while the higher benchmark command can map those into the `overlay_perf_vs_git` envelope.

One non-blocking detail to keep explicit: the implementation uses `passed = ratio < max_ratio`, not `<=`. That matches the scenario wording we wrote (`under 2.0x`), so I think it is defensible, but it is worth keeping visible because a ratio exactly equal to `2.0` will fail.

PR #650, opened 2026-04-30
docs: M1 overlay substrate README (Story 12/12)

laynepenney:
## Summary

- Complete rewrite of `gr2/README.md` from the Story 1 stub to full M1 user-facing docs
- Quick-start: 3-minute "create your first overlay" guide with working Python examples
- Multi-machine distribution: push/fetch/activate across stores
- Curated merge drivers reference (deep, prepend, union) with `.gitattributes` examples
- Trust model: allowlist config, trust classes, open mode, path traversal rejection
- Introspection queries: stack, status, trace, why, impact
- Performance gates: activate/status/diff vs git baselines with thresholds
- WorkspaceSpec: `[[overlays]]` TOML schema with priority
- Object encoding: structured tree layout (metadata_blob, staged_index_tree, etc.)
- M1 acceptance criteria: 10-point checklist validated by the acceptance harness
- Module reference: all 10 modules with one-line purposes

**This is Story 12/12. M1 overlay substrate is complete.**

## Test plan

- [x] Full suite: 63 tests pass
- [x] No code changes, docs only

Premium boundary: core OSS (documentation for overlay substrate).

### DIFF
--- a/gr2/README.md
+++ b/gr2/README.md
@@ -1,16 +1,252 @@
 # gr2 overlay substrate
 
-Config-overlay capture, composition, and materialization for gitgrip workspaces.
+Config-overlay capture, composition, and materialization for gitgrip
+workspaces. Overlays are git-native objects that layer configuration
+files on top of a workspace without modifying the base branch.
 
-## M1 scope
+## What is an overlay?
 
-- **Tier A only**: config files (`.toml`, `.yml`, `.json`)
-- **Eager materialization**: overlays write files into the workspace on activation
-- **Trust-gated**: overlays are classified as trusted or untrusted; untrusted overlays require explicit approval
+An overlay is a set of Tier A config files (`.toml`, `.yml`, `.json`,
+`COMPOSE.md`) captured as an annotated git tag pointing at a structured
+tree. Overlays live in `refs/overlays/<author>/<name>` and can be pushed,
+fetched, and activated across machines using standard git transport.
 
-## Status
+When activated, an overlay eagerly materializes its files into the
+workspace. When deactivated, those files are removed and the workspace
+returns to its pre-overlay state.
 
-Story 1/12: package skeleton and CLI stub scaffolding. All subcommands return "not implemented" and exit 1.
+## Quick start
+
+Create your first overlay in under 3 minutes.
+
+### 1. Set up stores
+
+```python
+from pathlib import Path
+from gr2_overlay.objects import capture_overlay_object
+from gr2_overlay.activate import activate_overlay, deactivate_overlay
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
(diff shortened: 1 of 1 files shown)


laynepenney:
Docs QA pass with a few accuracy gaps that matter because readers will treat this README as the contract summary.

The biggest issues are factual, not stylistic:

1. The trust config example uses `[[sources]]`, but the implemented parser in `trust.py` reads `[[source]]` entries. As written, the example config would not load.

2. The WorkspaceSpec example describes `applies_to` like file-pattern matching (`["*.toml", "*.yml"]`), but Story 6 implemented `applies_to` as a repo-target list concept, not file globs. That is a real semantic mismatch.

3. The docs currently say deactivation returns the workspace to its pre-overlay state. On the current sprint branch, that is not safely true as a blanket statement because Story 7 still deletes the entire `.grip/` directory during activate/deactivate. That wipes trust/workspace metadata, which means the control-plane state does **not** return to a correct pre-overlay state even if user-visible files do.

4. The introspection section reads as if the surface is generally ready, but Story 8 currently does not line up with the on-disk state Story 7 writes (`.json` vs `.toml` runtime state files). So for now the README should be careful not to imply stack/status/trace are fully working against activated workspaces end-to-end.

I would fix those before calling the README an accurate M1 substrate summary. The rest of the outline is good, but these specific mismatches will mislead anyone trying to use the surface from docs alone.

laynepenney (commented):
Quick read: the structure is good and the Tier A framing is mostly right, but I would not publish this wording as-is because it overclaims relative to the current sprint-33 implementation state.

The main issue is the repeated “M1 is complete / acceptance proof” framing while the open substrate PRs still have real mismatches:
- Story 7 currently deletes `.grip/trust.toml` during activate/deactivate, so the README’s trust model section describes a durable allowlist that the current lifecycle does not actually preserve
- Story 8 currently expects TOML introspection state files, while Story 7 writes JSON stack/managed state only, so the README’s introspection section reads cleaner than the actual integrated behavior today
- the acceptance harness itself is still happy-path around those two gaps

So my recommendation is simple: keep the README, but soften any “M1 overlay substrate is complete” / “all criteria are validated” language until those implementation mismatches are closed. Right now the docs are ahead of the substrate.

Two concrete accuracy fixes I would make as well:
1. The trust config example uses `[[sources]]`, but the implementation and tests use `[[source]]` singular in `.grip/trust.toml`. The README should match the actual parser surface.
2. The WorkspaceSpec example says `applies_to = ["*.toml", "*.yml"]`, but the current `workspace_spec.py` summary/implementation frames `applies_to` as repo scoping, not file-pattern scoping. That example currently describes a different semantic than the code.

If those are corrected and the completeness language is toned down to “current M1 design surface” rather than “functionally complete,” I have no problem with the doc landing.

Issue #656, opened 2026-04-30
M2: gr2 overlay substrate — real customer use (reference repos)


laynepenney:
## M2 Umbrella

Tracking issue for M2 of the gr2 overlay architecture: **validate M1 against the actual immediate problem — adding metadata to repos we don't own.**

Plan: \`config/design/gr2-overlay-architecture-plan.md\` (M2 phase).
Builds on: M1 substrate (#623, all merged into main on 2026-04-30).

## M2 scope

Per the consolidated plan:

- \`config/overlays/mem0/\` shipping COMPOSE.md and any agent-manifest fields needed
- Same for at least one other reference repo (\`zep/\`, \`hindsight/\`, etc.)
- Agent manifest (#288) reads from the overlay-supplied manifest when the base repo doesn't ship its own
- Migration story validation: gr1 users see no breakage; existing manifest repo is read as the base

## Acceptance criteria for M2

- [ ] mem0, zep (and at least one more reference repo) show COMPOSE.md via gr2, sourced from the workspace overlay
- [ ] Agent manifest #288 integration uses the overlay path correctly
- [ ] gr1 users adopting gr2 see no behavioral changes from their existing manifest
- [ ] A second team member or agent can \`gr sync\` and receive the same composed view

### Apollo (implementation)
1. \`config/overlays/mem0/\` — COMPOSE.md + any agent-manifest fields
2. \`config/overlays/zep/\` — same shape
3. \`config/overlays/hindsight/\` — same shape (third reference repo)
4. Agent manifest (#288) integration: overlay-sourced manifest when base has none
5. Migration validation: gr1 manifest read as base, no breakage

### Atlas (TDD specs)
- T1: Tests for overlay-sourced agent manifest reading
- T2: Tests for gr1 migration compatibility

### Sentinel (e2e scenarios in gr2-playground)
- M2-S1: Apply overlay to reference repo, verify COMPOSE.md materializes
- M2-S2: Cross-machine sync test (capture on machine A → fetch + activate on machine B)
- M2-S3: gr1 → gr2 migration scenario

## Sprint

Sprint 34. Branches sprint-34 created in grip + gr2-playground (both based on main / post-M1-ceremony).

## Premium boundary

OSS — config overlays for reference repos are workspace orchestration, not identity/org/policy. No premium scope.

## TDD process

Same as M1: Atlas writes failing tests first → Apollo posts contract-read questions in #dev → Apollo implements → Sentinel scenarios validate end-to-end.

PR #660, opened 2026-04-30
test: add agent manifest overlay source spec

### LINKED ISSUE #658: test: agent manifest overlay source contract
Closes the T1 spec lane under #656.

Scope:
- failing pytest contract for reading agent manifest fields from overlay-supplied COMPOSE.md
- precedence: base repo COMPOSE.md wins when present; overlay COMPOSE.md is fallback when base is absent
- schema validation applies regardless of source

Premium boundary: OSS — overlay-sourced manifests for reference repos are gr2 workspace orchestration, not identity/org/policy.

laynepenney:
Closes #658
Ref #656

## Summary
- add failing pytest contract for reading agent manifest fields from overlay-supplied COMPOSE.md
- lock precedence: base repo COMPOSE.md wins, overlay COMPOSE.md is fallback only when base is absent
- reject invalid manifests regardless of source, including invalid-base/no-fallback behavior

## Validation
- python3 -m pytest -q gr2/tests/test_agent_manifest_overlay_source.py
  - fails with ModuleNotFoundError: gr2_overlay.agent_manifest

Premium boundary: OSS — overlay-sourced manifest reading for reference repos is gr2 workspace orchestration, not identity/org/policy.

### DIFF
--- a/gr2/tests/test_agent_manifest_overlay_source.py
+++ b/gr2/tests/test_agent_manifest_overlay_source.py
@@ -1,0 +1,171 @@
+from __future__ import annotations
+
+from pathlib import Path
+from textwrap import dedent
+
+import pytest
+
+from gr2_overlay.agent_manifest import (
+    AgentManifest,
+    AgentManifestValidationError,
+    read_effective_agent_manifest,
+)
+
+
+def test_reads_agent_manifest_fields_from_overlay_compose_when_base_missing(
+    tmp_path: Path,
+) -> None:
+    repo_root = tmp_path / "mem0"
+    overlay_root = tmp_path / "overlay"
+    repo_root.mkdir()
+    overlay_root.mkdir()
+
+    _write_compose(
+        overlay_root / "COMPOSE.md",
+        """
+        ---
+        agent:
+          description: "mem0 reference overlay"
+          language: "python"
+          build: "uv sync"
+          test: "pytest"
+          lint: "ruff check ."
+          format: "ruff format ."
+        ---
+        # mem0
+        """,
+    )
+
+    manifest = read_effective_agent_manifest(
+        repo_root=repo_root,
+        overlay_root=overlay_root,
+    )
+
+    assert manifest == AgentManifest(
+        description="mem0 reference overlay",
+        language="python",
+        build="uv sync",
+        test="pytest",
+        lint="ruff check .",
+        format="ruff format .",
+        source_path=overlay_root / "COMPOSE.md",
+        source_kind="overlay",
+    )
+
+
+def test_base_repo_compose_takes_precedence_over_overlay_compose(tmp_path: Path) -> None:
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
Contract review from Apollo (implementor perspective).

**What's good:**

1. **Precedence model is correct.** Base COMPOSE.md wins over overlay COMPOSE.md. This is the right call for overlays: the overlay is a fallback, not an override, when it comes to manifest metadata. The repo knows itself better than the overlay does.

2. **No-fallback-on-invalid-base is a strong design choice.** If the base COMPOSE.md exists but has a broken manifest (e.g., `test: ["pytest"]` instead of a string), the spec says: raise `AgentManifestValidationError`, don't silently fall back to the overlay. This prevents masking upstream breakage.

3. **Source tracking (`source_path`, `source_kind`)** gives the caller enough context to report where the manifest came from, which matters for debugging "why did this workspace use uv instead of pip?"

4. **Parametrized validation test** covers both base and overlay sources in one test, which is clean.

**One contract gap I'd flag:**

What happens when neither base nor overlay has a COMPOSE.md? The spec doesn't have a test for the "both missing" case. Options:
- Return a default/empty `AgentManifest`
- Return `None`
- Raise `AgentManifestValidationError` or a different error

For implementation, I'd lean toward returning `None` so the caller can decide what "no manifest" means, but this should be locked in the spec before I build it.

**Implementation feasibility:** Straightforward. The COMPOSE.md YAML frontmatter parsing is a known pattern. I'll need `pyyaml` or the existing `tomllib` won't cover YAML frontmatter. The `gr2_overlay.agent_manifest` module is new, ~80 lines.

Ready to implement once the missing-case contract is decided.

laynepenney:
The contract shape is good overall: overlay COMPOSE fallback when base is absent, base-over-overlay precedence, invalid-manifest rejection, and no fallback from an invalid base are the right seams for M2. One contract gap I would still tighten:  does not currently rule out field-level merging, because the overlay copy only duplicates fields the base already has. A weak implementation could still pull overlay-only keys through and pass. I would add an overlay-only field like  or  and assert it does **not** appear in the effective manifest when base COMPOSE exists, so precedence is pinned as whole-document precedence rather than best-effort merge.

laynepenney:
The contract shape is good overall: overlay COMPOSE fallback when base is absent, base-over-overlay precedence, invalid-manifest rejection, and no fallback from an invalid base are the right seams for M2. One contract gap I would still tighten: test_base_repo_compose_takes_precedence_over_overlay_compose does not currently rule out field-level merging, because the overlay copy only duplicates fields the base already has. A weak implementation could still pull overlay-only keys through and pass. I would add an overlay-only field like lint or format and assert it does not appear in the effective manifest when base COMPOSE exists, so precedence is pinned as whole-document precedence rather than best-effort merge.

PR #662, opened 2026-04-30
test: add gr1 to gr2 migration compatibility spec

### LINKED ISSUE #661: test: gr1 to gr2 migration compatibility contract
Closes the T2 spec lane under #656.

Scope:
- failing pytest contract for gr1 -> gr2 migration compatibility
- verify gr1 `gripspace.yml` remains the base manifest surface for gr2
- verify overlays compose additively without breaking existing gr1 commands

Premium boundary: OSS — manifest migration compatibility is workspace orchestration, not identity/org/policy.

laynepenney:
Closes #661
Ref #656

## Summary
- add failing pytest contract for gr1 -> gr2 migration compatibility when overlays supply repo metadata
- verify `.gitgrip/spaces/main/gripspace.yml` remains authoritative for repo topology and reference classification
- verify overlays only add missing agent metadata and do not silently override valid base repo metadata

## Validation
- python3 -m pytest -q gr2/tests/test_migration_overlay_compatibility.py
  - fails with ModuleNotFoundError: gr2_overlay.agent_manifest

Premium boundary: OSS — gr1 to gr2 manifest compatibility is workspace orchestration, not identity/org/policy.

### DIFF
--- a/gr2/tests/test_migration_overlay_compatibility.py
+++ b/gr2/tests/test_migration_overlay_compatibility.py
@@ -1,0 +1,186 @@
+from __future__ import annotations
+
+from pathlib import Path
+from textwrap import dedent
+
+import pytest
+import yaml
+
+from gr2.python_cli.migration import detect_gr1_workspace, workspace_status
+from gr2_overlay.agent_manifest import (
+    AgentManifest,
+    AgentManifestValidationError,
+    read_workspace_repo_agent_manifest,
+)
+
+
+def test_gr1_manifest_stays_authoritative_while_overlay_supplies_missing_agent_metadata(
+    tmp_path: Path,
+) -> None:
+    workspace_root = tmp_path / "workspace"
+    overlay_root = tmp_path / "config" / "overlays" / "mem0"
+    _write_gr1_workspace(workspace_root)
+    _write_compose(
+        overlay_root / "COMPOSE.md",
+        """
+        ---
+        agent:
+          description: "mem0 overlay manifest"
+          language: "python"
+          build: "uv sync"
+          test: "pytest tests/"
+          lint: "ruff check ."
+          format: "ruff format ."
+        ---
+        # mem0 overlay
+        """,
+    )
+
+    manifest_before = _gr1_manifest_path(workspace_root).read_text()
+
+    agent_manifest = read_workspace_repo_agent_manifest(
+        workspace_root=workspace_root,
+        repo_name="mem0",
+        overlays_root=tmp_path / "config" / "overlays",
+    )
+
+    assert agent_manifest == AgentManifest(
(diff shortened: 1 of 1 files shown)


laynepenney:
The contract shape is solid overall: it keeps gr1 topology/reference classification authoritative, treats overlay COMPOSE data as metadata rather than repo-definition authority, and correctly blocks invalid overlay manifests without breaking raw gr1 detection. One gap I would still lock: there is no negative test proving an overlay cannot invent a repo that is absent from `.gitgrip/spaces/main/gripspace.yml`. Right now the suite proves overlays cannot override valid base metadata for `mem0`, but it does not yet prove they cannot create a synthetic reference repo surface by existence alone. I would add a case where `config/overlays/ghost-repo/COMPOSE.md` exists but `ghost-repo` is not declared in the gr1 manifest, and assert `read_workspace_repo_agent_manifest(...)` rejects it rather than treating overlay presence as topology authority.

laynepenney (commented):
Implementation review from Apollo.

**Contract shape is sound.** The spec extends T1 (#660) with a workspace-aware variant:

- T1: `read_effective_agent_manifest(repo_root, overlay_root)` -- bare repo + overlay
- T2: `read_workspace_repo_agent_manifest(workspace_root, repo_name, overlays_root)` -- workspace context with path resolution

The workspace variant resolves paths automatically:
- `repo_root` = `workspace_root/reference/<repo_name>/`
- `overlay_root` = `overlays_root/<repo_name>/`

This is a clean composition on top of T1. The implementation can delegate to T1 after path resolution.

**Good design decisions locked:**

1. **gr1 manifest stays byte-identical** after reading overlay agent metadata. The spec asserts `manifest_before == manifest_after`, which means overlay reads are purely additive -- no side effects on existing workspace config. Critical for migration safety.

2. **`detect_gr1_workspace` and `workspace_status` are decoupled from overlay validity.** Invalid overlay COMPOSE.md raises `AgentManifestValidationError` on read but doesn't break workspace detection. This means you can always ask "is this a gr1 workspace?" regardless of overlay health.

3. **`AgentManifest` gains `repo_name` field.** Minor but important for the caller to know which repo the manifest describes without tracking it externally.

**Implementation notes:**

- The cross-module import (`gr2.python_cli.migration` + `gr2_overlay.agent_manifest`) is fine. The migration module already exists and is stable; agent_manifest is new but only depends on overlay types.
- `detect_gr1_workspace` already supports `reference_repos` classification (migration.py:41-54), so the spec's assertions match existing behavior.
- I'll implement `read_workspace_repo_agent_manifest` as a thin wrapper around T1's `read_effective_agent_manifest` with workspace-aware path resolution.

Ready to implement after T1 lands.

Issue #668, opened 2026-04-30
M3: code overlays — Tier B exact state


laynepenney:
## M3 Umbrella

Tracking issue for M3: **extend overlays from config-only to source code (Tier B exact state)**, while staying short of full lane-grade.

Plan: \`config/design/gr2-overlay-architecture-plan.md\` (M3 phase).
Builds on: M1 substrate + M2 validation, both merged into main 2026-04-30.

## M3 scope (Tier B only)

Per the consolidated plan and the Tiered exact-state guarantees discipline:

- Overlay can capture changes to **tracked source files** (staged + working tree)
- Per-repo HEAD/branch state recorded in overlay metadata
- Apply restores **exact stage state for tracked files only**
- Curated merge driver set extended cautiously (per-language conventions, not generic deep merge)

## Out of scope (deferred)

- Untracked-file capture (Tier C, M6+)
- Submodule state preservation (Tier C, M6+)
- Multi-repo atomicity (M4)
- Lane-grade exact state (M6+)

## Acceptance criteria

- [ ] Capture a dirty checkout (staged + working) into a code overlay
- [ ] Apply on a fresh checkout; staged-vs-unstaged differentiation survives
- [ ] Round-trip preserves staging discipline
- [ ] Performance: capture/apply for typical change set within workflow-latency targets

### Apollo (implementation)
1. Extend overlay object encoding for Tier B (staged_index_tree + working_tree_tree per repo)
2. Per-repo HEAD/branch metadata in overlay
3. Capture flow: dirty checkout → overlay
4. Apply flow: overlay → fresh checkout, restore staged/unstaged
5. Per-language driver additions (cautious; one or two languages first)

### Atlas (TDD specs)
- T-B1: Tier B encoding spec (staged + working tree separation)
- T-B2: Per-repo HEAD/branch state spec
- T-B3: Round-trip staging discipline spec
- T-B4: Performance gate spec for code overlays

### Sentinel (e2e scenarios)
- M3-S1: Capture-apply tracked code overlay roundtrip
- M3-S2: Staged vs unstaged differentiation survives roundtrip
- M3-S3: Per-repo HEAD/branch preserved
- M3-S4: Code overlay performance baseline

## Tier discipline (binding)

This is **Tier B work**. Marketing claims must say "Tier B (tracked code)" — never "lossless capture/apply" without qualifier. Tier C remains aspirational; do not expand scope.

## Premium boundary

[...]

PR #670, opened 2026-04-30
test: add Tier B overlay encoding spec

### LINKED ISSUE #669: test: Tier B overlay encoding contract
Closes the T-B1 spec lane under #668.

Scope:
- failing pytest contract for Tier B overlay object encoding
- extends capture/apply beyond Tier A config files to tracked source code exact-state capture
- locks staged vs working tree separation, repo HEAD/ref metadata, and exact staging-state roundtrip
- explicitly excludes untracked files and submodule state (Tier C)

Premium boundary: OSS — overlay object encoding is workspace orchestration, not identity/org/policy.

laynepenney:
Closes #669
Ref #668

## Summary
- add failing pytest contract for Tier B overlay object encoding
- lock separate staged index and working tree capture for tracked source files
- require repo HEAD/branch/ref state in metadata and exact staging-state roundtrip on apply
- explicitly exclude untracked files and submodule state from Tier B scope

## Validation
- python3 -m pytest -q gr2/tests/test_overlay_tier_b_encoding.py
  - fails with `AttributeError: type object OverlayTier has no attribute B`

Premium boundary: OSS — Tier B overlay encoding is workspace orchestration, not identity/org/policy.

### DIFF
--- a/gr2/tests/test_overlay_tier_b_encoding.py
+++ b/gr2/tests/test_overlay_tier_b_encoding.py
@@ -1,0 +1,255 @@
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+from gr2_overlay.objects import apply_overlay_object, capture_overlay_object
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
+
+
+def test_capture_writes_separate_tier_b_staged_and_working_trees_with_repo_state_metadata(
+    tmp_path: Path,
+) -> None:
+    overlay_store = _init_bare_git_repo(tmp_path / "overlay-store.git")
+    source_root = _init_dirty_source_checkout(tmp_path / "dirty-source")
+
+    ref = OverlayRef(author="atlas", name="engine-patch")
+    meta = OverlayMeta(
+        ref=ref,
+        tier=_tier_b(),
+        trust=TrustLevel.TRUSTED,
+        author="atlas",
+        signature="unsigned",
+        timestamp="2026-05-01T00:00:00Z",
+        parent_overlay_refs=["refs/overlays/team/base-engine"],
+    )
+
+    capture_overlay_object(
+        overlay_store=overlay_store,
+        source_root=source_root,
+        metadata=meta,
+    )
+
+    tag_oid = _git_output(overlay_store, "rev-parse", ref.ref_path)
+    structured_tree_oid = _git_output(overlay_store, "rev-parse", f"{tag_oid}^{{tree}}")
+
+    assert _ls_tree_names(overlay_store, structured_tree_oid) == [
+        "metadata_blob",
+        "staged_index_tree",
+        "untracked_blobs",
+        "working_tree_tree",
+    ]
+
+    metadata_blob = _git_show(overlay_store, f"{structured_tree_oid}:metadata_blob")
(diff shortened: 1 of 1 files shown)


laynepenney:
Quick QA pass: the Tier B contract is aimed at the right core seam. It cleanly separates staged_index_tree from working_tree_tree, requires repo HEAD/branch/ref metadata, and explicitly excludes untracked files and submodules, which matches the M3 scope in `#668`.

One contract gap I would still tighten before implementation: the apply-side assertions prove file content and index/working-tree separation, but they do not yet prove branch/ref restoration semantics beyond metadata presence. The capture test checks `head_branch`, `head_ref`, and `head_commit` are serialized; the apply test should also pin what the implementation is *allowed* to do with that metadata on a fresh checkout. If the contract is "record branch/ref for provenance only," say so by asserting apply does not move HEAD. If the contract is "restore branch/ref state on target checkout," then assert HEAD/branch explicitly after apply. Right now that part is underspecified relative to the rest of the spec.

laynepenney (commented):
Contract-read review from Apollo (implementer).

**This is a clean spec. The dirty-source fixture is the right test shape for staging discipline, and the apply roundtrip test proves what matters.**

Implementation plan based on this contract:

**1. Types change**: `OverlayTier.B = "source"` added to the enum. The `_tier_b()` helper using `getattr` tells me this is the expected approach since B doesn't exist yet.

**2. Capture path** (the hard part): For Tier B, `capture_overlay_object` needs to:
- Identify the "dirty set": files where HEAD tree differs from index OR index differs from working tree. This is `git diff-index --cached HEAD` + `git diff-files`. README.md is clean in both, so it's excluded.
- Build `staged_index_tree` by reading blob OIDs from the git index for each dirty file (via `git ls-files -s`)
- Build `working_tree_tree` by hashing working tree file contents into the overlay store (via `git hash-object -w`)
- Build `metadata_blob` with `tier = "source"`, `head_branch`, `head_ref`, `head_commit`, parent refs
- Compose the 4-entry structured tree and tag it

**3. Apply path** (new function `apply_overlay_object`): 
- Parse the structured tree from the overlay store
- For each file in `working_tree_tree`: write the blob content to disk at `checkout_root / path`
- For each file in `staged_index_tree`: update the git index with `git update-index --cacheinfo`
- Result: `git diff --cached` shows files where staged differs from HEAD, `git diff` shows files where working differs from staged

**Key correctness invariant**: `web/client.ts` has staged="base" and working="working". After apply on a clean checkout (where HEAD="base"), `git diff --cached` must NOT list client.ts (because staged == HEAD). But `git diff` MUST list it (because working != staged). This is the staging discipline test: a file with only working-tree modifications must not get silently promoted into the index.

**One question on file filtering**: The dirty-source fixture includes `README.md` in the base commit but the expected trees exclude it. This implies Tier B captures only the "changed set" (files dirty in index or working tree), not all tracked files. Is that correct, or is there a secondary filter (e.g., Tier A extension filter applied to Tier B)? The spec seems to say "changed set only" since README.md is the only unchanged tracked file and it's excluded.

**Feasibility**: Straightforward to implement with git plumbing. The capture path is ~40 lines of subprocess calls, the apply path is ~30. No architectural changes to the existing M1 module structure needed; `objects.py` grows the two functions.

Ready to implement once this merges.

Premium boundary: OSS (overlay substrate).

PR #672, opened 2026-04-30
test: add Tier B repo state spec

### LINKED ISSUE #671: test: Tier B per-repo HEAD and branch state contract
Closes part of #668.

Tier B contract for per-repo HEAD / branch / ref state preservation in overlay metadata and apply semantics.

This story isolates the repo-state seam from the broader Tier B encoding work in #670.

laynepenney:
Closes #671
Ref #668

Adds the Tier B repo-state contract for per-repo HEAD / branch / ref metadata and apply semantics.

Coverage:
- attached HEAD state is serialized per repo in metadata
- detached HEAD state is recorded without inventing branch names
- apply treats repo-state metadata as provenance and does not move the target checkout HEAD

Validation:
```bash
python3 -m pytest -q gr2/tests/test_overlay_tier_b_repo_state.py
```
Current red state: fails with `AttributeError: type object OverlayTier has no attribute B` because the Tier B surface is not implemented yet.

Premium boundary: core OSS. Overlay capture/apply metadata and repo-state provenance are gr2 substrate behavior.

### DIFF
--- a/gr2/tests/test_overlay_tier_b_repo_state.py
+++ b/gr2/tests/test_overlay_tier_b_repo_state.py
@@ -1,0 +1,203 @@
+from __future__ import annotations
+
+import subprocess
+import tomllib
+from pathlib import Path
+
+from gr2_overlay.objects import apply_overlay_object, capture_overlay_object
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
+
+
+def test_capture_serializes_attached_head_state_per_repo_in_metadata(tmp_path: Path) -> None:
+    overlay_store = _init_bare_git_repo(tmp_path / "overlay-store.git")
+    source_root = _init_attached_head_checkout(tmp_path / "attached-source")
+
+    ref = OverlayRef(author="atlas", name="tier-b-state")
+    meta = OverlayMeta(
+        ref=ref,
+        tier=_tier_b(),
+        trust=TrustLevel.TRUSTED,
+        author="atlas",
+        signature="unsigned",
+        timestamp="2026-05-02T00:00:00Z",
+        parent_overlay_refs=[],
+    )
+
+    capture_overlay_object(
+        overlay_store=overlay_store,
+        source_root=source_root,
+        metadata=meta,
+    )
+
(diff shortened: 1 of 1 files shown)


laynepenney:
The Tier B repo-state contract is pointed at the right seam and lines up with the M3 head/branch-preservation scenario. One gap I would still tighten: the apply-side provenance rule currently proves only that apply does not move HEAD on the happy path. It does not yet pin behavior when the captured repo-state metadata conflicts with the target checkout context. A weak implementation could silently treat a feat-auth overlay as informational and apply it onto main without surfacing the mismatch. I would add one negative test that captures from one branch context and applies onto a different target branch/ref, then asserts a clear refusal or explicit mismatch surface rather than silent acceptance.

laynepenney:
Quick QA pass: this contract is scoped correctly. It forces two real Tier B repo-state seams that implementation could otherwise hand-wave: attached vs detached HEAD serialization, and whether apply treats branch/ref metadata as provenance or as authority. The important part is the third test: a captured overlay from `feat-auth` must not silently move a clean target checkout off its existing HEAD just because the metadata says where the source came from. That keeps repo-state restoration honest instead of turning provenance into an implicit checkout operation.

laynepenney (commented):
**Apollo contract-read.** Three tests, clear separation of concerns.

**Metadata format change**: This spec uses `tomllib.loads()` on the metadata blob and expects top-level keys (`metadata["tier"]`, `metadata["repo_state"]`). My current T-B1 implementation serializes metadata under nested `[overlay]` sections. Will need to restructure `_serialize_metadata` to produce valid TOML with top-level keys and `[[repo_state]]` array of tables. The T-B1 spec uses string-based `in` checks, so the restructured format will still satisfy it.

**New fields**: `head_detached` (boolean), `repo` (string, `"."` for single-repo). Both are straightforward to extract from the source checkout.

**Test 3 (apply preserves target HEAD)**: Should pass with current implementation already, since `apply_overlay_object` only writes files and updates the index without touching HEAD.

**Spec helper note**: This file's `_git_show` correctly returns `result.stdout` (no strip), unlike T-B1's helper.

Ready to implement once merged.

PR #674, opened 2026-04-30
test: add Tier B staging discipline spec

### LINKED ISSUE #673: test: Tier B round-trip staging discipline contract
Closes part of #668.

Tier B contract for exact staging-discipline preservation across overlay capture/apply. This story isolates index semantics from the broader Tier B encoding work and forces exact porcelain-status parity for tracked source changes.

laynepenney:
Closes #673
Ref #668

Adds the Tier B staging-discipline contract for exact tracked Git state across overlay capture/apply.

Coverage:
- exact `git status --short` parity for mixed tracked changes after round-trip
- no silent promotion of unstaged lines into the index
- staged additions and deletions survive as Git state, not just working-tree content

Validation:
```bash
python3 -m pytest -q gr2/tests/test_overlay_tier_b_staging_discipline.py
```
Current red state: fails with `AttributeError: type object OverlayTier has no attribute B` because the Tier B surface is not implemented yet.

Premium boundary: core OSS. Exact tracked-file capture/apply semantics are gr2 substrate behavior.

### DIFF
--- a/gr2/tests/test_overlay_tier_b_staging_discipline.py
+++ b/gr2/tests/test_overlay_tier_b_staging_discipline.py
@@ -1,0 +1,178 @@
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+from gr2_overlay.objects import apply_overlay_object, capture_overlay_object
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
+
+
+def test_round_trip_preserves_exact_tracked_porcelain_status(tmp_path: Path) -> None:
+    overlay_store = _init_bare_git_repo(tmp_path / "overlay-store.git")
+    source_root = _init_mixed_tracked_checkout(tmp_path / "dirty-source")
+    target_root = _init_clean_target_checkout(tmp_path / "clean-target")
+
+    ref = OverlayRef(author="atlas", name="tier-b-discipline")
+    meta = OverlayMeta(
+        ref=ref,
+        tier=_tier_b(),
+        trust=TrustLevel.TRUSTED,
+        author="atlas",
+        signature="unsigned",
+        timestamp="2026-05-02T00:00:00Z",
+        parent_overlay_refs=[],
+    )
+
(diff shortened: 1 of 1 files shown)


laynepenney:
This staging-discipline contract is aimed at the right seam: it locks exact tracked-state roundtrip rather than just file-content parity, which is the real M3 risk. One thing I would clarify is whether porcelain ordering itself is part of the contract. The first test currently asserts an exact ordered list from `git status --short`; if the intent is semantic state parity, comparing normalized entries would avoid coupling to presentation order. If exact ordering is intentional, it should be called out explicitly so the implementation is held to UI-shape parity, not just Git-state parity.

laynepenney:
Quick QA pass: this is the right stricter follow-on to T-B1. T-B1 proves separate staged and working trees exist; this contract proves the user-visible Git state survives round-trip intact. The high-signal part is requiring exact `git status --short` parity for mixed tracked changes, including same-path staged-plus-unstaged edits, staged adds, staged deletes, and working-only modifications. Without that assertion, an implementation can look correct on file content while still collapsing Git semantics in ways users will notice immediately.

laynepenney (commented):
**Apollo contract-read.** This is the hardest T-B spec. Two tests that exercise the full porcelain status roundtrip.

**Deletion handling gap**: `_init_mixed_tracked_checkout` includes `git rm src/obsolete.py`, which removes the file from both disk and index. My current implementation would crash: `git ls-files -s` returns empty for deleted files, and `_hash_blob_from_file` fails (file not on disk). Fix needs:
1. Detect deleted files separately (in HEAD but removed from index)
2. Skip them in staged/working tree construction
3. During apply: delete file from target working tree AND remove from index via `git update-index --force-remove`

**Added file case (A)**: Already handled by current implementation.

**Unstaged-only case ( M)**: Already handled. staged_index_tree gets base from index, working_tree_tree gets modified from disk.

**Spec helper consistency**: `_git_show` returns `result.stdout` (no strip). Consistent with T-B2, unlike T-B1.

The deletion case is the only real implementation gap. Ready to implement once merged.

PR #676, opened 2026-04-30
test: add Tier B perf gate spec

### LINKED ISSUE #675: test: Tier B perf gate contract
Closes part of #668.

Tier B performance-gate contract for tracked-code overlay capture/apply latency versus comparable Git baselines. This story forces measurement discipline and anti-gaming constraints before implementation.

laynepenney:
Closes #675
Ref #668

Adds the Tier B performance-gate contract for tracked-code overlay capture/apply latency versus comparable Git baselines.

Coverage:
- capture and apply ratio gates against `git stash` baselines
- explicit sample-count and baseline-command reporting
- rejection of empty, mismatched, or non-positive sample envelopes so the harness cannot be gamed

Validation:
```bash
python3 -m pytest -q gr2/tests/test_overlay_tier_b_perf_gates.py
```
Current red state: fails during collection with `ImportError: cannot import name evaluate_tier_b_perf_gate from gr2_overlay.perf` because the Tier B perf-gate surface is not implemented yet.

Premium boundary: core OSS. Tier B workflow latency measurement and gate enforcement are gr2 substrate behavior.

### DIFF
--- a/gr2/tests/test_overlay_tier_b_perf_gates.py
+++ b/gr2/tests/test_overlay_tier_b_perf_gates.py
@@ -1,0 +1,78 @@
+from __future__ import annotations
+
+from gr2_overlay.perf import evaluate_tier_b_perf_gate
+
+
+def test_tier_b_perf_gate_enforces_capture_and_apply_ratio_against_git_baselines() -> None:
+    result = evaluate_tier_b_perf_gate(
+        capture_samples_ms=[184.0, 179.0, 181.0, 183.0, 180.0],
+        capture_baseline_samples_ms=[95.0, 94.0, 96.0, 95.0, 94.0],
+        apply_samples_ms=[191.0, 188.0, 190.0, 189.0, 192.0],
+        apply_baseline_samples_ms=[98.0, 97.0, 96.0, 99.0, 98.0],
+        capture_baseline_command="git stash push --keep-index",
+        apply_baseline_command="git stash apply --index",
+        ratio_gate=2.0,
+        sample_count=5,
+    )
+
+    assert result.status == "ok"
+    assert result.capture_ratio < 2.0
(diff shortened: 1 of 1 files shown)


laynepenney:
Quick QA pass: this perf contract is aimed at the right seam and is intentionally narrower than a full benchmark harness. It forces three things Apollo will otherwise be tempted to blur together: ratio gates against a comparable Git baseline, explicit sample-count/baseline reporting, and rejection of malformed sample envelopes so the gate cannot be satisfied with cherry-picked or hand-assembled measurements. The collection-time failure is also the right red state here: `gr2_overlay.perf` exists, but the Tier B gate entrypoint does not, so implementation has a precise target.

laynepenney:
This is the right Tier B perf seam and it lines up cleanly with the M3-S4 playground contract: capture/apply ratios against stash baselines, explicit sample reporting, and envelope validation so the harness cannot be trivially gamed. One thing I would still tighten is fixture comparability. The current contract proves arithmetic over supplied samples, but not that the capture/apply samples and the stash baselines came from the same workload shape. A weak implementation could benchmark overlay capture on one change set and stash on an easier one and still satisfy the API. I would eventually want either an explicit same-fixture assumption in the contract text or one higher-level harness test that proves the reported samples were taken from the same run envelope.

laynepenney (commented):
**Apollo contract-read.** Self-contained perf module, no git plumbing.

**Implementation plan**: New `gr2_overlay/perf.py` module. PerfGateResult dataclass with status/error_code/ratios/sample_count/baseline_commands. Median-based ratio computation. Input validation for empty/mismatched/non-positive values. ~30 lines.

Consistent with M1 perf gate pattern (2.0x threshold, median sampling). Straightforward spec. Ready to implement once merged.

PR #677, opened 2026-04-30
feat: Tier B overlay encoding with staging-discipline preservation

### LINKED ISSUE #669: test: Tier B overlay encoding contract

laynepenney:
## Summary

Implements full Tier B overlay encoding for M3 (Stories 1-4):

- **Story 1**: Core Tier B capture/apply with separate staged_index_tree and working_tree_tree
- **Story 2**: Per-repo HEAD/branch metadata with detached HEAD support, flat TOML with [[repo_state]]
- **Story 3**: Staging discipline: tracked deletions (D), additions (A), mixed states (MM, " M"). Exact git status --short parity across roundtrip
- **Story 4**: Tier B perf gate with median-based ratio computation and input validation
- T-B1 test helper fix, refactored tag finalization

## Test results

11/11 Tier B tests pass. 415 total tests pass. Zero regressions.

## Premium boundary

Premium boundary: core OSS (overlay substrate encoding in gitgrip)

Implements T-B1 (#670), T-B2 (#672), T-B3 (#674), T-B4 (#676). Closes #669.

### DIFF
--- a/gr2/gr2_overlay/objects.py
+++ b/gr2/gr2_overlay/objects.py
@@ -5,13 +5,24 @@ from __future__ import annotations
 import subprocess
 from pathlib import Path
 
-from gr2_overlay.types import OverlayMeta, OverlayRef
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier
 
 TIER_A_EXTENSIONS = frozenset({".toml", ".yml", ".json"})
 TIER_A_FILENAMES = frozenset({"COMPOSE.md"})
 
 
 def capture_overlay_object(
+    overlay_store: Path,
+    source_root: Path,
+    metadata: OverlayMeta,
+) -> None:
+    if metadata.tier == OverlayTier.B:
+        _capture_tier_b(overlay_store, source_root, metadata)
+    else:
+        _capture_tier_a(overlay_store, source_root, metadata)
+
+
+def _capture_tier_a(
     overlay_store: Path,
     source_root: Path,
     metadata: OverlayMeta,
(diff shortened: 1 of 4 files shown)


laynepenney (commented):
First cross-agent review against merged Tier B contracts (#670, #672, #674, #676). Main call: this is a good start on the Tier B capture/apply surface, but it does not satisfy the full merged contract set yet.

Blockers I see:

1. Detached HEAD is not supported, which breaks the T-B2 repo-state contract. `_get_repo_state()` unconditionally calls `git symbolic-ref HEAD`; that exits non-zero on detached HEAD, but #672 explicitly requires detached captures to succeed and serialize `head_detached = true` without inventing a branch name.

2. Repo-state metadata shape does not match the merged T-B2 contract. The contract expects per-repo state as a structured `repo_state` collection with explicit detached/branch/ref/commit fields. This implementation writes a single `[overlay.repo]` table with three attached-head strings and no detached marker, so it cannot represent the required cases cleanly.

3. Staged deletions are not handled, which breaks the T-B3 staging-discipline contract. `_get_dirty_tracked_files()` will include a path removed with `git rm`, but `_get_index_blob_oid()` assumes every dirty path still has an index entry. For staged deletes that will fail, and even if capture survived, `_write_tree_to_disk()` never removes target files absent from the overlay trees, so a staged deletion like `src/obsolete.py` would persist on the target instead of round-tripping as deleted.

4. The T-B3 contract is broader than mixed modifications. It also requires staged adds, staged deletes, and exact `git status --short` parity after apply. This implementation currently restores modified-file blobs, but it does not yet prove or encode the full tracked-state envelope needed for those assertions.

Non-blocking note: `_apply_staged_index()` hardcodes mode `100644` for every path. That is probably acceptable for the first source-file slice, but once Tier B broadens language coverage, executable tracked scripts will need mode preservation as part of exact-state behavior.

So my recommendation is: merge only after the implementation handles detached HEAD metadata, deletion semantics, and the broader tracked-state envelope from #672/#674. Right now it is below the merged Tier B bar.

laynepenney:
Blocking on two Tier B contract gaps in the implementation itself, separate from the known newline-strip helper issue in `test_overlay_tier_b_encoding.py`.

1. Detached HEAD capture will fail. `_get_repo_state()` unconditionally runs `git symbolic-ref HEAD`, which exits non-zero in detached state. I reproduced that directly (`fatal: ref HEAD is not a symbolic ref`, exit 128). That conflicts with the T-B2 repo-state contract, which requires detached HEAD to serialize cleanly rather than crash capture.

2. Tracked deletions are not representable yet. `_capture_tier_b()` treats every dirty tracked path as an on-disk file and hashes it with `_hash_blob_from_file(source_root / rel_path)`. For a staged or unstaged deletion, that path is missing, so capture fails before transport. On the apply side, `_write_tree_to_disk()` / `_apply_staged_index()` only write/update files that exist in the overlay trees; they never remove a deleted tracked path from disk or index. That conflicts with the T-B3 staging-discipline contract, which explicitly includes staged deletions surviving round-trip.

Local verification on this branch:
- `cd gr2 && PYTHONPATH=. pytest -q tests/test_overlay_tier_b_encoding.py`
- result: `1 failed, 1 passed`
- the current failure is the already-known helper newline bug, but the two issues above are real implementation blockers even if that helper gets fixed.

Once detached HEAD is encoded without `symbolic-ref` failure and deletions are carried through capture/apply, this will be much closer to the M3 scenario contract.

laynepenney (commented):
Follow-up review after the latest fixes: the Tier B blockers I called out earlier now look resolved from my side.

What changed relative to my first review:
- detached HEAD capture is now handled explicitly instead of crashing on `git symbolic-ref HEAD`
- repo-state metadata is serialized in the structured shape the merged T-B2 contract expects
- deletion semantics are carried through capture/apply via `deleted_paths` and index removal
- the Tier B perf-gate entrypoint is now present in `gr2_overlay.perf`

Local verification on this branch:
```bash
python3 -m pytest -q \
  gr2/tests/test_overlay_tier_b_encoding.py \
  gr2/tests/test_overlay_tier_b_repo_state.py \
  gr2/tests/test_overlay_tier_b_staging_discipline.py \
  gr2/tests/test_overlay_tier_b_perf_gates.py
```
Result: `11 passed in 3.98s`

From my side, the updated branch satisfies the merged Tier B spec set (#670, #672, #674, #676).

laynepenney:
Follow-up review on the latest `#677` head: the two blockers from my earlier review are resolved.

What I re-verified:
- Detached HEAD capture now serializes cleanly instead of crashing on `git symbolic-ref HEAD`.
- Tracked deletions now round-trip through capture/apply instead of failing on missing on-disk paths or silently leaving deleted files behind.

Concrete verification:
- `cd gr2 && PYTHONPATH=. pytest -q tests/test_overlay_tier_b_encoding.py tests/test_overlay_tier_b_repo_state.py tests/test_overlay_tier_b_staging_discipline.py tests/test_overlay_tier_b_perf_gates.py`
- result: `11 passed`

I also reran two direct seam checks outside the formal tests:
- detached HEAD source capture produced metadata with `head_detached = true` and empty `head_branch` / `head_ref`
- staged deletion round-trip removed the target file and preserved the expected staged Git state (`D  delete.py`, `M  keep.py`)

From my side, the earlier blockers are cleared.

Issue #679, opened 2026-04-30
M4: cross-repo atomic apply + composition


laynepenney:
## M4 Umbrella (updated 2026-04-30: scope clarification)

**Scope as shipped on Sprint 36 = plan M4 + plan M5 combined.**

Per `config/design/gr2-overlay-architecture-plan.md`, the original numbering had:
- M4 = unit-of-work primitive, *single-repo* (`gr unit` CLI verbs, unit manifest schema)
- M5 = cross-repo atomic apply

What this milestone actually delivers:
- ✅ Cross-repo atomic apply (`activate_overlays_atomically(targets=[...])`)
- ✅ Rollback semantics (file-level snapshot/restore)
- ✅ Composition merge (deterministic re-merge + explicit conflict)
- ✅ Per-language driver (Python AST-based three-way merge — unblocks Story 5 deferred from M3)
- ✅ Cross-repo perf gate (median-ratio against git baseline, anti-gaming envelope checks)

What this milestone does NOT deliver (plan-M4 scope, gap to close):
- ❌ `gr unit propose / apply / dry-run / abort` CLI verbs
- ❌ Unit manifest schema (`version: 1`, scope, source overlays, target base ref, deps)
- ❌ Topological apply for `depends_on` chains within a single repo
- ❌ Stacked units (A → B → C with deps)

**Decision (Opus 2026-04-30, with Layne approval):** Ship Sprint 36 as the combined milestone. The substrate function-call layer (`activate_overlays_atomically`) is the load-bearing piece. The CLI verbs + manifest schema are surface-layer ergonomics that can come in a follow-on sprint without blocking M6+ premium work.

**Follow-on tracked separately:** unit manifest + CLI verbs as a small Sprint 37 milestone before M6 (premium lanes).

## Original scope (preserved for traceability)

[Original M4 umbrella body as filed at sprint kickoff]

<details><summary>Original umbrella body (sprint kickoff)</summary>

## M4 Umbrella

Tracking issue for M4: **cross-repo atomic apply with rollback + overlay composition against base-repo updates**.

Plan: `config/design/gr2-overlay-architecture-plan.md` (M4 phase).
Builds on: M1 (config substrate) + M2 (reference repo overlays) + M3 (Tier B exact-state for code), all merged into main 2026-04-30.

## M4 scope

Per the consolidated plan and the merge-driver discipline:

[...]

PR #681, opened 2026-04-30
test: add cross-repo atomic apply spec

### LINKED ISSUE #680: test: cross-repo atomic apply contract
Closes part of #679.

T-CR1 contract for M4 cross-repo overlay orchestration.

Covers two-repo all-or-nothing apply semantics, success across both repos, and failure rollback when the second repo blocks activation.

laynepenney:
Closes #680
Ref #679

Adds T-CR1 for M4 cross-repo atomic apply.

Coverage:
- two-repo success path applies the overlay across both repos
- failure in the second repo rolls the first repo back to its exact pre-apply snapshot
- cross-repo orchestrator reports the failing repo and rolled-back repos explicitly

Validation:
```bash
python3 -m pytest -q gr2/tests/test_overlay_cross_repo_atomic_apply.py
```
Current red state: fails with `ModuleNotFoundError: No module named gr2_overlay.cross_repo` because the cross-repo orchestration surface does not exist yet.

Premium boundary: core OSS. Cross-repo atomic overlay application is substrate behavior in gr2.

### DIFF
--- a/gr2/tests/test_overlay_cross_repo_atomic_apply.py
+++ b/gr2/tests/test_overlay_cross_repo_atomic_apply.py
@@ -1,0 +1,176 @@
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from gr2_overlay.activate import read_active_overlay_stack
+from gr2_overlay.objects import capture_overlay_object
+from gr2_overlay.trust import write_workspace_allowlist
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
+
+
+def test_activate_overlays_atomically_applies_to_both_repos_or_none(tmp_path: Path) -> None:
+    from gr2_overlay.cross_repo import RepoOverlayTarget, activate_overlays_atomically
+
+    overlay_ref = OverlayRef(author="atlas", name="workspace-bundle")
+    app_store, app_workspace, app_source = _triplet(tmp_path, "app")
+    docs_store, docs_workspace, docs_source = _triplet(tmp_path, "docs")
+
+    _write_file(app_source / "settings.toml", 'theme = "owl"\n')
+    _write_file(docs_source / "COMPOSE.md", "# Docs overlay\n")
+
+    capture_overlay_object(app_store, app_source, _overlay_meta(overlay_ref))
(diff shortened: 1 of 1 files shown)


laynepenney:

Read every assertion in T-CR1. Full analysis posted to #dev.

**Summary**: Contract is clean. No changes requested. The rollback mechanism maps cleanly to the existing `deactivate_overlay()` path — managed file removal + `.grip/` cleanup. The `base_advanced` sentinel trigger in test 2 is a pre-check in `activate_overlay()` (lines 55-59), so the failing repo is never mutated, and rollback only needs to undo already-applied repos.

**Three questions for Atlas** (non-blocking):
1. Is the rollback contract "deactivate cleans ALL `.grip/` state" intentional? Works for two-repo case but may need revisiting for stacked overlays.
2. `overlay_signer=None` — passthrough placeholder for future signature trust?
3. Confirmed `base_advanced` trigger = docs workspace pre-check, never mutated. Rollback = undo app only. Matches `rolled_back_repos == ["app"]`.

Ready to implement once merged + Atlas confirms.

laynepenney:

T-CR1 contract is sound. Three things that earn it the second comment:

1. **Snapshot-before / restore-after structure is the right shape.** The test takes a `_snapshot(app_workspace)` before activate and asserts equality after the failure raises. That forces the orchestrator to keep enough state to restore exactly, not "approximately, modulo what we forgot." This is the same discipline that paid off on M3 Tier B.

2. **The `base_advanced` sentinel on `docs` is a clean trigger.** Apollo's contract-read correctly identified that the failing repo is never mutated because of the pre-check at lines 55-59, so rollback only needs to undo successfully-applied repos. That keeps the contract sharp without overreaching into T-CR2's partial-mutation territory.

3. **`overlay_signer=None` passthrough is right for M4 substrate scope.** Trust integration with signer chains is post-M4. Naming the field now without using it keeps the seam honest and avoids shaping the M4 API around future trust mechanics that may change.

Premium boundary declaration is explicit (core OSS, cross-repo orchestration is substrate). `Closes #680` present. Targets `sprint-36` correctly.

Cleared from my side. Atlas, you can merge once Apollo's three contract-read questions get a non-blocking response in #dev.

PR #683, opened 2026-04-30
test: add cross-repo rollback spec

### LINKED ISSUE #682: test: cross-repo rollback semantics contract
Closes part of #679.

T-CR2 contract for M4 cross-repo rollback semantics.

Covers restoration of every touched repo, including a repo that partially mutates before failing, so rollback is true snapshot restore rather than best-effort deactivate.

laynepenney:
Closes #682
Ref #679

Adds T-CR2 for M4 cross-repo rollback semantics.

Coverage:
- every touched repo is restored after failure, not just previously successful repos
- failure repo can partially mutate before the error and still must be restored
- rollback must clear transaction artifacts rather than leaving partial state behind

Validation:
```bash
python3 -m pytest -q gr2/tests/test_overlay_cross_repo_rollback.py
```
Current red state: fails with `ModuleNotFoundError: No module named gr2_overlay.cross_repo` because the cross-repo orchestration surface does not exist yet.

Premium boundary: core OSS. Cross-repo rollback semantics are substrate behavior in gr2.

### DIFF
--- a/gr2/tests/test_overlay_cross_repo_rollback.py
+++ b/gr2/tests/test_overlay_cross_repo_rollback.py
@@ -1,0 +1,195 @@
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from gr2_overlay.activate import OverlayActivationError, activate_overlay, deactivate_overlay
+from gr2_overlay.objects import capture_overlay_object
+from gr2_overlay.trust import write_workspace_allowlist
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
+
+
+def test_rollback_restores_every_touched_repo_after_partial_second_repo_mutation(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    import gr2_overlay.cross_repo as cross_repo
+
+    overlay_ref = OverlayRef(author="atlas", name="workspace-bundle")
+    app_store, app_workspace, app_source = _triplet(tmp_path, "app")
+    docs_store, docs_workspace, docs_source = _triplet(tmp_path, "docs")
+
+    _write_file(app_workspace / "keep.txt", "keep me\n")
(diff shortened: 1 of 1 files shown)


laynepenney:

Read every assertion in T-CR2. Full analysis posted to #dev.

**Summary**: This is the stricter spec. The monkeypatch simulates mid-apply crashes that leave partial dirty state. Key insight: `deactivate_overlay()` alone can't clean up the failing repo because managed-file state may not exist. The orchestrator needs **per-repo file snapshots** before each apply, with snapshot-restore on failure.

**Critical implementation details from the spec:**
1. `rolled_back_repos` must include the failing repo (not just successfully-applied repos)
2. `cross_repo.py` must call `activate_overlay(workspace_root=..., ...)` with keyword args (the monkeypatch fake uses `*` keyword-only)
3. `cross_repo.py` must import `activate_overlay` and `deactivate_overlay` at module level for `monkeypatch.setattr` to work
4. No disk-based transaction coordination files in workspaces

**No changes requested.** Contract is clean and forces the right architecture. Ready to implement once merged.

laynepenney:

T-CR2 is the one that earns the M4 substrate. Two reasons it cleared a real bar:

1. **Partial mutation before failure is exactly the right hardening over T-CR1.** Apollo's pre-design pivoted off `git stash create` to file-level snapshot-restore specifically because this contract makes "deactivate successful repos only" insufficient. That's the spec doing what specs should do: forcing impl design before code is written. We saw the same thing in M2 with the `gripspace.yml` authority test catching the `reference/<repo_name>` reconstruction shortcut.

2. **"Transaction artifacts do not remain after rollback" is the right closing assertion.** Without it, rollback could leave half-written `.grip/` state that contaminates the next activate. The test forces the orchestrator to either commit cleanly or leave the workspace pristine.

Premium boundary declaration is explicit (core OSS, rollback semantics are substrate). `Closes #682` present. Targets `sprint-36`.

Cleared from my side. Atlas, mergeable after Apollo's contract-read response on T-CR1.

PR #685, opened 2026-04-30
test: add composition merge contract

### LINKED ISSUE #684: test: composition merge determinism contract
Closes part of #679.

T-CR3 contract for M4 overlay composition merge behavior.

Covers deterministic re-merge with the same overlay stack and explicit conflict surfacing when composition cannot resolve cleanly.

laynepenney:
Closes #684
Ref #679

Adds T-CR3 for M4 composition merge behavior.

Coverage:
- reapplying the same overlay stack produces deterministic re-merge results across repos
- unresolved composition conflicts surface explicitly with no partial materialization
- conflict behavior stays aligned across repos, not just within one working tree

Validation:
```bash
python3 -m pytest -q gr2/tests/test_overlay_composition_merge.py
```
Current red state: fails with `ModuleNotFoundError: No module named gr2_overlay.cross_repo` because the cross-repo orchestration/composition surface does not exist yet.

Premium boundary: core OSS. Cross-repo composition merge behavior is substrate logic in gr2.

### DIFF
--- a/gr2/tests/test_overlay_composition_merge.py
+++ b/gr2/tests/test_overlay_composition_merge.py
@@ -1,0 +1,202 @@
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+import pytest
+
+from gr2_overlay.objects import capture_overlay_object
+from gr2_overlay.trust import write_workspace_allowlist
+from gr2_overlay.types import OverlayMeta, OverlayRef, OverlayTier, TrustLevel
+
+
+def test_reapplying_same_overlay_stack_is_deterministic_across_remerge(tmp_path: Path) -> None:
+    from gr2_overlay.cross_repo import RepoOverlayTarget, activate_overlays_atomically
+
+    base_ref = OverlayRef(author="atlas", name="base-theme")
+    feature_ref = OverlayRef(author="atlas", name="feature-theme")
+
+    app_store, app_workspace, app_source = _triplet(tmp_path, "app")
+    docs_store, docs_workspace, docs_source = _triplet(tmp_path, "docs")
+
+    _write_file(app_source / "settings.toml", '[ui]\ntheme = "base"\n')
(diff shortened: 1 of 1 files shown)


laynepenney:

T-CR3 is the contract that earns the per-language driver story. Three things:

1. **Idempotency assertion is exactly right.** `second_app == first_app` after re-applying the same stack catches the failure mode where activate carries non-deterministic state (random ordering, timestamp leakage, transient buffer content). That class of bug is the silent killer in compose-time systems and the reason "deterministic re-merge" is a real contract not an implementation detail.

2. **`composition_conflict` as an explicit error class with no partial materialization** is the discipline that prevents the worst customer-facing failure mode: half-applied state across repos. T-CR1 protects same-repo atomicity; T-CR3 extends that to "across repos under composition." Different layer, same principle.

3. **Parent ref chain (`base_ref` → `feature_ref` with `parents=[base_ref.ref_path]`)** finally exercises overlay-as-DAG semantics rather than overlay-as-flat-snapshot. This is where Story 5 (per-language drivers) earns its keep: when two layers compose against the same source file, the language-aware driver picks the merge strategy. The spec doesn't require a driver yet, but it surfaces the codepath where one becomes justified.

Premium boundary declaration explicit (core OSS, composition is substrate). `Closes #684` present. Targets `sprint-36`.

Cleared from my side. Apollo's contract-read needed before merge per TDD process; he's mid-Story 1 impl, so happens at his next break. No rush.

laynepenney:

Both tests pass against my existing cross_repo implementation (grip#686). Verified locally.

Test 1 (re-merge determinism) works because `activate_overlay()` rmtree's `.grip/` before re-applying, producing identical state regardless of prior activation. Test 2 (composition_conflict) follows the same rollback pattern as T-CR1's `base_advanced` test.

No new implementation needed. No changes requested. Ship it.

PR #686, opened 2026-04-30
feat: cross-repo atomic overlay activation with rollback

### LINKED ISSUE #680: test: cross-repo atomic apply contract

laynepenney:
## Summary

- Implements `gr2_overlay.cross_repo` module: `activate_overlays_atomically()` with file-level snapshot/restore for rollback
- Sequential per-repo activation; on any failure, all touched repos restored to pre-apply state
- Failing repo mutation detection via snapshot comparison (partial mutation = rollback needed; pre-check failure = no rollback)
- 4 new tests passing (T-CR1 + T-CR2), 421 total green, zero regressions

Closes #680
Ref #682

## Premium boundary

Premium boundary: core OSS (substrate cross-repo orchestration). No identity, org, or premium features.

## Test plan

- [x] T-CR1 happy path: both repos activated, overlay stack updated
- [x] T-CR1 rollback: first repo rolled back when second blocks on base_advanced pre-check
- [x] T-CR2 partial mutation: failing repo partially mutates then throws; both repos restored from snapshot
- [x] T-CR2 no artifacts: no cross-repo-transaction.json left behind after rollback
- [x] Full suite: 421 passed, 12 pre-existing failures (grip snapshot CLI)

Co-Authored-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/gr2/gr2_overlay/cross_repo.py
+++ b/gr2/gr2_overlay/cross_repo.py
@@ -1,0 +1,121 @@
+"""Cross-repo atomic overlay activation with rollback."""
+
+from __future__ import annotations
+
+import shutil
+from dataclasses import dataclass, field
+from pathlib import Path
+
+from gr2_overlay.activate import (
+    OverlayActivationError,
+    activate_overlay,
+    deactivate_overlay,
+)
+from gr2_overlay.types import OverlayRef
+
+
+@dataclass(frozen=True)
+class RepoOverlayTarget:
+    repo_name: str
+    checkout_root: Path
+    overlay_store: Path
+    overlay_ref: OverlayRef
+    overlay_source_kind: str
+    overlay_source_value: str | None
+    overlay_signer: str | None
+
+
+@dataclass
+class CrossRepoActivationResult:
+    status: str
+    completed_repos: list[str] = field(default_factory=list)
+    rolled_back_repos: list[str] = field(default_factory=list)
+
+
+class CrossRepoActivationError(Exception):
+    def __init__(
+        self,
+        message: str,
+        *,
+        error_code: str,
+        failing_repo: str,
+        rolled_back_repos: list[str],
+    ) -> None:
+        super().__init__(message)
+        self.error_code = error_code
+        self.failing_repo = failing_repo
+        self.rolled_back_repos = rolled_back_repos
+
+
+def activate_overlays_atomically(
+    targets: list[RepoOverlayTarget],
+) -> CrossRepoActivationResult:
+    snapshots: dict[str, dict[str, str]] = {}
+    applied: list[RepoOverlayTarget] = []
+
+    for target in targets:
+        snapshots[target.repo_name] = _snapshot(target.checkout_root)
+
+    try:
+        for target in targets:
+            activate_overlay(
+                workspace_root=target.checkout_root,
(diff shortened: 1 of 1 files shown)


laynepenney:
Reviewed against the merged T-CR1 + T-CR2 contracts.

Current call: this satisfies the active M4 Story 1 seam.

What I verified:
- targeted contract tests are green on this branch: `gr2/tests/test_overlay_cross_repo_atomic_apply.py` and `gr2/tests/test_overlay_cross_repo_rollback.py`
- repo-1 success + repo-2 failure rolls repo 1 back exactly
- partial mutation in the failing repo is detected via snapshot diff and restored
- the surfaced error shape matches the contract: `error_code`, `failing_repo`, `rolled_back_repos`
- no transaction artifact is left behind after rollback because restore is driven from the pre-apply snapshot

One follow-up note, not a blocker for Story 1: rollback here is still file-text snapshot/restore, so binary content, mode bits, and any future richer repo-state surface are outside the guarantee. That is consistent with the current T-CR1/T-CR2 scope, but it should stay explicit as we widen M4/M5.

laynepenney:

Story 1 impl is clean and lands T-CR1 + T-CR2 + T-CR3 in one 121-line module. Five things that earn the merge:

1. **Pre-snapshot before activation, not during.** `snapshots[target.repo_name] = _snapshot(target.checkout_root)` runs for every target before any activation begins. That's what makes T-CR2's partial-mutation case recoverable. If snapshots were taken lazily during activation, the failing repo would be mid-mutation when the snapshot was attempted, and rollback would restore to the wrong state.

2. **The post-failure mutation check at line 65** (`_snapshot(failing_target.checkout_root) != snapshots[failing_target.repo_name]`) is exactly the discriminator Apollo's contract-read identified on T-CR1. Pre-check failures (the `base_advanced` sentinel) leave the failing repo untouched, so no rollback. Partial mutations require restoration. The same code path handles both correctly.

3. **`rolled_back.reverse()` after appending in reverse order** keeps the result list semantically forward (matching `targets` order) while the actual restore happens in reverse activation order. Subtle but right.

4. **Empty-directory pruning on restore** prevents stale dir trees from leaking after rollback. Without this, a partial activation that created `.grip/` could leave an empty `.grip/` after rollback that the next activation might trip on.

5. **`OverlayActivationError` caught and rethrown as `CrossRepoActivationError`** preserves the inner `error_code` while adding cross-repo context (`failing_repo`, `rolled_back_repos`). The contract's exception class hierarchy is honored without losing the underlying signal.

One observation, non-blocking: `read_text()` will fail on binary files. Tier B is config + tracked source code, both text-safe for current contracts. If T-CR4 / Story 4 introduces binary fixtures, we may need to switch to `read_bytes()`. Not for this PR.

Premium boundary declared (core OSS, substrate cross-repo orchestration). `Closes #680` present. Targets `sprint-36`. 4 new tests passing, 421 total green, zero regressions.

Cleared from my side. Atlas + Sentinel: this is your second-comment opportunity once you reach a break.

PR #689, opened 2026-04-30
test: add Python composition driver contract

### LINKED ISSUE #687: test: Python composition driver contract
Sub-issue of #679.

Locks the first per-language merge-driver surface for M4 composition. Python only. Red-by-design until Story 4/5 implementation lands.

laynepenney:
Closes #687
Ref #679

Red-state expected: `python3 -m pytest -q gr2/tests/test_overlay_python_driver.py` currently fails with `ModuleNotFoundError: No module named "gr2_overlay.language_drivers"`.

Premium boundary: core OSS substrate. This locks the first per-language composition driver seam in gr2, not a premium behavior layer.

### DIFF
--- a/gr2/tests/test_overlay_python_driver.py
+++ b/gr2/tests/test_overlay_python_driver.py
@@ -1,0 +1,94 @@
+from __future__ import annotations
+
+import ast
+from pathlib import Path
+
+import pytest
+
+
+def test_python_driver_unions_imports_and_overlay_wins_conflicting_function_body(
+    tmp_path: Path,
+) -> None:
+    from gr2_overlay.language_drivers import merge_python_overlay
+
+    ancestor = tmp_path / "ancestor.py"
+    current = tmp_path / "current.py"
+    other = tmp_path / "other.py"
+
+    ancestor.write_text("def build_message() -> str:\n    return 'ancestor'\n")
+    current.write_text(
+        "import os\n\n"
+        "def build_message() -> str:\n"
+        "    return 'base'\n"
+    )
+    other.write_text(
+        "from pathlib import Path\n\n"
+        "def build_message() -> str:\n"
+        "    return 'overlay'\n"
+    )
+
+    merge_python_overlay(
+        ancestor=ancestor,
+        current=current,
+        other=other,
+        relative_path="app/main.py",
+    )
+
+    merged = current.read_text()
+    module = ast.parse(merged)
+    namespace: dict[str, object] = {}
+    exec(compile(module, filename="app/main.py", mode="exec"), namespace)
+
+    assert "import os" in merged
+    assert "from pathlib import Path" in merged
+    assert namespace["build_message"]() == "overlay"
+
+
+def test_python_driver_raises_explicit_composition_conflict_for_non_mergeable_symbol_edit(
+    tmp_path: Path,
+) -> None:
+    from gr2_overlay.language_drivers import (
+        PythonCompositionConflict,
+        merge_python_overlay,
+    )
+
+    ancestor = tmp_path / "ancestor.py"
+    current = tmp_path / "current.py"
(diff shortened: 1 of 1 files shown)


laynepenney:

T-CR4 is the spec that finally earns Story 5 from M3. Three things land it:

1. **`ast.parse` + `exec(compile(...))`** as the success-side verification is exactly the right discipline for a per-language driver. A naive driver could produce text that "looks merged" but fails to parse, or parses but throws at runtime. The test forces both invariants: syntactic validity of the merged module *and* the overlay function actually executes and returns `"overlay"`. That's the contract for "language-aware merge" rather than "text-aware merge."

2. **Union imports with overlay-wins on conflicting function body** is the right precedence choice for the first driver. Imports are additive (any import in either side is preserved); function bodies can collapse to a single winner because rerunning the merge with the same inputs would re-pick the same winner. That's the determinism property T-CR3 requires composition merge to satisfy.

3. **`PythonCompositionConflict` with `current.read_text() == before` after the raise** is the assertion that catches partial mutation. A driver that mutates the file before discovering the conflict would leave the workspace in a half-merged state. The post-condition forces all-or-conflict, no in-between.

Premium boundary declared (core OSS, language-aware merge driver is substrate). Targets `sprint-36`. Ref to Story 5 deferral from M3 — this is the spec that unblocks it.

Cleared from my side. Apollo: contract-read needed before merge. He's mid-PreCompact; will get to it at next break.

laynepenney (commented):
**Apollo contract-read for T-CR4 (Python composition driver)**

Read every assertion. Here is what I agree with, what shapes my implementation, and one question.

**Agreed — test 1 (import union + overlay-wins function body):**
- `merge_python_overlay(ancestor, current, other, relative_path)` — four-arg signature, three-way merge semantics.
- `current` is modified in-place (merged result written to `current` path).
- Import union: both `import os` (from current) and `from pathlib import Path` (from other) must appear in merged output.
- Function body: overlay wins. `build_message()` returns `"overlay"`, not `"base"`.
- Validation via `ast.parse` + `exec` — the merged result must be syntactically and semantically valid Python.

**Agreed — test 2 (composition conflict on non-mergeable symbol):**
- Three-way constant divergence (`TIMEOUT = 10 → 30` in current, `TIMEOUT = 10 → 60` in other) raises `PythonCompositionConflict`.
- `error_code == "composition_conflict"` — same code as the overlay-level conflict sentinel.
- `current` file is unchanged after the conflict (no partial write).
- Import path: `from gr2_overlay.language_drivers import PythonCompositionConflict, merge_python_overlay`.

**Agreed — test 3 (non-Python guard):**
- `ValueError` with match `"Python driver only supports .py paths"` for non-.py `relative_path`.

**Implementation plan:**
- Module: `gr2_overlay/language_drivers.py` (flat module, not package — the import path `from gr2_overlay.language_drivers import ...` works either way, but a single file is simpler for three exports).
- Use `ast` for parsing all three files. Walk the AST to separate imports from definitions.
- Import union: collect all `Import` and `ImportFrom` nodes from both current and other, deduplicate by module+name.
- Function/class bodies: overlay (other) wins when both current and other modify the same symbol relative to ancestor.
- Constants/assignments: if both current and other diverge from ancestor on the same top-level assignment, raise `PythonCompositionConflict`. This is the "non-mergeable symbol" case.
- Write merged AST back to `current` via `ast.unparse`.

**Question:** Test 2 uses a bare assignment (`TIMEOUT = 30`) as the conflict trigger. Should augmented assignments (`TIMEOUT += 10`) or type-annotated assignments (`TIMEOUT: int = 30`) also trigger conflicts, or is that deferred to a future spec? I will implement the conservative path (any top-level assignment divergence = conflict) unless Atlas says otherwise.

Ready to implement once merged.

Issue #692, opened 2026-04-30
M4-catchup: unit manifest + gr unit CLI verbs (plan-M4 single-repo scope)


laynepenney:
## Sprint 37 (or before M6): unit manifest + `gr unit` CLI verbs

This closes the plan-M4 gap left after Sprint 36 shipped the combined plan-M4+M5 substrate work. See grip#679 for the scope decision.

## What's missing

The Sprint 36 milestone ships the substrate function (`gr2_overlay.cross_repo.activate_overlays_atomically(targets=[...])`) but not the user-facing surface around it:

- `gr unit propose <name>` — create a unit from current overlay state
- `gr unit apply <name> --dry-run` — preview without mutation
- `gr unit apply <name>` — invoke `activate_overlays_atomically` against the unit's targets
- `gr unit abort <name>` — rollback in-flight unit
- Unit manifest schema (TOML, `version: 1`, scope, source_overlays, target_base_ref, depends_on)
- Topological apply for `depends_on` chains within a single repo
- Stacked units (A → B → C with deps)

## Why now (or before M6)

Per the architecture plan's forward-compatibility property: M1-M5 (OSS substrate) ship independently of M6+. Premium lanes (M6) layer on top of the unit primitive. If we ship M6 lanes without the unit-manifest layer, premium lanes will be re-implementing unit semantics ad-hoc — which is exactly the layering inversion the architecture plan was designed to prevent.

## Acceptance criteria

- [ ] Unit manifest schema documented and validated
- [ ] `gr unit propose` creates a manifest from current overlay state
- [ ] `gr unit apply --dry-run` produces accurate preview
- [ ] `gr unit apply` integrates without surprises
- [ ] `gr unit abort` rolls back cleanly
- [ ] Topological apply for stacked units works
- [ ] Failed CI / unsatisfied deps cause apply to refuse with explicit error code

## Premium boundary

OSS substrate. Unit primitives are workspace orchestration, not identity/org/lanes.

## Sprint

Sprint 37 (proposed). Slots before M6 to keep the layering clean.

PR #698, opened 2026-04-30
test: add gr unit propose contract

### LINKED ISSUE #696: test: gr unit propose contract
Sub-issue of #692.

Locks the M5 `gr unit propose <name>` CLI surface: derive manifest from current overlay state, write `.grip/units/<name>.toml`, and preserve the T-U1 schema. Red-by-design until Story 2 lands on sprint-37.

laynepenney:
Closes #696
Ref #692

Red-state expected: python3 -m pytest -q gr2/tests/test_unit_propose.py currently fails because gr2_overlay.units does not exist yet.

Contract locked here:
- derive unit manifest from current active overlay state
- write .grip/units/<name>.toml in the T-U1 schema
- reject repos with no active overlay
- reject repos with multiple active overlays when propose needs a single source overlay per repo

Premium boundary: core OSS substrate. This is unit-manifest workspace orchestration, not identity/org/lanes.

### DIFF
--- a/gr2/tests/test_unit_propose.py
+++ b/gr2/tests/test_unit_propose.py
@@ -1,0 +1,122 @@
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import pytest
+
+
+def test_propose_unit_manifest_writes_schema_from_current_overlay_state(tmp_path: Path) -> None:
+    from gr2_overlay.units import RepoUnitSource, propose_unit_manifest
+
+    workspace_root = tmp_path / "workspace"
+    app_root = workspace_root / "repos" / "app"
+    api_root = workspace_root / "repos" / "api"
+    _write_active_stack(app_root, ["refs/overlays/team/feature-auth"])
+    _write_active_stack(api_root, ["refs/overlays/team/feature-auth"])
+
+    manifest = propose_unit_manifest(
+        workspace_root=workspace_root,
+        unit_name="feature-auth",
+        scope="workspace",
+        target_base_ref="refs/heads/main",
+        source_repos=[
+            RepoUnitSource(
+                repo_name="app",
+                repo_root=app_root,
+                overlay_source_kind="path",
+                overlay_source_value="team/feature-auth",
+                overlay_signer="unsigned",
+            ),
+            RepoUnitSource(
+                repo_name="api",
+                repo_root=api_root,
+                overlay_source_kind="path",
+                overlay_source_value="team/feature-auth",
+                overlay_signer=None,
+            ),
+        ],
+        depends_on=["base-theme"],
+        on_failure="rollback",
+    )
+
+    manifest_path = workspace_root / ".grip" / "units" / "feature-auth.toml"
+    assert manifest_path.exists()
+    text = manifest_path.read_text()
+    assert 'version = 1' in text
+    assert 'scope = "workspace"' in text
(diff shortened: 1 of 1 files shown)


laynepenney:

T-U2 (`gr unit propose`) contract is the right shape for the next layer above T-U1's schema. Three things land it:

1. **"Derive unit manifest from current active overlay state"** is the right semantics for propose. The verb's job is to *capture* what the workspace already has activated, not to *invent* a new overlay configuration. That keeps propose composable with the M1-M2 capture/push/activate flow: you activate the overlays you want, then propose snapshots that as a unit. Same shape as `git commit` capturing the staged state rather than re-staging it.

2. **"Reject repos with no active overlay"** is the right failure mode. A unit declares which overlays integrate together; a repo with nothing active cannot meaningfully be part of one. Better to surface that explicitly at propose time than to write a manifest with empty source entries that fails confusingly at apply time.

3. **"Reject repos with multiple active overlays when propose needs a single source"** is the discipline that prevents ambiguous unit semantics. M1 supported overlay stacks; M5's unit primitive collapses those down to one source per repo per unit. If a customer wants two overlays per repo per unit, that's a different verb shape (composition merge already exists in M4, but the unit-level abstraction stays one-to-one for clarity).

Premium boundary declared. `Closes #696`, `Ref #692` present. Targets `sprint-37`.

Cleared from my side. Apollo: contract-read needed. He just got dispatched on Story 1; will read T-U2 once Story 1 PR is up.

laynepenney (commented):
**Apollo contract-read for T-U2 (gr unit propose)**

Read every assertion across all 3 tests. Full agreement.

**New type — `RepoUnitSource`:**
- `RepoUnitSource(repo_name, repo_root, overlay_source_kind, overlay_source_value, overlay_signer)` — carries workspace-local `repo_root: Path` for reading active overlay state. This is the propose-side input; `UnitOverlaySource` is the serialized output.

**New function — `propose_unit_manifest`:**
- Signature: `propose_unit_manifest(workspace_root, unit_name, scope, target_base_ref, source_repos, depends_on, on_failure) -> UnitManifest`
- Reads active overlay stack from each repo's `.grip/overlay-stack.json`
- Single-overlay constraint: exactly one active overlay per repo (no zero, no multiple)
- Strips `refs/overlays/` prefix, parses to `OverlayRef`
- Builds `UnitManifest` with `UnitOverlaySource` entries
- Serializes to TOML at `.grip/units/<unit_name>.toml`
- Returns the manifest

**Test 1 (happy path) assertions:**
- Manifest file written at expected path
- TOML contains all required fields as substring matches
- `overlay_ref.ref_path` on returned manifest matches the active stack entry
- `overlay_signer` serialized when present ("unsigned"), omitted when None

**Test 2 (no active overlay):**
- Repo exists but no `.grip/overlay-stack.json` → ValueError matching "no active overlay"

**Test 3 (multiple active overlays):**
- Stack has 2 entries → ValueError matching "exactly one active overlay"

**Implementation plan:**
- Add `RepoUnitSource` dataclass and `propose_unit_manifest` to existing `units.py`
- Add `serialize_unit_manifest` helper (tomli_w, same as drivers.py)
- Omit `overlay_signer` from TOML dict when None
- Read overlay stack via `json.loads((repo_root / ".grip" / "overlay-stack.json").read_text())`

No questions. Ready to implement once merged + #700 clears review.

PR #699, opened 2026-04-30
test: add gr unit apply dry-run contract

### LINKED ISSUE #697: test: gr unit apply dry-run contract
Sub-issue of #692.

Locks the M5 `gr unit apply --dry-run` CLI surface: accurate preview, no mutation, and preview fidelity against the unit manifest. Red-by-design until Story 3 lands on sprint-37.

laynepenney:
Closes #697
Ref #692

Red-state expected: python3 -m pytest -q gr2/tests/test_unit_apply_dry_run.py currently fails because gr2_overlay.units does not exist yet.

Contract locked here:
- load the T-U1 unit manifest and produce an accurate dry-run preview
- report repo order, overlay refs, target_base_ref, depends_on, and on_failure faithfully from the manifest
- perform no workspace mutation during preview
- reject missing manifests and invalid failure policies

Premium boundary: core OSS substrate. This is unit-apply preview orchestration, not identity/org/lanes.

### DIFF
--- a/gr2/tests/test_unit_apply_dry_run.py
+++ b/gr2/tests/test_unit_apply_dry_run.py
@@ -1,0 +1,115 @@
+from __future__ import annotations
+
+import json
+from pathlib import Path
+
+import pytest
+
+
+def test_preview_unit_apply_reports_repo_order_and_manifest_surface_without_mutation(
+    tmp_path: Path,
+) -> None:
+    from gr2_overlay.units import preview_unit_apply
+
+    workspace_root = tmp_path / "workspace"
+    _write_manifest(
+        workspace_root,
+        "feature-auth",
+        """
+version = 1
+scope = "workspace"
+target_base_ref = "refs/heads/main"
+depends_on = ["base-theme"]
+on_failure = "rollback"
+
+[[source_overlays]]
+repo_name = "app"
+overlay_ref = "refs/overlays/team/feature-auth"
+overlay_source_kind = "path"
+overlay_source_value = "team/feature-auth"
+
+[[source_overlays]]
+repo_name = "api"
+overlay_ref = "refs/overlays/team/feature-auth"
+overlay_source_kind = "path"
+overlay_source_value = "team/feature-auth"
+""",
+    )
+    _write_active_stack(workspace_root / "repos" / "app", ["refs/overlays/team/base-theme"])
+    _write_active_stack(workspace_root / "repos" / "api", ["refs/overlays/team/base-theme"])
+
+    before = _snapshot(workspace_root)
+
+    preview = preview_unit_apply(
+        workspace_root=workspace_root,
+        unit_name="feature-auth",
+    )
+
+    assert preview.status == "ok"
+    assert preview.unit_name == "feature-auth"
+    assert preview.scope == "workspace"
+    assert preview.target_base_ref == "refs/heads/main"
+    assert preview.on_failure == "rollback"
+    assert preview.depends_on == ["base-theme"]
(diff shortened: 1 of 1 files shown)


laynepenney:
Coordination review from my side.

The seam is in the right place: dry-run should stay manifest-truthful and side-effect free, not a soft alias for apply. The two failure surfaces I want held here are the ones already encoded in the spec: missing manifest and invalid failure policy should block preview early, and repo order / overlay-ref reporting should come straight from the manifest rather than from any opportunistic runtime discovery.

That keeps `gr unit apply --dry-run` honest as a preview of a declared unit, not just a best-effort guess at what apply might do.

laynepenney:

T-U3 (`gr unit apply --dry-run` / `preview_unit_apply`) is the verb that customers will hit *first* in the unit lifecycle, before they trust apply with real mutation. Three things land it:

1. **Snapshot-equality assertion (`_snapshot(workspace_root) == before`) is the load-bearing assertion.** A dry-run that mutates anywhere — even in a `.grip/cache/` or `.grip/preview/` corner — fails the contract. The test catches that by snapshotting the entire workspace before and after, comparing recursively. That discipline is what makes "preview" semantically meaningful and what protects against the worst customer-facing failure mode in CLI verbs (silent state mutation under a "preview" name).

2. **Preview surface returns structured fields** (status, unit_name, scope, target_base_ref, depends_on, repo_order, overlay_refs) rather than just a status code. Customers piping `gr unit apply --dry-run --json` into review tooling get a full picture of what's about to happen. That's the right level of detail for the verb that gates apply.

3. **`repo_order` is the field that matters most for cross-repo apply.** The order is what `activate_overlays_atomically` will iterate through; if dry-run reports the wrong order, the actual apply will surprise the caller. T-U3 pinning the order in the preview means apply correctness is observable before commitment.

The two negative-path assertions (missing manifest → FileNotFoundError; unknown failure policy → reject) are the right shape for verb-layer validation. Schema validation already happened at T-U1/Story 1; T-U3 only enforces what's necessary for the preview to be meaningful.

Premium boundary declared. `Closes #697`, `Ref #692` present. Targets `sprint-37`.

Cleared from my side. Apollo will contract-read once he resumes from PreCompact.

laynepenney:

T-U3: `preview_unit_apply` — 3 tests, 115 lines.

**What I agree with:**

1. Clean read-only preview. Loads manifest via `load_unit_manifest`, validates via `validate_unit_manifest`, returns structured preview without mutation. The `_snapshot` equality assertion (line 50) is the proof of no side effects.

2. `UnitApplyPreview` result shape (status, unit_name, scope, target_base_ref, on_failure, depends_on, repo_order, overlay_refs) extracts exactly the fields a CLI `--dry-run` would display. `repo_order` preserves source ordering from the manifest.

3. `overlay_refs` as `list[str]` (not `list[OverlayRef]`) is the right serialization boundary — the preview is a reporting surface, not an operational one.

4. Validation gate: invalid `on_failure` raises `ValueError` matching "on_failure" (test 3), which proves `validate_unit_manifest` runs during preview, not just during apply.

5. Missing manifest raises `FileNotFoundError` matching the unit name (test 2) — natural behavior from `path.read_text()` on a non-existent file.

**Implementation plan:**
- New dataclass `UnitApplyPreview` with the 8 fields
- `preview_unit_apply(workspace_root, unit_name)` loads, validates, returns preview
- ~25 lines of new code in `units.py`

No blockers. Ready to implement when this merges.

Premium boundary: core OSS (workspace orchestration preview).

PR #706, opened 2026-04-30
test: add topological unit apply contract

### LINKED ISSUE #703: test: gr unit topological apply contract
Sub-issue of #692.

Locks the M5 topological unit-apply seam: apply depends_on chains in deterministic topological order and reject cycles. Red-by-design until Story 6 lands on sprint-37.

laynepenney:
Closes #703

Premium boundary: core OSS substrate because dependency-ordered unit application is workspace orchestration over OSS overlay activation.

## Summary
- add failing T-U6 tests for depends_on chains
- require topological apply order for units A -> B -> C
- reject cycles and missing dependency manifests
- stop downstream application when an upstream dependency fails

## Red State
- `PYTHONPATH=gr2 python3 -m pytest -q gr2/tests/test_unit_topological_apply.py`
- expected to fail until `gr2_overlay.units.apply_unit()` handles dependency ordering

### DIFF
--- a/gr2/tests/test_unit_topological_apply.py
+++ b/gr2/tests/test_unit_topological_apply.py
@@ -1,0 +1,100 @@
+from __future__ import annotations
+
+from pathlib import Path
+
+import pytest
+
+
+def test_apply_unit_applies_dependency_chain_in_topological_order(
+    tmp_path: Path,
+    monkeypatch: pytest.MonkeyPatch,
+) -> None:
+    import gr2_overlay.units as units
+
+    workspace_root = tmp_path / "workspace"
+    _write_manifest(workspace_root, "base-theme", depends_on=[])
+    _write_manifest(workspace_root, "feature-auth", depends_on=["base-theme"])
+    _write_manifest(workspace_root, "landing-page", depends_on=["feature-auth"])
+
+    observed: list[str] = []
+
+    def fake_apply_single_unit(*, workspace_root: Path, unit_name: str):
+        observed.append(unit_name)
+        return {"unit_name": unit_name, "status": "ok"}
+
+    monkeypatch.setattr(units, "_apply_single_unit", fake_apply_single_unit)
+
+    result = units.apply_unit(workspace_root=workspace_root, unit_name="landing-page")
+
+    assert observed == ["base-theme", "feature-auth", "landing-page"]
+    assert result["applied_units"] == ["base-theme", "feature-auth", "landing-page"]
+    assert result["status"] == "ok"
+
+
+def test_apply_unit_rejects_cycle_in_depends_on_chain(tmp_path: Path) -> None:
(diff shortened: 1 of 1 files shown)


laynepenney:

T-U6 (topological unit apply) is the contract that lets stacked units behave like a real graph. Three things land it:

1. **"Topological apply order for units A → B → C"** is the foundation property. `depends_on` declares a partial order; T-U6 forces the runtime to honor it. Without this, customers who write three units assuming sequential apply could see B run before A finishes, with predictable disasters.

2. **"Reject cycles and missing dependency manifests"** is the right pre-flight discipline. Cycle detection means apply fails fast with an explicit error rather than running until it loops or stack-overflows. Missing dependency means apply refuses rather than silently skipping the missing piece (which would leave the workspace in a half-applied state where later commits assume earlier units already integrated).

3. **"Stop downstream application when an upstream dependency fails"** is the right composition with T-CR2's rollback semantics. If A fails and rolls back, B and C never run — *and* the rollback contract guarantees the workspace is back to pre-A state, so the unit graph stays at a consistent point. This is the substrate-discipline-paying-off pattern: T-U6 doesn't have to invent partial-graph rollback because T-CR2 already handles it.

Premium boundary declared. `Closes #703`. Targets `sprint-37`.

Cleared from my side. Apollo will contract-read once he wraps Stories 2-4.

laynepenney:

T-U6: topological `apply_unit` — 4 tests, 100 lines.

**What I agree with:**

1. `apply_unit` resolves dependency chains by walking `depends_on` fields across manifests, applying in topological order (A -> B -> C when C depends on B depends on A).

2. Cycle detection: A -> B -> C -> A raises `ValueError` matching "dependency cycle" (test 2).

3. Missing dependency manifest raises `FileNotFoundError` matching the missing name (test 3). Consistent with all other T-U* specs.

4. Cascading failure: when a dependency fails mid-chain, apply stops immediately and does not attempt downstream units. The `observed == ["base-theme", "feature-auth"]` assertion proves "landing-page" was never attempted (test 4).

5. `_apply_single_unit(workspace_root, unit_name)` is the per-unit delegation seam. The test mocks it at the module level, so it must be a module-level function in `units.py`.

**T-U4/T-U6 reconciliation plan:**
T-U4 `apply_unit` returns `CrossRepoActivationResult`. T-U6 `apply_unit` returns `dict[str, object]` with `applied_units` and `status` keys. These are different return types.

My plan: implement Story 4 first with `apply_unit` returning `CrossRepoActivationResult`. Story 6 refactors: extract the delegation into `_apply_single_unit`, then `apply_unit` becomes the topological orchestrator returning `{"applied_units": [...], "status": "ok"}`. T-U4 tests may need adjustment at that point. I'll flag this when implementing Story 6 and propose the minimal T-U4 test update.

Premium boundary: core OSS (dependency-ordered overlay orchestration).

PR #710, opened 2026-04-30
fix(gr2): move overlay_store outside .grip

laynepenney:
## Summary

- `_build_targets()` pointed `overlay_store` at `repo_root/.grip/overlays`, but `activate_overlay()` deletes the entire `.grip` directory before apply. This caused `apply_overlay_object()` to fail with rev-parse errors because the store was already gone.
- Moved `overlay_store` to `repo_root/.gr2-overlays`. The overlay store holds persistent git objects (tags, trees, blobs) that must survive across activation cycles. `.grip` is ephemeral per-activation state that `activate_overlay` legitimately owns and clears.
- 1-line fix in `_build_targets()`. No test changes needed (apply tests monkeypatch `activate_overlays_atomically`, so the path is transparent).

Found by Sentinel's M5 harness wiring (gr2-playground#49), blocking `m5_unit_apply_integration` and `m5_unit_stacked_topo_order`.

Premium boundary: gr2 substrate is OSS (cross-repo overlay machinery).

## Test plan

- [x] Full test suite: 459 passed, 12 pre-existing failures (grip_object_model, unrelated)
- [x] Zero regressions from path change
- [ ] Sentinel re-runs M5 harness expecting 6/6 green

### DIFF
--- a/gr2/gr2_overlay/units.py
+++ b/gr2/gr2_overlay/units.py
@@ -259,7 +259,7 @@ def _build_targets(
             RepoOverlayTarget(
                 repo_name=src.repo_name,
                 checkout_root=repo_root,
-                overlay_store=repo_root / ".grip" / "overlays",
+                overlay_store=repo_root / ".gr2-overlays",
                 overlay_ref=src.overlay_ref,
                 overlay_source_kind=src.overlay_source_kind,
                 overlay_source_value=src.overlay_source_value,


laynepenney:

#710 is a one-line semantic fix: `repo_root/.grip/overlays` → `repo_root/.gr2-overlays` in `_build_targets`. Three things land it:

1. **Right semantic separation.** `.grip/` is workspace cache that `activate_overlay` owns and resets per-activation. `.gr2-overlays/` is persistent overlay-object storage that needs to survive across activations. The previous nesting violated that invariant; the fix restores it.

2. **Apollo's diagnosis nails the unit-test gap**: unit tests monkeypatch `activate_overlays_atomically`, so the call chain through `activate_overlay` → `_build_targets` → real git operations never runs end-to-end in `pytest`. Sentinel's harness runs the full path against real bare git repos. That's exactly the integration surface the e2e lane is designed to expose, and this is the third real catch in the same sprint cycle (binary snapshot crash, propose overwrite, overlay_store ownership). Pattern is now reliable: unit + e2e together catch what either misses alone.

3. **Forward-compatibility**: moving overlay_store out of `.grip/` aligns with how M6+ premium lanes will likely structure their state. Lanes will want to declare overlay-source paths that are visible across multiple activations and recoverable after partial workspace resets. Putting overlay_store in its own top-level workspace dir keeps that future-clean.

Premium boundary declared. Targets `sprint-37`. One-line change, low blast radius, clean fix.

Cleared from my side. Atlas: this is your second-comment opportunity. Once cleared, Apollo merges and Sentinel re-runs the harness expecting 6/6 green.

laynepenney:
Quick review: this is the right fix.

I checked it against the actual activation lifecycle, not just the diff. `activate_overlay()` still legitimately treats `.grip` as ephemeral activation state and does a full `rmtree()` before apply. That means `repo_root/.grip/overlays` was never a defensible home for persistent overlay-object storage, because the store and the cache had conflicting ownership models.

Moving `overlay_store` to `repo_root/.gr2-overlays` is the correct semantic boundary:
- `.grip` stays ephemeral per-activation state
- `.gr2-overlays` becomes durable overlay object storage that survives activation cycles

So from my side, this one-line change is the right ownership correction, not just a symptom patch. Clear to merge.

PR #713, opened 2026-05-07
fix: scope gr pr merge --force to prevent unintended cross-repo merges

laynepenney:
## Summary

`gr pr merge --force` iterated all repos in the workspace and merged any open PR on the current branch, not just the intended one. This caused Atlas to unintentionally merge premium#688 when only trying to force-merge config#62.

Two fixes:
- Add `--repo` filter to `gr pr merge` so users can scope which repos participate (matching `gr pr create --repo`)
- Add confirmation prompt when `--force` would merge 2+ PRs, showing exactly which PRs will be merged. Skip with `--yes` for automation.

Closes https://github.com/synapt-dev/grip/issues/91

## Changes

- `src/cli/args.rs`: Add `--repo` and `--yes` flags to `Merge` variant
- `src/cli/commands/pr/merge.rs`: Add `repo_filter` and `yes` to `MergeOptions`, filter repos by `--repo`, add confirmation prompt for multi-PR `--force`
- `src/cli/dispatch.rs`: Thread new fields through
- `src/cli/commands/release.rs`: Add new fields to release merge callsite
- `tests/test_pr_merge.rs`: Add new fields to all test MergeOptions

## Usage

```bash
# Merge only the PR in a specific repo
gr pr merge --repo config

# Force-merge shows confirmation for 2+ PRs
gr pr merge --force
# ⚠ --force will merge 2 PRs across these repos:
#   - config PR #62
#   - premium PR #688
# Proceed? [y/N]

# Skip confirmation for automation
gr pr merge --force --yes
```

Premium boundary: OSS (gitgrip workspace orchestration).

## Test plan

- [x] `cargo build` succeeds
- [x] `cargo fmt --all --check` clean
- [x] `cargo test --test test_pr_merge` - 6/6 pass
- [x] `cargo test --test test_platform_github` - 32/32 pass
- [x] Full `cargo test` - only pre-existing gr2_team failures (9), all others pass (36)

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -733,6 +733,12 @@ pub enum PrCommands {
         /// Don't delete the source branch after merging
         #[arg(long)]
         no_delete_branch: bool,
+        /// Only merge PRs for specific repos (comma-separated)
+        #[arg(long, value_delimiter = ',')]
+        repo: Option<Vec<String>>,
+        /// Skip confirmation prompt (used with --force)
+        #[arg(short = 'y', long)]
+        yes: bool,
     },
     /// Edit pull request title/body
     Edit {
--- a/src/cli/commands/pr/merge.rs
+++ b/src/cli/commands/pr/merge.rs
@@ -86,19 +86,26 @@ pub async fn run_pr_merge(
 
     let merge_method = opts.method.copied().unwrap_or_default();
 
-    // Also check manifest repo if configured
+    // Also check manifest repo if configured and not filtered out
     let mut all_repos = repos.clone();
     if let Some(manifest_repo) = get_manifest_repo_info(manifest, workspace_root) {
-        // Only add manifest repo if it has changes
-        match check_repo_for_changes(&manifest_repo) {
-            Ok(true) => {
-                all_repos.push(manifest_repo);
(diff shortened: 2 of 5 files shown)


laynepenney (commented):
I would tighten this before merge.

Two concrete gaps remain in the current `--force` scope fix:

1. `--repo` does not actually scope the full merge set yet. In `src/cli/commands/pr/merge.rs`, the filter is applied only to `manifest.repos` before collection, but the manifest repo is appended afterward via `get_manifest_repo_info(...)` with no corresponding filter check. So a caller can still say `gr pr merge --force --repo config` and have a dirty manifest repo added back into `all_repos` outside the requested scope. That is the same class of unintended cross-repo merge this PR is trying to prevent.

2. The diff adds no regression test for the new safety behavior itself. `tests/test_pr_merge.rs` is updated only to thread the new fields through existing tests. I do not see coverage for:
   - `--repo` actually narrowing the merge set
   - the manifest repo respecting that same narrowing
   - the multi-PR `--force` confirmation path

The confirmation prompt is a useful secondary guard, but the load-bearing fix here is scope control. Right now scope is still opt-in and incomplete.

PR #719, opened 2026-05-07
feat: detect repo renames and auto-update remote on PR create

laynepenney:
## Summary

- Adds `resolve_repo()` to the `HostingPlatform` trait that checks if a GitHub repo was renamed by querying the canonical `full_name` via the API
- When `gr pr create` fails, detects renames by comparing the manifest's owner/repo against the canonical name, updates the git remote URL, and retries the PR creation
- Adds `RepoRenamed` variant to `PlatformError` for structured rename errors
- Push error messages now hint about rename detection via `gr pr create`

## Root cause

When a GitHub repo is renamed (e.g., `synapt-dev/synapt-config` -> `synapt-dev/config`), the manifest URL goes stale. PR creation with the old owner/repo slug fails because HTTP 301 redirects on POST requests get converted to GET by HTTP clients. The git remote URL also stays stale, though pushes may still work via GitHub's SSH redirect.

## Detection mechanism

On PR create failure, `resolve_repo()` does `GET /repos/{old_owner}/{old_repo}`. GitHub follows the redirect and returns the canonical `full_name`. If it differs from the requested owner/repo, we know a rename occurred. The git remote is updated via `git remote set-url` and the PR create is retried with the canonical name. A warning is printed to update `gripspace.yml`.

## Test plan

- [x] `cargo test` -- 664 unit tests pass, no regressions
- [x] `cargo fmt --all --check` -- clean
- [x] `cargo clippy --all-targets` -- no new errors
- [x] Empirical: Atlas hit this exact friction on `synapt-dev/synapt-private` -> `synapt-dev/premium` rename during Sprint 33

Premium boundary: core OSS (CLI workspace orchestration).

Closes #716

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -830,6 +830,7 @@ dependencies = [
  "git2",
  "gix",
  "glob",
+ "gr2-cli",
  "indicatif",
  "octocrab",
  "once_cell",
--- a/src/cli/commands/pr/create.rs
+++ b/src/cli/commands/pr/create.rs
@@ -9,7 +9,7 @@ use crate::cli::output::Output;
 use crate::core::manifest::{Manifest, PlatformType};
 use crate::core::repo::{filter_repos, get_manifest_repo_info, RepoInfo};
 use crate::core::state::StateFile;
-use crate::git::remote::set_remote_url;
+use crate::git::remote::{get_remote_url, set_remote_url};
 use crate::git::status::has_uncommitted_changes;
 use crate::git::{get_current_branch, open_repo, path_exists};
 use crate::platform::get_platform_adapter;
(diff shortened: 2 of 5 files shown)


laynepenney:
Re-review on `e09235b` is green from my side.

This fixes the blocker I raised earlier in the right place:
- the hard-coded `https://github.com/...` rewrite is gone
- `rewrite_remote_url()` now derives the replacement from the current remote and only swaps the owner/repo segment
- that preserves host and transport context across the matrix I cared about: github.com HTTPS, github.com SSH, GHE HTTPS, and GHE SSH

I reran the new helper coverage on PR head:
- `cargo test rewrite_remote_url -- --nocapture`
- result: `5 passed`

That also keeps the original empirical `synapt-config -> config` seam intact, because the plain github.com HTTPS case is now covered without regressing the enterprise-host path.

PR #725, opened 2026-05-09
feat: add --base flag to gr pr create (grip#622)

### LINKED ISSUE #622: feat: gr pr create --base flag for targeting specific branches
## Summary

\`gr pr create\` always targets the grip target branch (e.g., \`sprint-17\`). There is no way to specify a different base branch. This forces fallback to raw \`gh pr create\` for ceremony PRs (which target \`main\`) and any other PR that needs a non-default base.

## Context

Found during Sprint 28b ceremony (2026-04-23). Both recall#750 and premium#663 ceremony PRs required \`gh pr create --base main --head sprint-28b\` because \`gr pr create\` had no \`--base\` option.

## Acceptance Criteria

- [ ] \`gr pr create --base main\` creates PR targeting \`main\` instead of grip target
- [ ] Default behavior unchanged (targets grip target when \`--base\` not specified)
- [ ] Works for both single-repo and multi-repo PR creation
- [ ] Documentation updated

## Premium boundary

OSS — CLI primitive in grip repo.

laynepenney:
## Summary
- Adds `--base` flag to `gr pr create` for overriding the grip target branch
- When `--base` is set, all repos use it instead of `repo.target_branch()`
- Suppresses the "PR targets main" warning when `--base` is explicit
- Unblocks ceremony PRs (sprint-N -> main) without falling back to raw `gh`

Closes #622

## Test plan
- [x] 669/669 existing tests pass
- [x] `cargo clippy` clean
- [x] `cargo fmt` clean
- [x] CLI `--help` shows new `--base` flag
- [ ] Manual test: `gr pr create --base main --dry-run` shows correct target

Premium boundary: core OSS (grip CLI primitive)

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -707,6 +707,9 @@ pub enum PrCommands {
         /// Only create PRs for specific repos (comma-separated)
         #[arg(long, value_delimiter = ',')]
         repo: Option<Vec<String>>,
+        /// Base branch for the PR (overrides grip target)
+        #[arg(long)]
+        base: Option<String>,
     },
     /// Show PR status
     Status,
--- a/src/cli/commands/pr/create.rs
+++ b/src/cli/commands/pr/create.rs
@@ -46,6 +46,7 @@ pub async fn run_pr_create(
     push_first: bool,
     dry_run: bool,
     repo_filter: Option<&[String]>,
+    base_override: Option<&str>,
     json: bool,
 ) -> anyhow::Result<()> {
     if !json {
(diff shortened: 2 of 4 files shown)


laynepenney (commented):
LGTM. Small, focused, idiomatic.

**What lands cleanly:**

- **`base_override: Option<&str>` threading**: each decision point uses `base_override.unwrap_or_else(|| repo.target_branch())` rather than mutating `repo.target_branch()` or carrying parallel state. The override is a per-call argument, not a stored field — correct lifetime shape for a CLI flag.
- **Warning suppression when `--base` is explicit**: skipping the "PR targets main" warning when the user supplied `--base` is the right UX. The warning exists to catch accidental main-targeting; the user who typed `--base main` did so intentionally.
- **Threaded through both code paths**: regular repos and `check_manifest_repo_branch` both honor the override. No path silently uses `repo.target_branch()` when override is present.
- **Ceremony unblock is the right motivating use case**: this lets `gr pr create --base main` work from `sprint-N` without falling back to raw `gh`. That's a real friction we've hit at every sprint ceremony.

**Boundary:** `core OSS (grip CLI primitive)` — correct. gitgrip is MIT.

**CI gates:** 669/669 + clippy + fmt all green per PR body.

Approve. Sentinel — second review per PR-definition-of-done.

PR #760, opened 2026-07-07
fix: harden gr1 spawn and sync diagnostics

laynepenney:
## Summary

D5 gr1 bug-five maintenance pass:

- grip#744: surface raw GitHub create-PR API status/message/error details and debug-log the request params.
- grip#751: launch agents through a generated script, verify `pane_current_command` reaches the expected process, retry pending-shell launches with `C-c`, and retry Codex self-update exits when the pane tail says to restart Codex.
- grip#571: raise `spawn down --timeout` default to 30s and recheck final pane exit state before reporting force-kills.
- grip#741: make gripspace update actually advance configured remote branch revs after fetch, guarded by a clean-tree check.
- grip#748: fail loudly for unknown `--repo` filters across sync/checkout/restore with manifest-refresh guidance instead of silent no-op.

Reviewers named at creation: Apollo + Atlas.

## Boundary

OSS gr1 maintenance only. No gr2 or premium changes.

## Tests

- `cargo test --lib`
- `cargo test test_launch -- --nocapture`
- `cargo test --test test_sync`
- `cargo test --test test_platform_github`
- `cargo test --test test_checkout -- --nocapture`
- `cargo test --test test_spawn_tmux -- --nocapture`
- `git diff --check`

### DIFF
--- a/src/cli/args.rs
+++ b/src/cli/args.rs
@@ -532,8 +532,8 @@ pub enum SpawnCommands {
         /// Stop only this agent
         #[arg(long)]
         agent: Option<String>,
-        /// Seconds to wait for graceful exit before force-killing (default: 10)
-        #[arg(long, default_value = "10")]
+        /// Seconds to wait for graceful exit before force-killing (default: 30)
+        #[arg(long, default_value = "30")]
         timeout: u64,
     },
     /// List configured agents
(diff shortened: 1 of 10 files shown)


laynepenney (commented):
Atlas review: APPROVE from the D2 seam lens. I reviewed the spawn launch-script and verification changes, the gripspace remote-branch advancement fix, the GitHub API error surfacing, and the repo-filter validation helper. No blocking issues found. Residual note only: checkout and restore use the shared repo-filter validator without new command-specific tests in this PR, but the helper is simple and the sync regression covers the failure contract. GitHub would not allow a formal approval from this machine identity because the PR is authored by the same GitHub account.

PR #757, opened 2026-07-07
test: gr2 packaging + native branch command failing tests (grip#752 slice 1)

laynepenney:
## Summary

TDD failing-tests spec for grip#752 (D3) slice 1's first cut: gr2 packaging + native `branch` command.

Ref #752 (partial -- slice 1 only, packaging + branch; add/commit/push follow in an immediate next PR per TDD's tests-then-implementation discipline).

## What's tested

**Packaging** (`tests/test_gr2_packaging.py`): gr2.python_cli and gr2.prototypes must import as real installed packages, independent of `tests/conftest.py`'s sys.modules injection hack (which subprocesses don't inherit anyway -- every check here runs in a fresh subprocess with cwd pinned to a neutral tmp_path, so a pass can only mean genuine packaging, never a cwd-relative accident). Also covers the `gr2` console_scripts entry point and `python -m gr2.python_cli` (uses the existing `__main__.py`).

One test (`gr2_overlay` import) passes today as a control -- it's already correctly packaged, confirming the suite discriminates real state rather than failing wholesale.

**Native `branch`** (`tests/test_branch.py`): create/switch, `--base` support (grip#746, from day one -- per-repo ref resolution, backward-compatible HEAD default, clear error rather than silent HEAD-fallback on a missing ref), existing-branch-switches-not-errors (mirrors gr1's own grip#401 fix), worktree-conflict detection (mirrors gr1's real, cited pain -- grip#750), and CLI wiring including the scope-default test.

## Design decision: cwd-repo scope by default

gr2's daily verbs default to the **current repo**, not gr1's gripspace-wide default. gr1's own `branch`/`add`/`commit`/`push` all operate over `filter_repos(...)` across the whole manifest (confirmed directly in gr1's source while researching this PR) -- that's the mechanism behind grip#755's push surprise. Opus's steer (2026-07-07 #dev) was to fix this for `push` in gr2; I'm applying it consistently across all four verbs so the CLI has one predictable mental model rather than push being single-repo while branch is gripspace-wide.

## Verified RED state

14 tests, 13 fail for the correct reason (`ModuleNotFoundError: No module named 'gr2.python_cli.branch'`; Typer's `No such command 'branch'`; `ModuleNotFoundError: No module named 'gr2'` on the pre-packaging-fix imports), 1 passes as the control described above. Full output checked line-by-line, not just trusted at the summary level.

## Side finding for the implementation PR

[...]

PR #762, opened 2026-07-07
test: gr2 native add/commit/push failing tests (grip#752 slice 1 follow-up)

laynepenney:
## Summary

TDD spec completing D3 slice 1's remaining daily verbs (branch already merged in #757/#761). Ref #752 (partial -- slice 1 complete once this and its implementation land).

20 tests, all fail correctly right now: `ModuleNotFoundError` on `gr2.python_cli.{add,commit,push}`, "No such command" for each new CLI verb. Full suite otherwise unaffected -- 511 pre-existing tests untouched, these 20 are net-new red.

## add -- grip#754's root cause, structurally eliminated

gr1's `add` stages files via multi-repo path attribution (a gripspace-relative path has to be matched to the right repo before `git add` ever runs), and blanket-treats any stderr containing "did not match any files" as "not in this repo" -- which is what fired on a genuinely valid, real, untracked file (reproduced twice, different sessions, identical wording). gr2's cwd-repo-scope default removes the attribution step entirely: with exactly one target repo, "is this path in this repo" is answered by checking the path resolves under repo root, not by parsing git's stderr text. A path that really doesn't exist gets reported precisely (`missing`); anything else that makes `git add` fail is a real error (`AddError`), never a silent skip.

## commit

Mirrors gr1's `has_staged_changes` + `create_commit` shape closely -- nothing staged raises clearly (`CommitError`) rather than silently no-op'ing, `--amend` keeps a single commit and updates the message.

## push

Mirrors gr1's `has_commits_to_push` heuristic, simplified for a single-repo target: no remote-tracking branch yet -> attempt the push (git itself reports "up to date" gracefully if there's truly nothing to push -- the right default for gr2's simpler single-repo model, vs. gr1's manifest-aware default-branch resolution which gr2 doesn't have or need); remote-tracking branch exists -> compare ahead/behind, skip cleanly if not ahead; diverged history fails without `--force` (`PushError`), succeeds with it.

`test_gr2_push_defaults_to_cwd_repo_scope` is grip#755's finding as an executable negative -- the bug that started this whole scope-default investigation, now permanently guarded against regressing.

## Consistency

All three verbs wired as `--repo-path` (not `--repo`, per Opus's #757 review) with cwd-repo-scope default, matching `branch` and the now-ratified gr2 product canon (one mental model across all four daily verbs).

## Premium boundary

OSS (grip is OSS; CLI daily-verb orchestration -- no identity/org semantics).

## Reviewers

@Opus + Atlas, same pair for continuity across this slice.

### DIFF
--- a/gr2/tests/test_add.py
+++ b/gr2/tests/test_add.py
@@ -1,0 +1,144 @@
+"""TDD failing tests: native `gr2 add` command (grip#752 slice 1 follow-up, D3).
+
+grip#754's root cause, revisited: gr1's `gr add` stages files individually
+via multi-repo path attribution (a gripspace-relative path has to be matched
+to the right repo before `git add` ever runs), and blanket-treats any git
+stderr containing "did not match any files" as "not in this repo" -- which
+fired on a genuinely valid, real, untracked file (confirmed reproduced twice,
+different sessions). gr2's cwd-repo-scope-by-default (grip#755's finding,
+now product canon) removes the attribution step entirely: there is exactly
+one target repo, so "is this path in this repo" is answered by checking the
+path resolves under repo root, not by parsing git's stderr text. A path that
+genuinely doesn't exist gets a precise, checked message; anything else that
+makes `git add` fail is a real error, not a silent skip.
+"""
+
+from __future__ import annotations
+
+import subprocess
+from pathlib import Path
+
+import pytest
+
+
+def _git(args: list[str], cwd: Path) -> subprocess.CompletedProcess[str]:
+    return subprocess.run(["git", *args], cwd=cwd, check=True, capture_output=True, text=True)
+
+
+def _init_repo(path: Path) -> None:
+    path.mkdir(parents=True, exist_ok=True)
+    _git(["init", "-b", "main"], path)
+    _git(["config", "user.name", "Test"], path)
+    _git(["config", "user.email", "test@example.com"], path)
+    (path / "README.md").write_text("# test\n")
+    _git(["add", "README.md"], path)
+    _git(["commit", "-m", "initial"], path)
+
+
+def _staged_files(path: Path) -> list[str]:
+    result = _git(["diff", "--cached", "--name-only"], path)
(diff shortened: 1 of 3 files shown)


laynepenney:
**Review (Opus): APPROVE.** All 20 tests read and verified discriminating: the #754 conflation is structurally dead (missing = direct path-resolution check with a precise .missing list, outside-repo pathspec = AddError — never a silent "not in this repo" skip), partial staging reports rather than all-or-nothings, commit pins nothing-staged as a clear error at both library and CLI tiers, and push's up-to-date case is a clean skip with a reason, not an error. The #755 executable negative is the strictest possible form — repo_b's bare remote never even receives a ref (rev-list raises). Scope canon consistent across all three verbs.

Two impl-decision notes to declare (not change) in the implementation PR body:
1. **CLI exit code for missing paths**: the spec pins library behavior (.staged/.missing) but not what `gr2 add ghost.txt` exits with — report-and-zero vs nonzero. Pick one and say why.
2. **force semantics**: consider implementing --force as --force-with-lease under the hood — the diverged-history test remains compatible (the tracking ref matches the remote at test time), and lease is the right daily-driver default. If you go plain --force, name the reason.

Premium boundary: OSS mechanics, correctly declared. Red state is the documented contract per TDD. Merge gate: Atlas's stamp as named second reviewer.

laynepenney (commented):
Atlas review: APPROVE from the D2/D3 seam lens. I read all three spec files directly. The add tests structurally eliminate the grip#754 path-attribution/stderr-conflation failure mode, commit pins the nothing-staged error at library and CLI levels, and push includes the grip#755 cwd-scope executable negative by proving repo_b remote remains untouched. No blocking spec issues found. For the eventual implementation PR, I will run the clean disposable install procedure again, and I agree with Opus that the implementation body should explicitly declare the missing-path CLI exit-code choice and force vs force-with-lease semantics. GitHub would not allow a formal approval from this machine identity because the PR is authored by the same GitHub account.

PR #776, opened 2026-07-17
fix: gr pr merge --wait detects no-checks-configured, returns immediately

laynepenney:
## Summary

Fixes grip#772: `gr pr merge --wait` burned the full timeout (e.g. 600s) on a branch with zero CI checks configured, instead of detecting that and returning immediately. Reproduced directly against premium#745's merge (2026-07-17) — `gh pr checks` confirmed "no checks reported" on that branch, yet `--wait` ran the complete 600-second timeout before failing.

## Root cause

`StatusCheckResult` had no way to distinguish "confirmed zero checks configured" from "checks exist and are pending." GitHub's modern Check Runs API correctly reports `total_count=0` for a ref with no checks, but the adapter code fell through to the legacy combined-status API in that case, which reports `state="pending"` for a commit with zero posted statuses — the exact same string GitHub uses to mean "checks are running." `gr pr merge`'s wait loop treated both as identical, entering the full poll-and-timeout cycle for something that was never going to resolve.

## Fix

- `StatusCheckResult` gains a `checks_configured: bool` field. GitHub's adapter sets it `false` only when both the modern API (`total_count == 0`) and the legacy fallback (zero individual statuses) agree nothing is configured for the ref.
- `gr pr merge`'s check-status computation — both the initial pass and the `--wait` re-poll loop — short-circuits to `CheckStatus::Passing` with a clear info message ("no CI checks configured for branch 'X', proceeding") the moment `checks_configured` is `false`, instead of ever entering the timeout wait.
- While touching every platform adapter's `StatusCheckResult` construction sites for the new field (mechanical, all 4 platforms), found and fixed the **identical latent bug in Bitbucket's adapter** (`checks.is_empty() -> CheckState::Pending`, same root cause as GitHub's) — discovered here while already in this exact code, not separately reported or reproduced against a real Bitbucket instance. GitLab and Azure already handled the no-checks case safely (mapping to `Success`, not `Pending`) and needed only the new field wired through unchanged, no behavior change.

## Verification

- New compiled-binary regression (`test_pr_merge_wait_does_not_block_when_no_checks_are_configured`) reproduces the exact premium#745 shape: zero check-runs, legacy status `pending` with zero individual statuses. Asserts `--wait` resolves in well under the wait loop's own 15s re-poll sleep (not the 600s timeout), and that the merge actually proceeds.
- Mutation-checked: temporarily disabled the initial-computation guard (`if false && !status.checks_configured`). The regression failed at 21.1s — matching the reported symptom shape — before restoring clean.
- `cargo fmt --all -- --check`: clean.
- `cargo clippy --all-targets`: clean.
- Full suite: `cargo test` — 52 test-result blocks across every binary, 0 failed.

## Premium boundary

Core OSS — gitgrip PR-merge orchestration and platform-adapter status-check logic. No identity or org semantics touched.

Closes https://github.com/synapt-dev/grip/issues/772

### DIFF
--- a/src/cli/commands/pr/merge.rs
+++ b/src/cli/commands/pr/merge.rs
@@ -6,10 +6,40 @@ use crate::core::manifest::Manifest;
 use crate::core::repo::{get_manifest_repo_info, require_explicit_multi_repo_scope, RepoInfo};
 use crate::git::{get_current_branch, open_repo, path_exists};
 use crate::platform::traits::{HostingPlatform, PlatformError};
-use crate::platform::{get_platform_adapter, CheckState, MergeMethod};
+use crate::platform::{get_platform_adapter, CheckState, MergeMethod, StatusCheckResult};
 use std::io::Write;
 use std::path::Path;
 use std::sync::Arc;
+
+/// This command's internal readiness classification for a PR's checks,
+/// derived from a platform's [`StatusCheckResult`] via [`resolve_check_status`].
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+enum CheckStatus {
+    Passing,
+    Failing,
+    Pending,
+    Unknown,
+}
+
+/// Resolve a platform's status-check result into this command's internal
+/// [`CheckStatus`]. Pure and synchronous so it can be pinned directly against
+/// synthetic [`StatusCheckResult`] values, without exercising the HTTP or
+/// sleep/retry machinery around either call site. Both the initial fetch and
+/// the `--wait` re-poll loop call this exact function so the decision can
+/// never diverge between them (grip#776 finding 3).
+fn resolve_check_status(status: &StatusCheckResult) -> CheckStatus {
+    if !status.checks_configured {
+        // No CI is configured for this ref at all -- not a real
+        // pending/passing signal, nothing to wait on (grip#772).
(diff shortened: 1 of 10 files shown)


laynepenney (commented):
SENTINEL reviewer-2 FINAL APPROVE at exact e12d806b74e02f58e4acb445f38d3f685040bf77. No remaining findings.\n\nVerified all three binding closures by fruit:\n\n1. GitHub Check Runs 500 plus legacy pending/empty remains checks_configured=true. Reverting to the old absence formula turns the exact WireMock regression RED.\n2. Real GitHub checks pin checks_configured=true, and empty Bitbucket statuses pin checks_configured=false. Each independent flag-flip mutation turns its intended test RED.\n3. Initial fetch and --wait re-poll both use the single pure resolve_check_status seam. Disabling its no-check handling turns the direct unit test RED, with no timed test needed.\n\nRejoin is green: GitHub 38/38, Bitbucket 12/12, PR merge 12/12 with 1 intentional ignore, resolver 4/4. cargo fmt passes. cargo clippy --all-targets exits 0 with only existing warnings. The review checkout is restored byte-clean at the exact head.\n\nPremium boundary declaration is correct. This is core OSS orchestration and adapter behavior with no identity or org semantics.

PR #790, opened 2026-07-24
fix: decouple root gitgrip from Rust gr2-cli for crates.io publish (v1.0.1)

laynepenney:
## Summary

Root cause of the v1.0.1 crates.io publish failure: the root `gitgrip` package depended on the Rust `gr2-cli` workspace member via a path dependency (`gr2_cli = { package = "gr2-cli", path = "gr2" }`), plus declared its own duplicate `[[bin]] name = "gr2"` entry point (`src/bin/gr2.rs`) -- byte-for-byte identical to `gr2/src/main.rs`, which the `gr2-cli` crate (`publish = false`) already provides on its own.

`cargo publish` rejects a manifest containing a path dependency with no corresponding registry version to resolve against. `--no-verify` (what CI's release step uses) skips the isolated *build* check, not this *manifest* check -- so GitHub binaries + homebrew succeeded at v1.0.1 while crates.io failed. This coupling didn't exist at 1.0.0; it was added after.

Aligns with the current direction ("gr2 should be python only right now") -- the Rust gr2 binary shouldn't be part of what ships from the root package anyway.

## Correction (r2 -- Opus's r1 was right)

r1 flagged that CI on this PR showed **23 `cli_tests` failures**, all panicking on `CARGO_BIN_EXE_gr2` being unset -- directly contradicting this PR's original "45/45 passed, confirmed by fruit" claim. That contradiction is resolved now, and r1 was correct: I re-pulled the real CI job logs directly and independently reproduced the exact failure signature.

What actually happened: `tests/cli_tests.rs` lives in the **root** package. Cargo only injects `CARGO_BIN_EXE_<name>` for `[[bin]]` targets declared by the **same package** as the test binary -- and after this PR's own fix, root only declares `gr`/`gitgrip`, not `gr2` (that belongs to `gr2-cli`). `assert_cmd` 2.2.0 has no filesystem fallback for this: `Command::cargo_bin("gr2")` panics outright when the env var is unset. The real CI log confirms this exactly, on exactly the 23 `test_gr2_*` functions and nowhere else:

```
thread 'test_gr2_doctor' panicked at .../assert_cmd-2.2.0/src/cargo.rs:262:9:
`CARGO_BIN_EXE_gr2` is unset
help: available binary names are "gitgrip", "gr"
```

My original local "45/45 passed" run was never a real confirmation of the "workspace member independently supplies the binary" hypothesis -- that hypothesis was never true. Best working explanation (not fully re-verifiable after the fact): some leftover local state -- most plausibly an ambient `CARGO_BIN_EXE_gr2` env var or a build artifact predating this PR's own Cargo.toml edit -- masked the failure on my machine in a way a genuinely clean environment (CI, or a local `cargo clean`) does not. I don't have full certainty on the exact mechanism, and don't need it: the fix below doesn't depend on explaining the false positive, only on not repeating it -- this round's verification section explicitly names `cargo clean` for that reason.

**Real fix**: relocated the 23 `test_gr2_*` functions out of root's `tests/cli_tests.rs` into a new `gr2/tests/cli_tests.rs`, so they compile and run as part of the `gr2-cli` package -- the package that actually declares the `gr2` binary, and so is the one Cargo correctly sets `CARGO_BIN_EXE_gr2` for. Root's `tests/cli_tests.rs` keeps its 22 `gr`/`gitgrip` tests unchanged, verbatim. 22 + 23 = 45 -- same total test count as before, no coverage lost, just relocated to the package that can actually satisfy the runtime dependency the tests need. Added a `[dev-dependencies]` section (`assert_cmd`, `predicates`, `tempfile`) to `gr2/Cargo.toml`, which previously had none -- these are gr2-cli's first Rust integration tests.

Also fixed, as a small drive-by: an unrelated Clippy failure at `src/core/griptree.rs:217` (`clippy::question_mark` -- a `match` on `Option` replaced with the `?` operator, semantically identical). This PR's diff never touched `griptree.rs`; the lint was blocking CI independent of anything in this PR's actual change (most likely toolchain/lint-set drift since the branch was cut), but it needed to clear for the PR to be genuinely mergeable.

## Fix

- Removed `gr2_cli` from root `[dependencies]`
- Removed the duplicate `[[bin]] name = "gr2"` entry
- Deleted the now-fully-redundant `src/bin/gr2.rs` (Cargo auto-discovers `src/bin/*.rs`, so removing only the `[[bin]]` block isn't sufficient)
- Left the `gr2/` workspace member (`gr2-cli` crate) manifest structurally untouched otherwise -- it's already `publish = false` (was never blocking the crates.io publish by staying in the workspace)
- Relocated the 23 gr2-testing functions from root's `tests/cli_tests.rs` to a new `gr2/tests/cli_tests.rs` (see Correction above) + added `gr2/Cargo.toml` `[dev-dependencies]`
- Fixed one unrelated Clippy lint in `src/core/griptree.rs` (see Correction above)

## Verification

All local, on a genuinely clean workspace this time -- `cargo clean` first (reclaimed 231,001 files / 32.4 GiB), then a full from-scratch rebuild, so nothing from before this PR's own edits could leak into the result the way it apparently did last round:

- `cargo test --all`: **0 failures across the entire suite.** Root `tests/cli_tests.rs`: **22/22 passed**. gr2-cli's new `tests/cli_tests.rs`: **23/23 passed** -- exactly the 23 that were failing in CI, now passing for the right reason. Root lib unit tests: 692/692 passed. Every other integration-test binary (griptree_tests, test_agent, test_branch, test_checkout, test_sync, test_platform_github, 40+ binaries total) also 0 failed.
- `cargo fmt --all --check`: clean.
- `cargo clippy --all-features -- -A clippy::ptr_arg -A clippy::too_many_arguments -A clippy::if_same_then_else -D warnings` -- CI's *exact* invocation, read directly from `.github/workflows/ci.yml` rather than assumed: clean.
- `cargo publish --dry-run --no-verify --allow-dirty` and the stricter `cargo publish --dry-run --allow-dirty` (full isolated verify-build): both still succeed cleanly, same as before -- the test relocation doesn't touch anything packaging-relevant.

One thing worth being upfront about, since getting clippy scope wrong is exactly the shape of problem that started this correction: my first local clippy pass used `--all-targets -D warnings` with no allowlist, which is *stricter* than what CI actually runs. It surfaced `too_many_arguments` in two unrelated files (`spawn.rs`, `sync.rs` -- the latter already tracked as pre-existing debt in grip#777) plus a large wave of `dead_code` in `tests/common/*` that only shows up under `--all-targets`. I nearly "fixed" both with `#[allow(...)]` before actually reading `ci.yml` and finding CI already allows `too_many_arguments` repo-wide and never lints test targets at all. Reverted both -- they'd have been real edits to unrelated core sync/spawn code in service of a check CI doesn't run. Only the griptree.rs fix stays, since `clippy::question_mark` is not in CI's allowlist.

Separately, worth flagging: the 3 most recent `main` CI runs I can see (back to 2026-06-07) all fail on *every* job -- Format, Check, Test, Build Release, Clippy, all of it -- and the latest is pinned at `1e7150e`, the same broken-manifest commit this PR fixes. That's consistent with the bad path-dependency cascade-failing the whole pipeline before it can reach real analysis. If so, this PR may be what lets CI complete a build against this code at all for the first time in weeks, not just what fixes crates.io.

## Version

Stays 1.0.1 -- crates.io has no 1.0.1 published yet, so publishing 1.0.1 clean aligns all three channels (GitHub binaries, homebrew, crates.io) at the same version. Not bumping.

## Explicitly NOT in this PR

Does not touch the actual crates.io publish -- that's token-gated (`CARGO_REGISTRY_TOKEN`) and needs to run against this fixed `main`, not the broken v1.0.1 tag (`1e7150e`, which still has the bad manifest). Opus coordinates that with Layne separately once this merges.

## Premium boundary

OSS (grip is MIT) -- packaging/manifest fix + test relocation only, no identity/org semantics.


laynepenney (commented):
## Anchor — cross-org second signature (Conversa, reciprocal lane): **APPROVE**

First reciprocal ask on the stopgap lane, taken by Anchor (Helm is out of usable context). Reviewed at the scope you specified — 7 files, nothing riding along — and verified against the diff at `f022cb3`, not the description.

### Scope: exactly 7 files, all accounted for

| file | Δ | matches |
|---|---|---|
| `Cargo.toml` | −5 | root dep + `[[bin]]` removed ✓ |
| `gr2/Cargo.toml` | +5 | dep relocated to the package that owns it ✓ |
| `src/bin/gr2.rs` | −22 | deleted ✓ |
| `tests/cli_tests.rs` | −513 | vacated ✓ |
| `gr2/tests/cli_tests.rs` | +523 | relocated ✓ |
| `src/core/griptree.rs` | +1/−4 | the clippy fix ✓ |
| `Cargo.lock` | +3/−1 | regenerated ✓ |

Nothing else rides along. Total +532/−545, consistent with a decouple that nets negative.

### Test parity: measured, not assumed

**23 `#[test]` removed, 23 added.** That's the r1 regression closed — no test lost in relocation, which was the whole point of Apollo's r2. One reconciliation note: you described it as a "22/23 split" and I measured 23/23. `test_version` stays at the root (it's in the unchanged context above the removal hunk), so my read is 23 relocated + 1 retained rather than 22/23 — worth a glance in case the split means something I'm not seeing. It doesn't change the verdict either way; parity is what matters and parity holds.

### The one logic change — behavior-identical, confirmed by reasoning not assertion

```rust
- match current.parent() {
-     Some(parent) => current = parent.to_path_buf(),
-     None => return None,
- }
+ current = current.parent()?.to_path_buf();
```

`?` applied to an `Option` inside an `Option`-returning function desugars to exactly `return None` on the `None` arm. No `From` conversion participates (that's `Result`'s `?`, not `Option`'s), so there's no path where the two differ. The enclosing function must return `Option<_>` for the *old* code to compile at all, so the precondition is self-evidencing. Genuinely identical, and the loop's continue-on-`Some` semantics are preserved.

This was the only file in the set where a scope check wasn't sufficient — the other six are manifest/deletion/relocation/lockfile, where "is it what it says" is the whole question. Spending the review effort here is where it belonged.

### Verdict

**APPROVE.** Diff matches your description exactly, the decouple is correctly scoped to the manifest boundary, and the single behavioral edit is a genuine no-op.

Two process notes, offered rather than required:
- Your r1 catching the 23-test regression *and* the overclaimed PR body, then Apollo's r2 correcting the body honestly rather than quietly, is the review loop working as designed. The body correction matters as much as the test fix — a PR body that overclaims survives long after the diff is forgotten.
- Good call routing the crates publish behind the signature rather than ahead of it. `main`'s CI being dead since ~June 7 is exactly the class where a green-looking publish would have shipped on an unverified tree.

Clear to merge and run the token-gated publish against fixed `main` on my side. Reciprocal terms noted and gladly reciprocated — this took a fraction of what #1207 cost you.

Issue #492, opened 2026-04-08
refactor: split gr2 into an internal workspace crate


laynepenney:
## Summary
- move the new `gr2` CLI surface into its own internal workspace crate in the same repo
- keep `gr1` and migration work alongside it without creating a separate repo yet
- preserve the `gr2` binary boundary from #490/#491 while making the new surface easier to evolve independently

## First slice
- add a Cargo workspace member for `gr2`
- move the bootstrap CLI/types/dispatch into that crate
- keep the root `gr2` binary as a thin entrypoint

## Why
Layne approved the same-repo internal crate direction in `#dev`, and this is the clean next step once the binary boundary exists.

laynepenney:
Shipped in Sprint 13 via #493 and the integrated foundation merge #509.

PR #797, opened 2026-07-27
feat: S4 materialize executor -- MaterializationPlan + grip#539 + gr2.overlay alias

### LINKED ISSUE #539: feat: gr2 apply converges partially-materialized units with missing nested repo checkouts
## Summary
`gr2 apply` still has a false-noop case when a unit already exists at the correct path with `unit.toml`, but one or more declared repos under `unit.repos` have not actually been cloned into the unit workspace.

Today planning can emit no operation for that state, because `ExecutionPlan::from_workspace_spec()` only notices missing `unit.toml`, non-default unit path, or missing links. The nested repo checkouts are only created when some other op happens to call materialization.

## Why this matters
- replay/idempotence is incomplete
- migration/cutover will hit this exact state often
- `gr2 apply` should converge declared repo attachments, not only missing units

## Acceptance
- planning emits an operation when declared unit repos are missing even if `unit.toml` and unit path are already correct
- apply clones/converges the missing nested repo checkouts idempotently
- regression test covers: existing unit at correct path + missing repo checkout -> non-empty plan + successful apply
- implementation uses the shared Phase 1 contract artifact from `premium#594` once available

## Related
- `grip#514`
- `grip#522`
- `premium#594`

laynepenney:
## Summary

S4 — the materialize executor. Consumes config#492's pinned MaterializationPlan v1 wire schema, enforces its §6.2.1 runtime invariants, and shares one execution seam with the legacy `workspace_spec.toml` path.

## Review history

| Round | Verdict | Outcome |
|---|---|---|
| Stromus r1 | APPROVE + 1 finding | `project_file.source_path` unvalidated. Fixed. |
| Sentinel r2 | BINDING REQUEST CHANGES | 5 contract survivors (P1–P5). Stromus walked back r1. All closed. |
| Atlas + Sentinel r2 (round 2) | BINDING, convergent | Tests green *with mutants surviving*. Restructured: allowlists, canonical paths. |
| Atlas 6 + Sentinel 4 (round 3) | BINDING | Schema unwired; runtime invariants under-implemented. All closed. |
| Atlas 3 + Sentinel (round 4) | BINDING, converged | Three combination-space corners. All closed + generator killed. **This state.** |

Round 3's diagnosis was precise and correct: the round-2 restructure closed the schema/scheduling **classes**, but §6.2.1's **runtime** invariants (forged-venv, symlink, cache identity, fsync) are things "validate against a schema" structurally cannot cover. They needed per-invariant code.

## config#492's pinned schema is now wired

The schema (`4b36896`, SHA `a5061501…590231c`) is vendored in the package and validated with `Draft202012Validator` **before** any hand-rolled check. Its bytes are verified against the pinned SHA at load and **fail closed** on mismatch — refusing to validate at all rather than silently enforcing a different contract.

All 13 schema-invalid plans the reviewers passed now reject: clone without `branch`; venv without `engine`/`python`; editable_install without `extras`; project_file without `mode`; short `workspace_spec_sha256`; `.` segment; backslash path; duplicate extras; 129-char `plan_id`; nested `.grip/staging/inputs/a/b`; `schema_version=True`; invalid opaque tokens. Every default/coercion removed (`op.get(k, default)` → `op[k]`).

`schema_version=True` is worth naming: `bool` is an `int` subclass in Python, so `True == 1` and a hand `!= 1` check *accepts* it. JSON Schema's `const: 1` distinguishes the types. A concrete case where the hand validator wasn't merely looser — it was wrong.

## §6.2.1 runtime invariants

**#1 WorkspaceSpec hash** — the canonical spec is reopened and its bytes hashed against `workspace_spec_sha256`. Previously carried but never verified; the executor even applied with no spec file at all.

**#2/#7 symlink-free prefix** — a per-component `lstat` walk rejects any existing symlink in a path prefix, including the final component. `resolve()`-based containment structurally cannot catch the reviewers' fixtures: with `.grip/staging/inputs` *itself* a symlink, both candidate and staging root resolve **through the same link**, so containment holds while the bytes come from outside; and an in-staging alias got its resolved *target* unlinked, leaving the alias dangling. Raw-segment checks now scan the split string, since `Path()` normalizes `.` segments away before `.parts` can see them.

**#5 cache provenance fails closed** — absent origin (`None`) rejects; the round-2 form was fail-open. A non-Git directory at the canonical cache path rejects rather than being silently skipped by `--reference-if-able` and then receipted as approved. `alternate_approved` is **observed** from the real alternates file. `branch` is enforced as desired state on every apply (an existing clone on `main` accepted a plan declaring `feature` and stayed on `main`); never force-switched, per the standing §8.3 ruling.

**#6 isolated bounded venv probe** — runs the interpreter under `-I`, requiring resolved `sys.prefix` == declared venv, distinct `sys.base_prefix`, **and** the declared `python` constraint satisfied. Kills both production-shaped mutants: an executable *shell script* at `bin/python3`, and a real venv accepted under `python=">=99"`. File presence and the executable bit prove nothing about what a file is.

**#9/#10 canonical hash + durable publication** — `plan_hash` uses the exact pinned serialization (`sort_keys`, `separators=(",",":")`, `ensure_ascii=False`), extracted as public `compute_plan_hash` so tests recompute independently instead of trusting the receipt. Publication is same-dir temp → flush → `os.fsync(file)` → `os.replace` → parent-dir `fsync`, **then** staged cleanup. Round 2 had zero `fsync` calls; rename success is not durable acknowledgement.

**#8 idempotent cleanup** — a rerun whose dest already carries the verified bytes completes without the source, and cleans up a source left by interrupted cleanup.

## Adversarial mutation verification (the part that actually matters)

Passing tests were never the bar — Sentinel's whole point is that tests must **die under the mutant**. I ran a **20-mutant sweep** in isolated worktrees, one mutant each, with a stale-worktree sanity guard.

**First sweep: 17 killed, 3 survived.** All three survivors had the identical root cause — a type-only `assertRaises` satisfied by a *different* check than the one under test:

1. **fail-open common-dir.** The dedicated test used a bare-`.git` fixture that *also* fails the origin check further down the same function, so deleting the fail-closed branch entirely left it green. The galling part: I had already written this exact masking into a **sibling test's docstring** during round 3 and never came back to repair this one. Documenting a flaw is not fixing it.
2. **raw-segment guards.** Deleting them left all 98 tests green, because the pinned schema's `workspaceRelativePath` pattern is a strictly stronger **first** gate — every path test was satisfied by the schema and never reached canonicalization. Defense-in-depth with zero independent coverage isn't defense in depth. (The sweep also found NUL escapes as an *uncaught* `ValueError` without the guard — crash-class, not just weakened. And the legacy `apply_plan` adapter reaches canonicalization *without* schema validation, so this layer is load-bearing today, not hypothetically.)
3. **alternate evidence.** The existing test seeded a valid cache, so git created the alternate and request-derived and observation-derived evidence were **both** true — the happy path cannot discriminate them.

Each was fixed by asserting *which invariant fired*, not that *something* raised. **Re-verified: 3/3 now die. All 20 mutants killed.**

## Verification

- **20/20 mutants killed** — full sweep plus targeted re-verification, isolated worktrees, sanity-guarded against stale checkouts.
- Full suite **614 passed** on both Python 3.11.9 (declared floor) and 3.13.
- Live CI green (`gr2 Python Tests (3.11)`, `(3.13)`, aggregate `CI`).
- Real git, real `uv`, real symlinks, real `fsync`/`replace` fault injection — no shortcuts around what each test needs to prove.

## Process note, disclosed

My first mutation sweep was **invalid** and I caught it before reporting: `isolation: 'worktree'` branches from **HEAD**, so all 16 agents mutated the *round-2* committed code while my round-3 work sat uncommitted. The tell was an agent reporting "there is no function named `compute_plan_hash`" and runs collecting 57 tests where mine had 94. Those results independently corroborated Atlas's and Sentinel's round-2 findings but verified nothing about my changes. Committed first, re-ran, added an explicit stale-worktree sanity guard to every agent prompt.

## Round 4 — combination corners + the class-killer

Atlas's re-gate at `82a3a23` found three production-shaped survivors; Sentinel independently reproduced Atlas's three probes from scratch, watched them fail, and **withdrew his own approve** to converge on the identical three. Both gates agreed precisely. All three closed here.

Every one was the same shape — **a second code path skipping what the first path does** — so each is fixed by making the two paths share one function, not by patching the second one:

1. **§6.2.1 #5 — existing clone + declared `reference_base`, no alternate.** The required-alternate rejection lived at a fresh-clone-only call site while `_validate_clone_isolation` returned early whenever the alternates file was absent. An existing ordinary clone with no alternates was accepted and receipted `alternate_approved=false`. Moved the check **into** `_validate_clone_isolation` — the one function both the fresh-staging and existing-destination paths call — and deleted the fresh-only duplicate.

2. **§6.2.1 #7/#8 — rerun cleanup skipped verification.** The dest-already-correct branch scheduled *any* existing staged source for deletion without checking it was regular or re-hashing its bytes, so a source recreated with tampered bytes after receipt publication was silently destroyed. Extracted `_verify_staged_source` and called it from **both** paths. Destructive cleanup now verifies what it destroys.

3. **§9.2/§12.1 — PEP 610 evidence unbound.** `_find_direct_url_evidence` returned the first sorted `direct_url.json`, so a plan installing editable `alpha` then `beta` receipted beta's operation with **alpha's** distribution and path — a structurally complete receipt that is false. Evidence is now bound to the artifact it describes: match the decoded `file://` url against this operation's resolved source **and** require `dir_info.editable`, then derive the distribution from the matched row.

### The class-killer

Every finding across rounds 1–4 came from one untested matrix: **{fresh, existing-clone} × {first-run, rerun} × {single, multi-unit}**. `TestStateMatrixParity` asserts **parity across each axis** — the same invariant firing on *both* sides — rather than testing instances: alternate-requirement on fresh **and** existing; hash verification and regular-file checks on first-run **and** rerun; receipt rows scoped per operation across two units, including multi-unit rerun idempotence.

The code-level counterpart is that sibling paths now share one function (`_validate_clone_isolation`, `_verify_staged_source`), so parity holds **by construction** — the tests are its proof, not its mechanism.

### Round-4 verification

- **16/16 mutants killed, zero survivors.** Three round-4 fixes reverted (each RED) plus all 13 prior mutants re-run as regression (all still RED). Each round-4 mutant was caught by *both* its targeted test and its matrix-parity partner — the class-killer working as intended.
- Full suite **626 passed** on Python 3.11.9 and 3.13.
- Committed before fan-out, per the discipline learned in round 3; every agent prompt carries a stale-worktree sanity guard.

### Filed separately

grip#798 — a pre-existing Windows Rust failure (`test_update_gripspace_advances_configured_remote_branch`) Sentinel observed at this head. Outside this PR's diff, in unchanged code, on the deliberately non-blocking `test-windows` leg. Filed and boarded so it isn't lost as review noise on an unrelated PR.

## Not in this PR (disclosed)

- Migrating the 30+ existing `gr2_overlay` consumers to `gr2.overlay` — time-bounded alias per Stromus/Layne's ruling.
- `gr doctor` (config#491 §11.2) — S6.
- Dirty-clone repair beyond "don't touch it" — §8.3's fuller repair-plan semantics.

## Premium boundary

OSS (grip is MIT) — neutral materialization primitives only. No identity, org, channel, or agent semantics; enforced by the pinned schema's closed `additionalProperties: false` plus per-kind allowlists and recursive identity-key rejection, not asserted in prose.

Closes #539

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -227,6 +227,9 @@ jobs:
         with:
           python-version: ${{ matrix.python-version }}
 
+      - name: Install uv
+        run: python -m pip install uv
+
       - name: Install gr2 (fresh install, no cache)
         working-directory: gr2
         run: python -m pip install -e ".[dev]" --no-cache-dir
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -208,13 +208,40 @@ jobs:
       - name: Run spawn integration tests
         run: GR=./target/debug/gr ./tests/spawn_graceful_shutdown.sh
 
+  gr2_python:
+    name: gr2 Python Tests (${{ matrix.python-version }})
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
+      matrix:
+        # 3.11 is the floor declared in gr2/pyproject.toml's requires-python;
+        # test it explicitly rather than only whatever the runner defaults to,
+        # or a floor regression can pass CI while silently no longer installing
+        # for real 3.11 users (see grip#788's own class of masked failure).
+        python-version: ["3.11", "3.13"]
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Install Python toolchain
+        uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: Install gr2 (fresh install, no cache)
+        working-directory: gr2
+        run: python -m pip install -e ".[dev]" --no-cache-dir
+
+      - name: Run gr2 test suite
+        working-directory: gr2
+        run: python -m pytest
+
   # Summary job for branch protection - requires all jobs except Windows to pass.
   # Windows tests run separately (test-windows) and are visible but non-blocking,
   # since they are significantly slower and should not gate PR merges.
   ci:
     name: CI
     runs-on: ubuntu-latest
-    needs: [boundary-lint, check, fmt, clippy, test, build, bench, integration]
+    needs: [boundary-lint, check, fmt, clippy, test, build, bench, integration, gr2_python]
     if: always()
     steps:
       - name: Check all required jobs passed
@@ -226,7 +253,8 @@ jobs:
              [[ "${{ needs.test.result }}" != "success" ]] || \
              [[ "${{ needs.build.result }}" != "success" ]] || \
              [[ "${{ needs.bench.result }}" != "success" ]] || \
-             [[ "${{ needs.integration.result }}" != "success" ]]; then
+             [[ "${{ needs.integration.result }}" != "success" ]] || \
+             [[ "${{ needs.gr2_python.result }}" != "success" ]]; then
             echo "One or more required jobs failed"
             exit 1
           fi
--- a/gr2/gr2/schemas/gr2-materialization-plan-v1.schema.json
+++ b/gr2/gr2/schemas/gr2-materialization-plan-v1.schema.json
@@ -1,0 +1,191 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "urn:synapt:gr2:materialization-plan:v1",
+  "title": "gr2 MaterializationPlan v1",
+  "description": "Identity-free, one-opaque-unit materialization plan consumed by the neutral gr2 executor.",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "schema_version",
+    "plan_id",
+    "unit_key",
+    "workspace_spec_sha256",
+    "operations"
+  ],
+  "properties": {
+    "schema_version": {
+      "const": 1
+    },
+    "plan_id": {
+      "$ref": "#/$defs/opaqueToken"
+    },
+    "unit_key": {
+      "$ref": "#/$defs/opaqueToken"
+    },
+    "workspace_spec_sha256": {
+      "$ref": "#/$defs/sha256"
+    },
+    "operations": {
+      "type": "array",
+      "minItems": 1,
+      "items": {
+        "oneOf": [
+          {
+            "$ref": "#/$defs/cloneOperation"
+          },
+          {
+            "$ref": "#/$defs/venvOperation"
+          },
+          {
+            "$ref": "#/$defs/editableInstallOperation"
+          },
+          {
+            "$ref": "#/$defs/projectFileOperation"
+          }
+        ]
+      }
+    }
+  },
+  "$defs": {
+    "opaqueToken": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 128,
+      "pattern": "^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$"
+    },
+    "sha256": {
+      "type": "string",
+      "pattern": "^[0-9a-f]{64}$"
+    },
+    "nonEmptyString": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 2048,
+      "pattern": "^[^\\u0000]+$"
+    },
+    "workspaceRelativePath": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 1024,
+      "pattern": "^(?![~/])(?!.*\\\\)(?!.*//)(?!.*(?:^|/)\\.{1,2}(?:/|$))[^\\u0000]+$"
+    },
+    "cacheReferencePath": {
+      "type": "string",
+      "pattern": "^\\.grip/cache/repos/[A-Za-z0-9][A-Za-z0-9._-]{0,127}\\.git$"
+    },
+    "stagedInputPath": {
+      "type": "string",
+      "pattern": "^\\.grip/staging/inputs/[A-Za-z0-9][A-Za-z0-9._-]{0,127}$"
+    },
+    "cloneOperation": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "kind",
+        "repo_url",
+        "dest_path",
+        "branch"
+      ],
+      "properties": {
+        "kind": {
+          "const": "clone"
+        },
+        "repo_url": {
+          "$ref": "#/$defs/nonEmptyString"
+        },
+        "dest_path": {
+          "$ref": "#/$defs/workspaceRelativePath"
+        },
+        "branch": {
+          "$ref": "#/$defs/nonEmptyString"
+        },
+        "reference_base": {
+          "$ref": "#/$defs/cacheReferencePath"
+        }
+      }
+    },
+    "venvOperation": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "kind",
+        "dest_path",
+        "engine",
+        "python"
+      ],
+      "properties": {
+        "kind": {
+          "const": "venv"
+        },
+        "dest_path": {
+          "$ref": "#/$defs/workspaceRelativePath"
+        },
+        "engine": {
+          "const": "uv"
+        },
+        "python": {
+          "$ref": "#/$defs/nonEmptyString"
+        }
+      }
+    },
+    "editableInstallOperation": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "kind",
+        "venv_path",
+        "source_path",
+        "extras"
+      ],
+      "properties": {
+        "kind": {
+          "const": "editable_install"
+        },
+        "venv_path": {
+          "$ref": "#/$defs/workspaceRelativePath"
+        },
+        "source_path": {
+          "$ref": "#/$defs/workspaceRelativePath"
+        },
+        "extras": {
+          "type": "array",
+          "uniqueItems": true,
+          "items": {
+            "type": "string",
+            "minLength": 1,
+            "maxLength": 128,
+            "pattern": "^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$"
+          }
+        }
+      }
+    },
+    "projectFileOperation": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "kind",
+        "source_path",
+        "source_sha256",
+        "dest_path",
+        "mode"
+      ],
+      "properties": {
+        "kind": {
+          "const": "project_file"
+        },
+        "source_path": {
+          "$ref": "#/$defs/stagedInputPath"
+        },
+        "source_sha256": {
+          "$ref": "#/$defs/sha256"
+        },
+        "dest_path": {
+          "$ref": "#/$defs/workspaceRelativePath"
+        },
+        "mode": {
+          "const": "copy"
+        }
+      }
+    }
+  }
+}
--- a/gr2/prototypes/lane_workspace_prototype.py
+++ b/gr2/prototypes/lane_workspace_prototype.py
@@ -47,6 +47,13 @@ class LaneMetadata:
     handoff_source: dict[str, str] | None
 
     def as_toml(self) -> str:
+        # repos/shared_with are pre-formatted here, not inlined into the
+        # f-strings below: an f-string expression part can't safely nest a
+        # string literal using the same quote character as the outer
+        # f-string (pre-3.12 tokenizers can't tell a nested string's quote
+        # from the outer delimiter -- see grip#793's Python 3.11 CI catch).
+        repos_toml = ", ".join(f'"{r}"' for r in self.repos)
+        shared_with_toml = ", ".join(f'"{u}"' for u in self.shared_with)
         lines = [
             f"schema_version = {self.schema_version}",
             f'lane_name = "{self.lane_name}"',
@@ -55,8 +62,8 @@ class LaneMetadata:
             f'lane_type = "{self.lane_type}"',
             f'creation_source = "{self.creation_source}"',
             "",
-            f"repos = [{', '.join(f'\"{r}\"' for r in self.repos)}]",
-            f"shared_with = [{', '.join(f'\"{u}\"' for u in self.shared_with)}]",
+            f"repos = [{repos_toml}]",
+            f"shared_with = [{shared_with_toml}]",
             "",
             "[branch_map]",
         ]
@@ -116,6 +123,11 @@ class SharedScratchpad:
     updated_at: str
 
     def as_toml(self) -> str:
+        # See LaneWorkspaceSpec.as_toml's comment above: pre-formatted here
+        # rather than inlined, to avoid nesting a same-quote string literal
+        # inside another f-string's expression part (grip#793).
+        participants_toml = ", ".join(f'"{p}"' for p in self.participants)
+        linked_refs_toml = ", ".join(f'"{r}"' for r in self.linked_refs)
         lines = [
             f"schema_version = {self.schema_version}",
             f'name = "{self.name}"',
@@ -126,8 +138,8 @@ class SharedScratchpad:
             f'created_at = "{self.created_at}"',
             f'updated_at = "{self.updated_at}"',
             "",
-            f'participants = [{", ".join(f"\"{p}\"" for p in self.participants)}]',
-            f'linked_refs = [{", ".join(f"\"{r}\"" for r in self.linked_refs)}]',
+            f"participants = [{participants_toml}]",
+            f"linked_refs = [{linked_refs_toml}]",
             "",
             "[paths]",
             f'docs_root = "{self.docs_root}"',
--- a/gr2/pyproject.toml
+++ b/gr2/pyproject.toml
@@ -24,12 +24,19 @@ dev = [
 ]
 
 [tool.setuptools]
-packages = ["gr2_overlay", "gr2", "gr2.python_cli", "gr2.prototypes"]
+# gr2_overlay is a time-bounded compat alias for gr2.overlay (grip config#491
+# gap#13): same physical directory, two import names. New materialization
+# code must import gr2.overlay; gr2_overlay stays for existing consumers
+# until they're migrated separately. Don't mix the two names for the same
+# object within one execution path -- Python treats them as distinct module
+# identities even though the underlying files are identical.
+packages = ["gr2_overlay", "gr2", "gr2.python_cli", "gr2.prototypes", "gr2.overlay"]
 
 [tool.setuptools.package-dir]
 gr2 = "gr2"
 "gr2.python_cli" = "python_cli"
 "gr2.prototypes" = "prototypes"
+"gr2.overlay" = "gr2_overlay"
 
 [tool.pytest.ini_options]
 testpaths = ["tests", "gr2_overlay/tests"]
--- a/gr2/pyproject.toml
+++ b/gr2/pyproject.toml
@@ -12,6 +12,14 @@ dependencies = [
     "tomli_w>=1.0",
     "pyyaml>=6.0",
     "toml>=0.10",
+    # Draft 2020-12 validation of the pinned MaterializationPlan v1 schema
+    # (config#492). 4.18 is the first release with the modern referencing
+    # stack; anything older silently falls back to a legacy resolver.
+    "jsonschema>=4.18",
+    # PEP 440 specifier evaluation for the venv probe (config#492 §6.2.1
+    # #6): an existing venv must satisfy the plan's declared `python`
+    # constraint, not merely exist.
+    "packaging>=23",
 ]
 
 [project.scripts]
@@ -37,6 +45,12 @@ gr2 = "gr2"
 "gr2.python_cli" = "python_cli"
 "gr2.prototypes" = "prototypes"
 "gr2.overlay" = "gr2_overlay"
+
+[tool.setuptools.package-data]
+# The pinned MaterializationPlan v1 schema (config#492) ships inside the
+# gr2 package -- spec_apply verifies its bytes against the pinned SHA-256
+# at load and fails closed on any mismatch.
+gr2 = ["schemas/*.json"]
 
 [tool.pytest.ini_options]
 testpaths = ["tests", "gr2_overlay/tests"]
--- a/gr2/python_cli/spec_apply.py
+++ b/gr2/python_cli/spec_apply.py
@@ -61,6 +61,12 @@ _FORBIDDEN_IDENTITY_KEYS = frozenset(
 
 
 def _reject_identity_fields_recursive(value: object, *, path: str) -> None:
+    """Defense in depth alongside the per-kind field allowlist below: even
+    an allowed field's VALUE should never carry an identity-shaped key.
+    The allowlist is what actually PROVES identity-freedom (round 2,
+    Atlas/Sentinel: "a blacklist cannot prove identity-free construction"
+    -- display_name, or any field name nobody enumerated, sails through a
+    blacklist undetected); this stays as a second layer, not the primary one."""
     if isinstance(value, dict):
         present = _FORBIDDEN_IDENTITY_KEYS & value.keys()
         if present:
@@ -86,23 +92,31 @@ def _validate_path_safe_token(value: object, *, field_name: str) -> str:
     return value
 
 
-def _resolve_contained(workspace_root: Path, relative: str, *, field_name: str) -> Path:
-    """config#491 §3: absolute paths, .., symlink escapes, and duplicate
-    destinations are blockers. Path.resolve() normalizes .. lexically and
-    follows any existing symlinks, so a single containment check after
-    resolving catches all three escape shapes. Shared by dest_path,
-    reference_base, and (via _resolve_staging_source) source_path."""
+def _canonicalize_workspace_path(workspace_root: Path, relative: str, *, field_name: str) -> Path:
+    """config#491 §3: absolute paths, .., symlink escapes, and duplicate/
+    case-folded-colliding destinations are blockers. Round 2 (Atlas): a
+    literal '..' segment that happens to resolve back inside the workspace
+    was being accepted -- the spec forbids '..' appearing at all, not just
+    "does it escape after resolution." Checked as a literal path-part scan
+    BEFORE resolving, in addition to the post-resolve containment check
+    (which independently catches symlink escapes resolve() reveals).
+    Returns the fully resolved (symlink-dereferenced) canonical path --
+    callers needing the original relative string for receipts/display keep
+    that separately; this return value is for filesystem operations and
+    ALL comparison (duplicate detection, cache-namespace checks)."""
     if not relative or relative.startswith("/") or relative.startswith("~"):
         raise MaterializationPlanError(f"{field_name} must be relative to the workspace root: {relative!r}")
+    if ".." in Path(relative).parts:
+        raise MaterializationPlanError(f"{field_name} must not contain '..' segments: {relative!r}")
     workspace_resolved = workspace_root.resolve()
     candidate = (workspace_root / relative).resolve()
     if candidate != workspace_resolved and workspace_resolved not in candidate.parents:
         raise MaterializationPlanError(f"{field_name} escapes the workspace root: {relative!r}")
-    return workspace_root / relative
+    return candidate
 
 
 def _resolve_safe_dest(workspace_root: Path, dest_path: str) -> Path:
-    return _resolve_contained(workspace_root, dest_path, field_name="dest_path")
+    return _canonicalize_workspace_path(workspace_root, dest_path, field_name="dest_path")
 
 
 _STAGING_INPUTS_SUBPATH = Path(".grip") / "staging" / "inputs"
@@ -112,14 +126,45 @@ def _resolve_staging_source(workspace_root: Path, source_path: str) -> Path:
     """config#491 §6.2, Sentinel r2 P4 ruling: source_path MUST be
     workspace-relative AND resolve beneath .grip/staging/inputs/ specifically
     -- not just anywhere in the workspace. The normative example is
-    `.grip/staging/inputs/f_01`; S7 emits `.grip/staging/inputs/<opaque-key>`."""
-    resolved = _resolve_contained(workspace_root, source_path, field_name="source_path").resolve()
+    `.grip/staging/inputs/f_01`; S7 emits `.grip/staging/inputs/<opaque-key>`.
+    _canonicalize_workspace_path already follows symlinks via .resolve(), so
+    a symlink planted inside staging/inputs pointing outside it is caught by
+    the same containment check -- pinned by a real symlink test, not just
+    reasoned about."""
+    resolved = _canonicalize_workspace_path(workspace_root, source_path, field_name="source_path")
     staging_root = (workspace_root / _STAGING_INPUTS_SUBPATH).resolve()
     if resolved != staging_root and staging_root not in resolved.parents:
         raise MaterializationPlanError(
             f"source_path must resolve beneath .grip/staging/inputs/: {source_path!r}"
         )
-    return workspace_root / source_path
+    return resolved
+
+
+_CACHE_NAMESPACE_SUBPATH = Path(".grip") / "cache" / "repos"
+
+
+def _validate_reference_base(workspace_root: Path, reference_base: str, *, repo_url: str) -> Path:
+    """config#491 §8.2, round 2 (Atlas/Sentinel): reference_base must be
+    bound to the declared cache NAMESPACE (.grip/cache/repos/*.git), not
+    just anywhere inside the workspace -- a real in-workspace mirror
+    outside that namespace was being accepted and retained as the clone's
+    alternate. Separately, if the referenced cache already exists, its own
+    canonical origin must equal repo_url -- otherwise a cache seeded from a
+    different remote could be used as if it were the right one."""
+    resolved = _canonicalize_workspace_path(workspace_root, reference_base, field_name="reference_base")
+    cache_namespace = (workspace_root / _CACHE_NAMESPACE_SUBPATH).resolve()
+    if resolved.parent != cache_namespace or resolved.suffix != ".git":
+        raise MaterializationPlanError(
+            f"reference_base must be a *.git cache directly inside .grip/cache/repos/: {reference_base!r}"
+        )
+    if resolved.exists():
+        cache_origin = remote_origin_url(resolved)
+        if cache_origin is not None and cache_origin != repo_url:
+            raise MaterializationPlanError(
+                f"reference_base {reference_base!r} has canonical origin {cache_origin!r}, "
+                f"expected {repo_url!r} -- a cache seeded from a different remote cannot be reused"
+            )
+    return resolved
 
 
 # ---------------------------------------------------------------------------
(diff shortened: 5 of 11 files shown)


laynepenney (commented):
REQUEST CHANGES — binding r2, source-author fidelity against config#491 §§6.2, 7.3, 8.1-8.3, 12.1, and 15.

The good fruit is real: the grip#539 correction is right, `gr2.overlay` is the correct canonical import, explicit destination paths work, and the targeted executor suite is 17/17 green here. The blockers below are contract survivors, each reproduced against the current implementation rather than inferred from style.

## P1 — §7.3 still has two executors

`apply_plan()` at `spec_apply.py:288-378` and `apply_materialization_plan()` at `:709-750` are independent operation loops with separate dispatch and receipt behavior. Sharing the leaf `clone_repo()` primitive does not satisfy “one executor, not a second materializer.”

Closure: retain two public adapters if useful, but both must validate/normalize into one typed operation model and delegate to one canonical execution/receipt path. Add a mutation guard proving that bypassing the shared executor from either adapter turns RED.

## P2 — §8.3 atomic clone publication is absent

`_apply_clone_operation()` clones directly into the final `dest` and only then checks out the branch and validates isolation. Exact fruit: I supplied a valid remote plus a nonexistent branch. Apply failed during checkout, but the final destination and its `.git` directory remained published.

Closure: clone into a sibling staging path, checkout and validate there, then atomically rename into the final destination. A post-clone checkout/validation failure must leave the final destination absent and must not create a plausible completed receipt. Pin this with an injected failure test.

## P3 — §8.1/§8.2 validation accepts three forbidden clone states

1. `reference_base` is not resolved through a containment guard. An absolute cache outside the workspace was accepted, passed to `--reference-if-able`, and then treated as the approved alternate because approval is derived from the caller-supplied path itself.
2. An existing clone whose actual `origin` was remote A was silently accepted under a new plan declaring remote B.
3. An existing clone with `.git/worktrees` present was silently accepted. The implementation checks only “`.git` is a directory” plus the alternates file. It does not prove the §8.1 common-dir, origin, or no-worktrees invariants.

Closure: `reference_base` is workspace-relative and confined to `.grip/cache/repos/<declared-repo>.git`. Reject absolute paths, `..`, symlink escape, or another cache root before invoking Git. Existing and staged clones must prove canonical origin, `git rev-parse --git-common-dir` resolves to their own `.git`, `.git/worktrees` is absent, and every alternate resolves only to the validated declared cache. Add one mutation-loud test per degree of freedom.

Source ruling on branch state: `branch` is first-materialization checkout intent. Do not force-switch a healthy existing developer clone on rerun. Origin and isolation still must match.

## P4 — `project_file.source_path` contract is workspace-relative staging, not absolute

The normative §6.2 value is `.grip/staging/inputs/f_01`, and §3 says all generated paths are relative to the team root. The exact spec-shaped plan currently fails because `Path(source_path)` resolves against the process cwd. Conversely, the current absolute-path fixtures authorize a plan to read and unlink an arbitrary file outside the workspace. I reproduced both outcomes.

Source ruling for S4 and S7:

- `source_path` MUST be workspace-relative.
- It MUST resolve beneath `<workspace_root>/.grip/staging/inputs/`.
- Absolute, `~`, `..`, outside-staging paths, and symlink escapes are blockers.
- S7 emits `.grip/staging/inputs/<opaque-artifact-key>`.
- `mode` is `copy` for v1. Any other value is rejected, not ignored.
- The staging artifact is removed only after the hash-verified projection has durable acknowledgement.

Replace the absolute happy-path fixtures with the normative relative shape. Add RED tests for absolute, traversal, workspace-inside-but-outside-staging, and symlink escape. The current absolute-source deletion test must invert.

## P5 — the plan is not closed or validate-before-touch, and gap #5 is not yet closed

The executor validates one operation immediately before executing it. It does not validate the whole plan first. Exact fruit:

- operation 1 created a venv, then operation 2’s absolute destination failed, leaving mutation before plan validation completed
- top-level `agent_name` plus nested `metadata.channel` were accepted
- `plan_id="../../../../escaped_receipt"` wrote the receipt outside the workspace
- an arbitrary existing directory with `engine="not-uv"` and `python=">=99"` was accepted as a valid existing venv

The receipt also omits the neutral plan hash and opaque unit key required by §12.1, is written directly rather than atomically, and carries too little evidence to verify clone origin/HEAD/cache, venv interpreter, or editable PEP 610 state.

I own one ambiguity in the merged JSON example: it mentions `unit_key` in prose but does not place it structurally. Ruling: one MaterializationPlan is scoped to one opaque unit and carries required top-level `unit_key`. gr2 validates it as a path-safe opaque token without interpreting identity, and the neutral receipt records it. S7 compiles one plan per unit.

Closure:

1. Validate the complete plan before the first filesystem mutation.
2. Use a closed schema per operation kind, including safe `plan_id`, required top-level `unit_key`, recursive identity-field rejection, duplicate-destination rejection, `engine=="uv"`, and `mode=="copy"`.
3. Validate existing venv interpreter and editable evidence instead of treating path existence as success.
4. Hash the accepted canonical plan and write the §12.1 structural receipt atomically last.
5. Mutation-test late-invalid-operation/no-touch, top-level and nested identity smuggling, receipt-path traversal, fake existing venv, missing unit key, and `>`/field-removal equivalents for receipt evidence.

This is a REQUEST CHANGES verdict, not a rejection of the implementation direction. The grip#539 and namespace pieces stand. The S4 executor needs these seams closed before S7 compiles premium plans into it.

laynepenney (commented):
ATLAS spec-author r2: **BINDING REQUEST CHANGES** at `03fe3109fdb5fb5dd654450ddcb75f15d8a93fff`.

I read the revised executor and tests against config#491 §§3, 6.2, 7.3, 8.1-8.3, and 12, then replayed the prior failures against real local remotes. The revision genuinely closes several seams:

- `project_file.source_path` is now workspace-relative beneath `.grip/staging/inputs/`, mode is `copy`, and the spec-shaped path passes.
- clone publication is stage, validate, then rename. A nonexistent branch leaves neither the final destination nor staging residue.
- existing-clone origin, common-dir, `.git/worktrees`, and worktree-pointer checks are real.
- the required opaque `unit_key` is validated without identity interpretation and appears in the receipt. This satisfies the gr2 half of the one-plan-per-unit ruling. S7 remains responsible for compiling exactly one plan per unit.
- 43/43 targeted tests and 551/551 full gr2 tests pass in the declared uv environment.

Four source-contract survivors remain.

### 1. BLOCKER: P1 is still conditional, and §7.3 still has two executors

`build_plan()` emits `converge_unit_repos` only when at least one declared repo is absent (`spec_apply.py:529-531`). The new loop validates `all_repos`, but only after that operation exists (`:614-631`). The test deliberately includes a second missing repo to trigger the path (`test_materialization_plan.py:318-321`), so it cannot prove "all declared repos every pass."

Fruit: with every declared unit repo present and one clone carrying the wrong origin, `apply_plan()` returned successfully with `operation_count=0`. The exact isolation violation the MaterializationPlan path rejects therefore still bypasses the legacy path.

The structural issue also remains: `apply_plan()` owns one dispatch loop (`:561`) and `apply_materialization_plan()` owns another (`:1048`); neither delegates to the other. `_clone_isolated` is now an excellent shared clone primitive, but config#491 §6.2/§7.3 says one `apply_plan` executor serves direct OSS and premium-compiled plans, not merely one clone primitive.

Closure: route both plan forms through one canonical materialization operation executor, validate every declared existing unit repo even when none are missing, and add the all-present/wrong-origin mutation fixture. Removing that unconditional validation must make the test RED.

### 2. BLOCKER: §8.2 cache-only alternates are not enforced

`reference_base` is only checked for workspace containment (`:936-938`), then whatever caller path was supplied becomes the approved alternate (`:198-209`). Two real probes remain green:

- `reference_base="rogue/object-bank.git"` inside the workspace was accepted and retained as the clone alternate.
- `.grip/cache/repos/product.git` backed by remote B was accepted while cloning remote A.

Config#491 §8.2 permits only `<team-root>/.grip/cache/repos/<declared-repo>.git/objects`, requires canonical cache remote verification, and records dependents.

Closure: bind `reference_base` to the declared cache namespace, verify it is the expected bare cache with canonical origin equal to `repo_url`, and mutation-test both an in-workspace non-cache path and a canonical-looking wrong-origin cache.

### 3. BLOCKER: the MaterializationPlan is not a closed typed contract

The revision validates operations but not the whole plan recursively. A top-level `agent_id` is accepted because `_reject_identity_fields_recursive()` is called only on each operation (`:977-981`). The validator also accepts:

- `schema_version=999`
- a missing `workspace_spec_sha256`
- a `..` path that normalizes to the same destination
- case-folded destination collisions such as `.venv` and `.VENV`

Those contradict the v1 shape and config#491 §3's explicit `..`, duplicate-normalized-destination, and case-folded-collision blockers. Per-kind unknown fields also remain open, so a blacklist cannot prove identity-free construction.

Closure: exact top-level and per-kind allowlists with field types, exact `schema_version == 1`, required verified WorkspaceSpec hash, recursive plan-level identity rejection, rejection of every `..` segment, and collision detection over normalized case-folded destinations. Each named survivor needs a RED mutation.

### 4. BLOCKER: §12.1 receipts are still materially incomplete

The conforming cached-clone receipt has top-level keys:

`applied_at, operations, plan_hash, plan_id, schema_version, unit_key, workspace_spec_sha256`

and clone evidence keys:

`dest_path, first_materialize, head_sha, kind, repo_url`

It omits §12.1's structural stage, observed origin/clone-isolation evidence, cache path and approved-alternate evidence, venv interpreter, editable extras and distribution, and full projection path evidence. The current receipt test checks only that broad keys exist (`test_materialization_plan.py:590-607`); a constant plan hash or deletion of every operation-specific evidence field remains GREEN.

Closure: define and assert the exact per-operation receipt schemas, recompute the plan hash in the test, and mutation-test deletion or falsification of each load-bearing evidence class. The atomic temp-to-final write itself is clean.

★ Insight

- The revision hardened the shared clone primitive, but a primitive cannot establish one-executor equivalence when the legacy scheduler can skip it entirely.
- The current tests prove the five reported examples only under their triggering fixtures. The surviving all-present clone, in-workspace rogue cache, and sparse receipt show why source-contract review must test the complement as well.

I will re-review the narrowed closure. The project-file ruling, atomic clone path, and opaque-unit ruling stand and do not need redesign.

laynepenney (commented):
SENTINEL R2 RE-GATE: REQUEST CHANGES remains binding at `03fe3109fdb5fb5dd654450ddcb75f15d8a93fff`.

I re-ran the closure fruit rather than accepting the revision narrative. Pristine is green: 43/43 targeted and 551/551 full gr2 tests. P2 is genuinely mutation-loud. P3 origin and `.git/worktrees`, P4 staging-root confinement, and P5 whole-plan-before-touch are also genuinely loud. The full five-finding closure is not complete:

1. **P1 still skips the claimed every-pass validation.** I reproduced an existing unit with one declared repo, all repos present, but its origin differs from the declared URL. `build_plan` emits no `converge_unit_repos` because `missing_repos` is empty, so `apply_plan` returns successfully and `_clone_isolated` is never reached. The new test avoids that state by adding a second missing repo solely to schedule convergence. That does not prove “all declared repos every pass,” and the code still has two independent dispatch/receipt loops rather than §6.2/§7.3's one `apply_plan` executor. Closure needs the all-present mismatch test and a canonical shared execution path, not a missing-repo trigger.

2. **P3 remains incomplete and partly mutation-silent.** Removing both `reference_base` containment calls leaves `test_clone_rejects_reference_base_escaping_workspace` GREEN because its `https://example.com/x.git` clone fails later and raises the same broad exception. The observed error was `clone staging failed`, not a containment rejection. A real in-workspace mirror at `units/u1/rogue-cache.git` is accepted as `reference_base`, and its alternate is recorded, despite the contract limiting this to `.grip/cache/repos/<declared-repo>.git`. Removing the entire `git-common-dir` comparison also leaves all 34 materialization tests GREEN. Use a valid local remote or a clone-call spy for the containment test, add the other-cache-root case, and pin both common-dir mismatch and nonzero-command fail-closed behavior. The origin and nested-worktrees mutants are closed.

3. **P4 still lacks two required witnesses.** Replacing `_resolve_staging_source` with a lexical staging-prefix check leaves all 34 tests GREEN and permits a symlink under `.grip/staging/inputs` to copy bytes from outside the workspace. Add the exact symlink-escape test required by the prior closure. Separately, faulting `_write_materialization_receipt` after a valid `project_file` operation leaves `source_exists=False`, `dest_exists=True`, `receipt_exists=False`. The staged artifact is deleted before durable acknowledgement. Pin receipt failure and keep the source until durable operation acknowledgement exists.

4. **P5 is not a closed schema or mutation-locked receipt.** A venv op carrying undeclared `metadata.display_name="Apollo"` is accepted and mutates disk. Raw-string duplicate checking accepts `units/u1/.venv` plus `units/u1/./.venv`; op1 creates the venv, then op2 fails, violating validate-before-touch. A directory containing only a forged `pyvenv.cfg`, with no interpreter, is accepted under `python=">=99"`, although the prior closure required interpreter validation. Replacing temp-write-plus-rename with direct `receipt_path.write_text(...)` leaves all 34 tests GREEN, so `test_receipt_write_leaves_no_temp_file_behind` does not prove atomic publication. Finally, §12.1 evidence is still absent: clone results do not record observed origin/cache/approved-alternate evidence, venv results do not record the interpreter, and editable results omit extras/distribution. Close allowed-key schemas at plan and operation levels, canonicalize duplicate destinations before comparison, validate the actual venv interpreter/constraint, and add fault/mutation tests for atomic receipt publication and required evidence fields.

This is not a rejection of the direction. The shared clone primitive is materially better, and P2 is closed. But I cannot flip my binding verdict while exact prior closure clauses and mutants still survive.

laynepenney (commented):
ATLAS spec-author r2: **BINDING REQUEST CHANGES** at exact `e46b8e729c53407b4bbb34cd4a5e3aee8812303b`.

I reviewed the published round-2 code and tests, ran the full suite (`565 passed` on Python 3.11.15), replayed the original survivors, and compared the executor directly with config#492 v1 (`f57b3e4`, schema SHA `a5061501…590231c`). P1 is genuinely closed. P3, P4, P5, and the receipt contract are not.

### Clean

1. **P1 / one executor and present-unit validation: CLEAN.** Both legacy clone call sites now delegate through `_apply_clone_operation`, and `converge_unit_repos` is scheduled whenever a unit declares repos. I restored the old `if missing_repos` mutant locally. Both the all-present/wrong-origin test and the scheduling test went RED (`2 failed`). This is a real guard.

### Blocking survivors

1. **P5: config#492 v1 is not wired.** There is no packaged schema, no pin check, no Draft 2020-12 validator, and no schema dependency. `_validate_operation_schema` is a separate, looser hand validator. I passed nine plans that the pinned schema rejects:
   - clone without required `branch`
   - venv without required `engine` and `python`
   - editable install without required `extras`
   - project file without required `mode`
   - six-character `workspace_spec_sha256`
   - `.` path segment
   - backslash path
   - duplicate extras
   - 129-character `plan_id`

   The executor also applies with no canonical WorkspaceSpec file at all, so `workspace_spec_sha256` is carried but never verified against reopened bytes. Closure requires validating the parsed plan against the exact packaged/pinned v1 schema before touch, then enforcing every §6.2.1 runtime invariant. Do not preserve the current defaulting/coercion behavior.

2. **P5/P1 runtime invariants: forged venv and declared branch still pass.** A directory with `pyvenv.cfg` plus an executable shell script at `bin/python3` was accepted and receipted as an existing venv. The required isolated bounded probe (`sys.prefix == resolved venv`, `sys.prefix != sys.base_prefix`) is absent. Separately, an existing healthy clone on `main` accepted a plan declaring `branch: feature` and remained on `main`. Required plan fields must be validated as desired state, not only used during first creation.

3. **P3: cache provenance is still fail-open, and the common-dir guard test is masked.** `_validate_reference_base` accepts an existing cache when `remote_origin_url` returns `None`. A non-Git directory at the canonical `.grip/cache/repos/product.git` path was accepted. The clone completed without an alternates file, yet the receipt said `alternate_approved=true`. Require a real workspace-managed bare cache, exact canonical remote equality including rejecting absent provenance, and proof that the resulting clone's only alternate is that cache's `objects` directory.

   I also restored the exact old fail-open common-dir block (`if returncode == 0`) and `test_clone_rejects_when_common_dir_check_cannot_run` still passed (`1 passed`). Its invalid repo later fails the origin check, satisfying the same broad `assertRaises`. Isolate this guard with a healthy repo whose origin is correct while only the common-dir probe is forced to fail, or match the specific failure.

4. **P4 and §12: staging cleanup still precedes durable acknowledgement.** An in-workspace symlink prefix `.grip/staging/inputs -> <workspace>/rogue` was accepted. The projected file was copied and the rogue source deleted. The contract requires a symlink-free prefix and a regular non-symlink source, not merely a resolved path that remains somewhere inside the workspace.

   Receipt publication calls `write_text` then `Path.rename` with no file flush/fsync and no parent-directory fsync. Patching `os.fsync` to raise produced zero calls, a final receipt, and a deleted staging source. This is not the durable acknowledgement pinned in §6.2.1. The receipt's `plan_hash` also differs from the pinned canonical serialization because it uses default JSON separators/escaping. The success receipt still omits structural stage and the project source path. Implement same-directory temp write, flush+file fsync, atomic replace, directory fsync, then cleanup, with idempotent post-receipt cleanup on rerun.

### Invariant ruling

**Pin both.** Unit-clone origin validation when present and forged-venv rejection are contract invariants, not executor-local choices. They are already canonical in config#492 §6.2.1 at `f57b3e4`, with binding r1+r2 approval. Round 2 currently satisfies the first but not the second.

The restructuring direction is right, and P1 proves it can be mutation-loud. The gate remains RED because the executable contract is still materially weaker than the now-pinned v1 contract.

laynepenney (commented):
SENTINEL R2 RE-GATE — BINDING REQUEST CHANGES HOLDS at exact `e46b8e729c53407b4bbb34cd4a5e3aee8812303b`.

Pristine fruit is green: 57/57 targeted and 565/565 full gr2 tests. Several closures are genuine and mutation-loud: all-present convergence scheduling, atomic clone publication, real-URL cache containment, wrong-origin and nested-worktree rejection, nonzero common-dir fail-closed behavior, staging-root and outside-symlink rejection, pre-receipt unlink ordering, top-level and per-kind unknown-field rejection, canonical dot-alias collision, direct receipt-write rejection, and the new operation-evidence fields.

I cannot flip while the following exact survivors remain.

1. **P1 behavior is fixed, but the one-executor seam is still mutation-silent.** Replacing the legacy `converge_unit_repos` call to `_apply_clone_operation` with a behavior-equivalent direct `_clone_isolated` call leaves `TestSharedExecutor` plus all convergence tests 11/11 GREEN. `apply_plan` and `apply_materialization_plan` still own independent dispatch and receipt loops. The new all-present scheduling mutant correctly turns RED, but it does not prove that bypassing the shared operation executor from either adapter turns RED, which was the prior closure. Route both adapters through one shared dispatch seam and pin a spy or mutation test for each adapter.

2. **P3 still lacks the common-dir comparison guard, and missing cache identity fails open.** Replacing `if common_dir != git_path.resolve()` with a disabled comparison leaves all 48 materialization tests GREEN. I built a real production-shaped fixture with `.git` as a directory and `git rev-parse --git-common-dir` resolving to another checkout. Pristine rejects it, so it is locally testable without mocking. Separately, a canonical `.grip/cache/repos/product.git` directory with no canonical origin is accepted, used as `reference_base`, and receipted as `alternate_approved=True`. Wrong-origin, wrong-namespace, and command-failure mutants are loud, but `None` origin must reject and the actual comparison needs its own fixture.

3. **config#492's closed schema is not wired.** Running config#492's exact 12 malformed fixtures against `_validate_plan_shape` yields only 8/12 RED. These four survive: missing clone `branch`, a literal `.` path segment, backslash path, and nested `.grip/staging/inputs/a/b`. Additional schema-invalid inputs also survive: missing `engine`, `python`, `extras`, or `mode`; non-SHA WorkspaceSpec hash; `schema_version=True`; and invalid opaque tokens. The executor also accepts and mutates when `.grip/workspace_spec.toml` is absent, so `workspace_spec_sha256` is never reopened or verified. This is a partial handwritten allowlist, not conformance to the approved v1 contract. Validate the complete normative schema before mutation, then enforce the runtime-only clauses.

4. **P4's symlink-free and durable-ack clauses are incomplete.** An internal symlink `.grip/staging/inputs/alias -> .../inputs/real` is accepted. The projection is written and cleanup unlinks the resolved target, leaving the alias dangling. config#492 rejects any existing symlink in the prefix and requires the source itself to be non-symlink, not only symlink escapes outside staging. Receipt publication currently performs `write_text` plus `rename` with zero `fsync` calls, then deletes the staged source. Rename success is not the approved durable acknowledgement. Add explicit symlink-component checks plus file flush and fsync, atomic replace, and parent-directory fsync before unlink, with ordering instrumentation.

5. **The venv proof still trusts an executable-looking file.** A directory with `pyvenv.cfg` and executable `bin/python` whose body exits 42 is accepted under `python=">=99"` as an existing venv. The approved contract requires a bounded isolated probe proving resolved `sys.prefix == declared venv`, `sys.base_prefix != sys.prefix`, and the declared interpreter constraint. The no-interpreter mutant is loud, but this stronger production-shaped mutant survives pristine.

6. **P5 hash, collision, and receipt locks remain incomplete.** Setting `plan_hash = "0" * 64` leaves all seven relevant receipt and evidence tests GREEN. The pristine hash also differs from config#492's required `json.dumps(..., sort_keys=True, separators=(",", ":"), ensure_ascii=False)` bytes. Removing `.casefold()` leaves all 13 plan-shape tests GREEN because no case-only collision fixture exists. Finally, the observed project-file receipt has only `kind`, `dest_path`, and `source_sha256`. It omits the source path required by §12.1, and the receipt has no structural-stage evidence. Add recomputation, case-only collision, and exact receipt-schema assertions.

This is a narrowed hold, not a rejection of the restructure. The named green fruit above stands. The binding criterion remains: every prior mutant must die under production-shaped config, and #492 is now that config.

laynepenney (commented):
ATLAS spec-author r2, final round: **BINDING REQUEST CHANGES** at exact PR head `82a3a232322dee729e67ebe5352575be376a1052`.

The round-3 restructure closes most of the prior gate. I independently verified the vendored schema is byte-identical to config#492 (`a5061501ba6651d7432d87d57f1c85902e5dec076f860a47faa299f5f590231c`), Draft 2020-12 validation runs first and fails closed, the full malformed-plan set rejects, WorkspaceSpec bytes are reopened and hashed, shared-executor/all-present-origin routing is real, forged venvs and wrong branches reject, the old common-dir/hash/casefold/symlink/fsync mutants are now loud, and the clean suite is `614 passed` on both Python 3.11 and 3.13.

Three production-shaped survivors remain:

1. **§6.2.1 #5 is still fresh-clone-only for required alternate realization.** `_validate_clone_isolation` returns when the alternates file is absent (`spec_apply.py:353-355`) even when `reference_repo_root` is declared. The missing-alternate rejection at `443-452` runs only after a fresh clone. Fruit: create a valid canonical mirror cache, create an existing ordinary clone with no alternates, then apply a plan declaring that `reference_base`. Apply succeeds and receipts `alternate_approved=false`. The contract says an existing destination is validated on every apply and, when `reference_base` is present, the resulting clone's only alternate is that cache. Add the existing-clone/no-alternate fixture and reject on the specific alternate invariant.

2. **§6.2.1 #7/#8 rerun cleanup bypasses staged-source verification.** The `dest`-already-correct branch at `1216-1228` schedules any existing `source_path` for deletion without checking that it is regular or reopening and hashing its bytes. Fruit: apply a project file, recreate the staged path with tampered bytes, rerun. Apply succeeds and deletes the tampered source. A cleanup rerun must verify the staged source and the receipt evidence before scheduling destructive cleanup. Add the exact tampered-after-receipt fixture and require an invariant-specific hash rejection with the source preserved.

3. **§9.2/§12.1 PEP 610 evidence is not bound to the editable being installed.** `_find_direct_url_evidence` returns the first sorted `direct_url.json` in the venv (`1151-1160`). Fruit: one plan installs editable `alpha` then editable `beta`. The second operation receipts `distribution="alpha"` and alpha's PEP 610 path. This makes a structurally complete receipt false. Select the evidence whose decoded `url` equals the requested resolved source and whose `dir_info.editable` is true, then derive the distribution from that matched row. Add a two-editable fixture so a first-row/unfiltered mutant goes RED.

My three direct fruit tests are `3 failed`, each because the expected guard did not fire. The committed suite remaining green is therefore evidence of missing gates, not harmless edge behavior. No other formula or architecture change is requested.

laynepenney (commented):
SENTINEL R2 FINAL RE-GATE: BINDING APPROVE at exact `82a3a232322dee729e67ebe5352575be376a1052`.

This supersedes my binding REQUEST CHANGES at `e46b8e7`.

I independently reran the full round-3 union in a detached exact-head worktree, one source mutation at a time with a clean restoration between probes. I executed 21 one-seam probes, splitting receipt durability into separate file-fsync and parent-directory-fsync mutations. Every mutation turned RED for its intended invariant, then the restored code returned GREEN.

The four exact prior survivors are closed:

1. Replacing the legacy convergence call to `_apply_clone_operation` with a behavior-equivalent direct `_clone_isolated` call fails the shared-seam witness with only one observed dispatch.
2. Disabling `common_dir != git_path.resolve()` makes the foreign-common-dir fixture fail with `MaterializationPlanError not raised`.
3. Returning a constant 64-character plan hash fails both independent canonical-byte recomputation witnesses.
4. Removing `.casefold()` makes both the ASCII case-only and Unicode `ß`/`SS` collision fixtures fail with no rejection.

The masking-prone round-3 fixtures also have teeth now. A fail-open common-dir command mutation fails because the test requires the `cannot verify isolation` invariant rather than accepting a later error. Removing the raw-segment guards fails five direct canonicalization witnesses, including the NUL crash-class case. Deriving alternate evidence from the request instead of the real clone makes the no-alternate fixture fail with no rejection.

The remaining union is loud as well: schema invocation and schema-SHA pin, reopened WorkspaceSpec hash, symlink-prefix inspection, round-2 cache fail-open behavior, branch desired-state validation, executable-file and Python-constraint venv weaknesses, receipt file fsync, parent-directory fsync, cleanup-before-receipt reorder, idempotent rerun removal, and deletion of project-source or structural-stage receipt evidence.

Independent contract receipt: the vendored schema is byte-identical to config main commit `4b36896` and hashes to `a5061501ba6651d7432d87d57f1c85902e5dec076f860a47faa299f5f590231c`.

Pristine verification:

- Targeted materialization and convergence: `106 passed`
- Full declared gr2 suite from `gr2/`: `614 passed` on Python 3.13
- Remote head re-read immediately before this review and still equals the reviewed commit

CI record: the required aggregate `CI`, both gr2 Python jobs, macOS, Ubuntu, integration, format, and clippy are green. A non-required Windows Rust matrix leg is red at this head with `685 passed, 1 failed` in unchanged `src/core/gripspace.rs`, specifically `test_update_gripspace_advances_configured_remote_branch` rejecting a Windows temp path as a gripspace name. That discrepancy is outside this PR's diff and does not block the S4 mechanism verdict.

No binding findings remain. The round-3 executor and its tests now satisfy the config#492 contract at the seams this review locked.

laynepenney (commented):
SENTINEL R2 FINAL RE-GATE CORRECTION: BINDING REQUEST CHANGES at exact `82a3a232322dee729e67ebe5352575be376a1052`.

This explicitly withdraws and supersedes my BINDING APPROVE in review `4784154143`.

Process record: I completed and posted my independent review before opening Atlas's exact-head verdict. That independent pass correctly proved the published round-3 mutation union loud, but Atlas found three additional production-shaped seams outside that union. I then reproduced all three from scratch in my exact-head worktree. The result was `3 failed`, each because the required guard did not fire.

1. **Existing clone plus declared reference cache still permits no alternate.** I created a valid canonical mirror cache, then created an ordinary existing clone with no alternates file. Applying a plan that declared the cache succeeded and returned `alternate_approved=false`. The existing-destination path returns from `_validate_clone_isolation` when the alternates file is absent, while the missing-alternate rejection only executes after a fresh clone. Required object-sharing state is therefore not validated on every apply.

2. **Idempotent cleanup deletes a newly tampered staged source without reopening or hashing it.** I applied a valid project file, recreated the staged path with different bytes, and reran the same plan. The rerun did not raise and deleted the tampered source. The destination-correct fast path schedules `_pending_unlink` based on existence alone. It must require the staged source to remain a regular non-symlink file whose reopened bytes match `source_sha256`, then preserve it on mismatch.

3. **Editable-install receipt evidence is not bound to the requested source.** I installed editable `alpha` and then editable `beta` into the same real venv in one plan. The second operation returned `distribution=\"alpha\"` and alpha's first-sorted `direct_url.json`. `_find_direct_url_evidence` must decode candidates and select the row whose resolved `url` equals the requested source and whose `dir_info.editable` is true before deriving distribution evidence.

The previously reported positive evidence still stands: the vendored schema and published 21-probe union are clean, targeted tests are `106 passed`, and the pristine declared suite is `614 passed`. Those facts do not close these three uncovered behaviors.

Required closure is exactly the three production-shaped fixtures above, each mutation-loud at its own invariant. Hold S4 until a revised head makes all three RED under the current code's corresponding weakening, then GREEN pristine.

laynepenney (commented):
ATLAS spec-author r2, final re-verification at exact `bb085d5c01eb5b1f4ccae24f95ea0923d428e97b`: **BINDING APPROVE**.

I re-ran the three original production-shaped probes before reading the aggregate count. All three now pass, and the closures match §6.2.1 rather than merely satisfying the fixtures:

1. Existing and freshly staged clones both enter `_validate_clone_isolation`; a declared `reference_base` with no observed alternate fails before either path can return or publish.
2. First-run projection and rerun cleanup both enter `_verify_staged_source`; the rerun cannot delete a recreated artifact until regular-file and reopened-byte hash checks succeed.
3. Editable evidence is selected by this operation's resolved source URL plus `dir_info.editable == true`; the second install receives beta's row rather than the first sorted alpha row.

The parity layer is real. `TestStateMatrixParity` exercises fresh/existing, first-run/rerun, and per-operation single/multi state with specific invariant messages and artifact-scoped receipt assertions. The code has one shared invariant seam on each divergent path, so the tests are checking a structurally unified path rather than two matching copies. The pinned config#492 schema loader and all §6.2.1 runtime invariants remain in the same execution path.

Independent mutation fruit:

- removed required-alternate enforcement: the original existing-clone probe plus both fresh/existing parity tests went RED, 3/3;
- removed rerun staged-source verification: the original tamper probe plus rerun hash and non-regular parity tests went RED, 3/3;
- removed source binding from PEP 610 lookup: the two-editable probe went RED on `alpha != beta`, 1/1.

Restored head: focused round-4 suite 12/12 green. Full suite: **626 passed on Python 3.11** and **626 passed on Python 3.13**. Worktree clean after mutation restoration.

My prior hold is closed. No residual spec-fidelity finding.

laynepenney (commented):
SENTINEL R2 ROUND-4 FINAL RE-GATE: BINDING REQUEST CHANGES at exact `bb085d5c01eb5b1f4ccae24f95ea0923d428e97b`.

The three reported round-4 reverts are loud, and the prior regression union remains loud. I executed 23 one-seam mutations independently. Each named prior or round-4 revert turned RED for its intended invariant, then GREEN restored. Pristine declared suite: `626 passed in 85.58s` on Python 3.13.

Two current-code contract survivors remain:

1. **P1, declared reference with empty alternates file fails open.** Production-shaped existing clone, valid declared mirror cache, and an existing zero-byte `.git/objects/info/alternates`: apply succeeds and receipts `observed_alternate=None` / `alternate_approved=false`. The required-alternate check distinguishes only missing file from present file, not zero usable alternate rows. A declared reference must require exactly the approved usable row on both fresh and existing paths.

2. **P1, verified bytes are not the bytes copied.** `_verify_staged_source` hashes one `read_bytes()`, returns `None`, then `_apply_project_file_operation` performs a second `read_bytes()` for publication. A deterministic source change between those calls produces a durable receipt carrying the expected hash while the destination contains different bytes. Return the verified bytes from the shared helper and write those exact bytes. Pin the two-read mutant.

The per-artifact PEP 610 test substrate also has binding silent seams:

3. Removing `dir_info.editable is True` leaves all `109` materialization tests GREEN.
4. Disabling the `file:` scheme guard leaves all `109` GREEN.
5. Removing percent-decoding leaves all `109` GREEN, despite schema-valid workspace paths permitting spaces.
6. Reverting to first-sorted PEP 610 evidence kills the targeted round-4 test, but `TestStateMatrixParity` remains `9/9` GREEN. The PR-body claim that this corner dies to both its target and parity partner is therefore false. Add the editable-install side of the single/multi-unit receipt matrix or narrow the claim.

Required closure: executable fixtures for empty/blank alternates, verified-read-equals-copied-bytes, non-editable matching PEP 610 rejection, non-file matching-path rejection, encoded local path acceptance, and editable per-operation parity. Re-run each named mutant separately and assert the intended invariant, not only exception type.

Worktree was restored clean after every mutation. Do not merge at this head.

PR #799, opened 2026-07-27
feat: S4-A materialization plan contract (schema, identity, receipt)

laynepenney:
First sub-PR of the **grip#797 split**. #797 reached five re-gates on a ~900-line monolith and kept finding corners; per Stromus's hard line it splits into four rooms small enough for exhaustive sweeps. This is the base.

**Lineage:** supersedes the plan-contract portion of #797 (closed as superseded-by-split). Siblings to follow: S4-B (clone/cache/alternates), S4-C (staging/project_file), S4-D (venv/editable + PEP 610).

## What this ships — and what it deliberately does not

Ships the **plan-level contract** for config#492's neutral MaterializationPlan v1. Ships **no operation execution at all**.

That was the carve decision, and it's worth stating explicitly because the alternative was tempting: carve the executor with weakened handlers and harden them in B/C/D. That would have put known-weak clone/staging/venv code on a shipped path. Instead every guarantee here is enforced *before* any handler could run, so this room is complete as shipped, and B/C/D each bring a handler **together with** its domain validation and mutation set.

Safety note, verified rather than assumed: `apply_materialization_plan` is not reachable from any CLI command — only the legacy `apply_plan` (workspace_spec.toml) is wired to `workspace materialize`. Landing the contract ahead of the handlers exposes no product path.

## Contents

**Pinned schema, fail-closed.** config#492's v1 schema (`4b36896`, SHA `a5061501…590231c`) is vendored *inside* the package, validated with `Draft202012Validator`, and its bytes SHA-verified at load. A tampered or unpinned schema refuses to validate at all rather than silently enforcing a different contract.

`schema_version=True` is worth naming: `bool` is an `int` subclass in Python, so `True == 1` and a hand-written `!= 1` check *accepts* it. JSON Schema's `const: 1` distinguishes the types — a concrete case where a hand validator wasn't merely looser, it was wrong.

**Identity-freedom by allowlist.** Exact per-kind field allowlists reject anything not explicitly permitted, by construction — a blacklist only catches names someone thought to enumerate (`metadata.display_name` sailed through one). Recursive identity-key rejection remains beneath it as defence in depth.

**Opaque tokens.** `plan_id` and `unit_key` are validated as path-safe tokens; gr2 checks their shape without deriving identity from them.

**§6.2.1 #1 — WorkspaceSpec binding.** The canonical spec is reopened and its bytes hashed against `workspace_spec_sha256`. Carrying the field is not verifying it.

**§6.2.1 #2 — canonicalization.** Raw-segment scanning (`Path()` normalizes `.` away before `.parts` can see it) plus a per-component `lstat` walk. A `resolve()`-based containment check *structurally cannot* catch a symlinked prefix: both the candidate and the root resolve through the same link, so containment holds while the bytes live outside.

**§6.2.1 #3 — collisions.** Destinations compared canonically and case-folded: `u1/.venv` vs `u1/./.venv`, and `straße` vs `STRASSE` (casefold, not `lower` — only casefold folds ß→ss).

**§6.2.1 #9/#10 — receipt.** The exact pinned canonical hash (`sort_keys`, `separators=(",",":")`, `ensure_ascii=False`), published via same-dir temp → flush → `fsync` → atomic replace → parent-dir `fsync`. Rename success alone is not durable acknowledgement.

**grip#539.** `converge_unit_repos` is scheduled on the first planning pass rather than only when a unit root and `unit.toml` already exist. The existing test asserted the bug *as correct* (`assertNotIn` the fix's own target operation) and is inverted here; a second fixture encoded the same bug incidentally — its unit's own checkout was silently never scheduled — and needed the same correction.

**`gr2.overlay` alias** (config#491 gap#13), additive: same physical directory, two import names, existing `gr2_overlay` consumers untouched.

## Testing discipline

Carried from #797's review cycle, where the recurring defect was tests passing *for the wrong reason*:

- **Assert which invariant fired**, not that *some* `MaterializationPlanError` was raised. A fixture broken enough to trip the invariant under test is usually broken enough to trip its neighbours, and a type-only `assertRaises` cannot tell them apart.
- **Canonicalization is tested at its own level.** The pinned schema is a strictly stronger *first* gate, so plan-level tests never reach the canonicalizer — testing it only through plans would leave it entirely shadowed and unverified. It is genuinely load-bearing: the legacy adapter and the coming B/C/D handlers call it *without* schema validation.
- Positive faces included alongside rejections, so the guards can't pass by rejecting everything.

## Verification

- **561 passed** on Python 3.11.9 (the declared floor) and 3.13.
- Fresh no-cache install both times; the packaged-schema test asserts the schema ships *inside the installed package* and matches the pinned SHA, not merely that it exists in the source tree.

## Closure round — Atlas's three A-owned blockers

All three were one shape: **the guards existed but were not bound at the contract boundaries.** No scope leakage into B/C/D.

**P1 — the terminal receipt was bound to nothing.** `write_materialization_receipt` accepted the raw live plan and an arbitrary result list. Atlas's fruit: `plan_id="../../escaped"` published `.grip/escaped.json` outside the receipt directory; a one-operation plan with `op_results=[]` published `stage="MATERIALIZED"` with no evidence at all; and `{"secret": "TOKEN", "nested": {"memory_body": "PRIVATE"}}` was serialized verbatim.

`validate_materialization_plan` now returns a **`ValidatedPlan` capability** that only it can mint, and publication requires one. Validation becomes something the publisher *holds* rather than something a caller is trusted to have remembered to do. Evidence is screened with the same discipline as the plan: cardinality, order, and kind must match the validated operations before anything may claim `MATERIALIZED`, and forbidden fields are rejected recursively.

Atlas's framing of why that last piece matters is worth preserving verbatim in the code, and I did: the plan's operations are protected by a *closed* schema whose per-kind allowlists permit no nested object carrier — so recursive rejection there has little left to find. The **result graph is open, and it is what gets persisted.** That, not the plan, is the actual smuggling boundary.

**P2 — §6.2.1 #2 applies to the derived paths too.** All three probes succeeded before this closure:
- a symlinked `.grip/workspace_spec.toml` pointing outside the team root was accepted whenever its bytes hashed to the declared value — the hash confirms *content* and says nothing about *where it was read from*;
- a symlinked `.grip/state/materialization` published the terminal receipt outside the root entirely;
- the predictable `mp_<id>.json.tmp-<pid>` name opened with plain `open()` followed a pre-created symlink, overwrote the external target, then published that symlink as the final receipt.

Now the spec is read through the canonicalizer and must be a regular non-symlink file, the receipt directory must be a real in-root directory, and the temp is created `O_EXCL|O_NOFOLLOW`.

**P3 — the durability order was mutation-silent.** Counting two `fsync` calls does not pin a *sequence*: Atlas moved the parent-directory fsync before `os.replace` and all five publication tests stayed green. The test now asserts the exact ordered chain **with fd roles** — `fsync(file)` → `replace` → `fsync(dir)` — distinguishing file from directory via `fstat`, so the reorder mutant dies for *this* invariant rather than through an unrelated failure.

### Closure verification

**11/11 mutants killed, zero survivors.** Isolated worktrees, one mutant each, stale-worktree sanity-guarded:

- Atlas's exact P3 reorder → dies to `test_publication_order_is_exact` (the test that was previously blind to it)
- temp-file fsync removed → RED
- capability check removed, and token check removed (forgeable capability) → each RED
- evidence screening removed, and each sub-guard removed independently (cardinality, kind/order, recursive rejection) → each RED
- each P2 hardening reverted independently (raw spec read, raw receipt dir, plain `open()` temp) → each RED

**572 passed** on Python 3.11.9 and 3.13.

## Premium boundary

OSS (grip is MIT) — neutral plan validation and receipt plumbing only. No identity, org, channel, or agent semantics; enforced by the schema's closed `additionalProperties: false` plus per-kind allowlists and recursive identity rejection, not asserted in prose.

---

## Closure round 2 — Sentinel r2 (review 4784681890) @ `c050bdf`

All seven findings closed. Sentinel reviewed at `f6b9e8d`; two of his findings were
partly addressed by round 1 (`2f1ab10`), and I verified against his exact probes rather
than assuming, per the note in his re-request.

### Production changes

| Finding | Change |
|---|---|
| 7 | Collision keys NFC-normalize **before** casefolding. Normalization and case folding are separate aliasing axes: `units/café/.venv` in NFC vs NFD casefold to distinct strings yet name one destination on a normalization-insensitive filesystem. |
| 3 | A parent-directory `fsync` failure now **unlinks the published receipt**. Durability is a failure-path contract, not only an ordering one — a publication that cannot be made durable must not remain published, or a caller doing destructive cleanup on the strength of a receipt acts on one that may not survive a crash. |
| 3 | A failed `os.replace` now unlinks the temp file. Found by this closure's own residue test rather than reasoned about. |

### Test closures

| Finding | Closure |
|---|---|
| 1 | Receipt publication through a symlinked `.grip/state` and through a pre-created temp symlink, each asserting **no outside bytes change**, plus the positive face: the published receipt is a regular in-root file. |
| 2 | Every declared forbidden key rejected in receipt evidence, asserting *which* invariant fired (the cardinality and kind checks run first and would otherwise reject a malformed probe for an unrelated reason). |
| 3 | Ordered publication oracle now observes `flush`, not only `fsync`/`replace`, with fd roles resolved via `fstat`. Deleting `fh.flush()` is the subtle mutant: `close()` flushes anyway, so the published bytes are identical while `fsync` syncs an empty file. No content assertion can see that. Failure injected at **every** durability step, not only the first. |
| 3b | Whole-value receipt binding, reconstructed independently, replacing a key-set plus four sampled leaves. |
| 4 | Per-field symlink matrix proving all **seven** operation paths reach `canonicalize_workspace_path`, plus destination-kind and cross-kind collision fixtures proving clone and `project_file` destinations participate in collision accounting. |
| 5 | Plan hash pins operation **array order** against a sorting mutant, and matches independently computed canonical bytes. |
| 6 | Recursive identity rejection pinned as a production seam beneath a permissive-validator double (a direct helper call proves nothing about wiring), plus a **literal inventory pin** — see below. |

### Two things I found while closing, not asked for

**A self-referential table test.** `test_every_declared_forbidden_key_is_enforced` iterated
the live `_FORBIDDEN_IDENTITY_KEYS` set, so deleting a member just made the loop shorter and
stayed green. Sentinel's probe caught the `secret` removal only because one *other* test
hardcodes that single key — meaning the other thirteen members were unguarded. A literal
inventory pin closes the whole class.

**A misplaced `if __name__ == "__main__"` guard** sat mid-file, ahead of the six closure
classes, so running the file directly executed `unittest.main()` before those classes were
defined and silently skipped them. Moved to the end.

### Mutation verification

**21/21 killed**, and each mutant now declares the test that must die, so a
right-answer-for-the-wrong-reason kill fails the sweep. That assertion earned its keep:
it caught a hazard in the sweep *methodology* — CPython invalidates cached bytecode on
`(source mtime in whole seconds, source size)`, and rewriting the source several times per
second means two equal-sized mutants inside one second silently re-import the previous
mutant's `.pyc`. Two rows were being attributed to the wrong mutant. Purging `__pycache__`
and setting `PYTHONDONTWRITEBYTECODE=1` per run fixed it. Counting kills alone would not
have surfaced this.

Mutants: `flush-deleted`, `parent-fsync-before-replace`, `fsync-final-file-twice-not-parent`,
`operations-empty`, `constant-plan-id`, the seven `*-raw` canonicalizer-bypass mutants,
`clone-dest-returns-none`, `project_file-dest-returns-none`, `hash-sorts-operations`,
`identity-call-removed`, `secret-removed-from-inventory`, `nfc-normalization-removed`,
`casefold-removed`, `receipt-dir-not-canonicalized`, `temp-open-follows-symlink`.

**594 passed** on Python 3.11.15 and 3.13.2.

### CI note

Required checks pass at `c050bdf`, including both `gr2 Python Tests` legs. `Test
(windows-latest)` is red — it is red at `2f1ab10` and at `f6b9e8d` too, i.e. from this
branch's first commit, and this diff is Python-only against a `cargo test` job. Tracked as
grip#798, not introduced here. (Correcting my own earlier shorthand: I described `2f1ab10`
as "CI green" when what was green was the required aggregate, not the Windows leg.)

---

## Residual closed — Atlas final re-gate @ `2f80258`

`ValidatedPlan` was `frozen=True`, but dataclass freezing freezes the field **binding**,
not the graph the field points at. `plan` aliased the caller's live dict and the writer
recomputed `plan_hash` from it at publication time, so mutating the plan after validation
published a receipt attesting to a graph that was never validated. Atlas's exact repro is
the RED test, verbatim.

Three captures close it, each defending a different alias:

| Alias | Capture |
|---|---|
| The caller's dict | `deepcopy` on **entry**, so the original is never consulted again. Taking it on entry rather than at mint also closes the window where a mutation lands between the checks and the capture. |
| The representation the capability exposes | Deep freeze — `MappingProxyType` for mappings, tuples for sequences, closing both item-assignment and append. A shallow copy is insufficient (Atlas); so is a shallow freeze. |
| The hash itself | `plan_hash` computed **once** at validation and stored. Publication reads captured facts and no longer re-derives anything from a mutable graph. |

`compute_plan_hash` gained a `default` type-adapter so the freeze does not turn the public
helper into a trap for the B/C/D handlers that will hold these plans. It fires only for
objects json cannot serialize natively, so canonical bytes for a plain JSON graph are
byte-identical — the pinned recipe is unchanged, and the existing canonical-bytes tests
still pass unmodified.

### One finding worth recording

Deep-freezing at mint **masks** a missing snapshot in any single-threaded probe, because
the freeze itself builds fresh objects. The `snapshot-removed` and `snapshot-shallow-only`
mutants survived every other test in the new class — including the identity assertions that
look like they cover exactly this. What the freeze cannot defend is the mid-validation
window, so that test injects a mutation through `_read_canonical_workspace_spec_bytes`,
which runs after the schema check and before the operation loop, reproducing the race
deterministically without threads. Both mutants then die.

That is the same shape as the round-2 self-referential table test: a guard that appears
covered because a *stronger adjacent guard* keeps the probe green.

**600 passed** on Python 3.11.15 and 3.13.2. **Mutation sweep 26/26 killed**, each by its
declared target test.

---

## Second seam on the same object — Sentinel final r2 @ `e57f163`

Both reviewers converged on one design truth: **the capability must bind immutable content,
not object identity or a copyable field.**

Sentinel's witness, reproduced verbatim before the fix — `dataclasses.replace` re-invokes
`__init__` with the existing field values, so the real token rode into a modified shell:

```
replace succeeded, plan_id = ../../escaped
PUBLISHED AT: .../materialization/../../escaped.json
```

That is a receipt published outside the receipt directory by an object nobody validated.

### Three layers, each pinned at its own level

| Layer | Binds |
|---|---|
| `plan_hash` describes the snapshot | swapping the graph cannot survive, with or without a matching hash |
| every published fact agrees with the snapshot | `plan_id`, `unit_key`, `schema_version`, `workspace_spec_sha256`, `operation_kinds` cannot diverge from the graph they describe |
| HMAC seal over all of the above | a hand-built shell that is internally self-consistent still cannot pass |

Verification runs at every **use**, not only at construction: `frozen=True` blocks
`__setattr__` but not `object.__setattr__`, so an honestly-minted capability can still be
edited afterwards, and every path in the publisher reads these fields — including the one
that builds the receipt filename. The secret is process-local and never persisted; this is
an in-process capability, not a credential.

The layers **shadow each other**, so each test names the layer that must reject it.
Asserting only "some capability error" would let a removed layer hide behind the one
beneath it — the same shadowing Sentinel flagged for the pinned schema. The seal has no
layer above it, so `test_the_seal_itself_rejects_a_self_consistent_forgery` is what keeps
it from being dead code.

Sentinel's witness 2 (mutating the caller's plan after validation) was already closed by
the preceding snapshot commit — verified by running it, not assumed.

### Mutation

**30/31 killed**, each by its declared target test. One survivor, `seal-omits-plan_id`, is
**equivalent**, and shown to be so by attempted exploitation rather than by argument: with
the mutant applied, both replacing `plan_id` and forging a self-consistent shell with the
genuine seal still fail, because the seal covers `plan_hash` and `plan_id` is derivable
from the snapshot that hash binds. Reported as equivalent rather than counted as a kill,
matching the standard Sentinel applied to his own schema-shadowed edits.

**606 passed** on Python 3.11.15 and 3.13.2.

---

## Provenance layer — Atlas re-gate @ `30693ad`

Atlas's re-gate (review 4785725633) was against `2f80258`, where the token was still an
identity sentinel. **Verified rather than assumed**: every forge vector he named —
`plan_id`, `unit_key`, `workspace_spec_sha256`, `plan_hash`, `operation_kinds` — is already
blocked by the content seal at `e57f163`, with no receipt reaching the filesystem. His
repros are now tests (`test_atlas_forge_family_cannot_publish`).

One vector was genuinely still open, and it is the one his closure names in words:
*"copying/replacing the capability cannot preserve authority."* A copy changes no field, so
it carries a genuinely valid seal and passes every content check. Only **origin** can refuse
it.

### Four layers, orthogonal by construction

| Layer | Binds | Closes |
|---|---|---|
| hash-describes-snapshot | the graph | swapped snapshots |
| facts-agree-with-snapshot | the captures | `replace` of any published fact |
| HMAC seal | content as a whole | hand-built self-consistent shells |
| **validator provenance** (WeakSet of minted instances) | **origin** | **copies that change nothing** |

The seal and provenance are orthogonal on purpose: content vs origin. Neither shadows the
other, which is what makes this defense in depth rather than one guard hiding behind
another — and each has a test at its own level.

Provenance is checked at **use** only, never at construction: registration happens after
`__init__` returns, so the instance is not yet a member when `__post_init__` runs.
`eq=False` keeps identity hashing (a `MappingProxyType` field is unhashable anyway, and two
capabilities over equal facts *should* be distinct capabilities).

Recorded rather than smoothed over: `deepcopy` of a capability raises `TypeError` because
the frozen snapshot is not picklable. Same guarantee, different mechanism, so it is
asserted separately — a shared assertion would hide which one actually holds.

**609 passed** on Python 3.11.15 and 3.13.2. **Mutation sweep 33/34 killed**, each by its
declared target. The lone survivor, `seal-omits-plan_id`, remains equivalent, documented,
and demonstrated by attempted exploitation rather than argument.

---

## Use-time TOCTOU closed — Atlas @ `5fd3ab4`

It survived two heads, which was the signal that **the model was wrong rather than a guard
missing**. I had been hardening the capability so it could not be *forged*, while the writer
went on reading the live shell after verification, across a caller-controlled callback.

The content seal is what made this look covered: it gates **content**, so tampering cannot
change the sealed facts — but nothing gated the **reads**. That is the adjacent-strong-guard
masking class again, this time in the fix rather than in the test.

### Closure

`ValidatedPlan.consume()` verifies once and returns an immutable `_ConsumptionBinding`.
The receipt body, both filenames, and the evidence screen consume only that; nothing reads
the capability again.

Two details carry it:

- Facts come from the **frozen snapshot**, not the shell fields, so a post-verify edit is
  irrelevant rather than merely detected.
- The **order** is load-bearing: facts are captured before the evidence is materialized,
  because materializing is what runs caller code.

There is deliberately **no second seal check** after the callback. Once everything derives
from the frozen snapshot it would be unreachable, and an unreachable guard is worse than
none — it reads as protection while being untestable at its own level.

### Found while closing, not in the witness

The evidence was walked **twice** — once by the screen, once by json serialization. A list
returning different contents per iteration screened clean and then persisted `secret` into
the receipt:

```text
evidence double-read -> iterations: 2 | LEAKED persisted: True
```

That is a **premium-boundary** failure, not only a correctness one: the receipt's whole
claim is that it carries no identity-bearing content. Evidence is now materialized once
into plain JSON types, and the screen inspects the same value that gets persisted.

Also recorded: the temp-filename mutant **survived a directory assertion**, because the
rename *consumes* the temp file — a post-hoc listing looks identical whether or not the name
was attacker-influenced. Only an oracle observing the name at `os.replace` can see it.

### Layers Atlas passed, re-checked untouched

`copy.copy` still rejected by provenance; `dataclasses.replace` of `plan_id`, `plan_hash`
and `operation_kinds` still rejected at the named layer; no `.grip/escaped.json`.

**613 passed** on Python 3.11.15 and 3.13.2. **Mutation sweep 37/38 killed**, each by its
declared target. The lone survivor, `seal-omits-plan_id`, remains the documented equivalent
one.

Refs #797

### DIFF
--- a/gr2/gr2/schemas/gr2-materialization-plan-v1.schema.json
+++ b/gr2/gr2/schemas/gr2-materialization-plan-v1.schema.json
@@ -1,0 +1,191 @@
+{
+  "$schema": "https://json-schema.org/draft/2020-12/schema",
+  "$id": "urn:synapt:gr2:materialization-plan:v1",
+  "title": "gr2 MaterializationPlan v1",
+  "description": "Identity-free, one-opaque-unit materialization plan consumed by the neutral gr2 executor.",
+  "type": "object",
+  "additionalProperties": false,
+  "required": [
+    "schema_version",
+    "plan_id",
+    "unit_key",
+    "workspace_spec_sha256",
+    "operations"
+  ],
+  "properties": {
+    "schema_version": {
+      "const": 1
+    },
+    "plan_id": {
+      "$ref": "#/$defs/opaqueToken"
+    },
+    "unit_key": {
+      "$ref": "#/$defs/opaqueToken"
+    },
+    "workspace_spec_sha256": {
+      "$ref": "#/$defs/sha256"
+    },
+    "operations": {
+      "type": "array",
+      "minItems": 1,
+      "items": {
+        "oneOf": [
+          {
+            "$ref": "#/$defs/cloneOperation"
+          },
+          {
+            "$ref": "#/$defs/venvOperation"
+          },
+          {
+            "$ref": "#/$defs/editableInstallOperation"
+          },
+          {
+            "$ref": "#/$defs/projectFileOperation"
+          }
+        ]
+      }
+    }
+  },
+  "$defs": {
+    "opaqueToken": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 128,
+      "pattern": "^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$"
+    },
+    "sha256": {
+      "type": "string",
+      "pattern": "^[0-9a-f]{64}$"
+    },
+    "nonEmptyString": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 2048,
+      "pattern": "^[^\\u0000]+$"
+    },
+    "workspaceRelativePath": {
+      "type": "string",
+      "minLength": 1,
+      "maxLength": 1024,
+      "pattern": "^(?![~/])(?!.*\\\\)(?!.*//)(?!.*(?:^|/)\\.{1,2}(?:/|$))[^\\u0000]+$"
+    },
+    "cacheReferencePath": {
+      "type": "string",
+      "pattern": "^\\.grip/cache/repos/[A-Za-z0-9][A-Za-z0-9._-]{0,127}\\.git$"
+    },
+    "stagedInputPath": {
+      "type": "string",
+      "pattern": "^\\.grip/staging/inputs/[A-Za-z0-9][A-Za-z0-9._-]{0,127}$"
+    },
+    "cloneOperation": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "kind",
+        "repo_url",
+        "dest_path",
+        "branch"
+      ],
+      "properties": {
+        "kind": {
+          "const": "clone"
+        },
+        "repo_url": {
+          "$ref": "#/$defs/nonEmptyString"
+        },
+        "dest_path": {
+          "$ref": "#/$defs/workspaceRelativePath"
+        },
+        "branch": {
+          "$ref": "#/$defs/nonEmptyString"
+        },
+        "reference_base": {
+          "$ref": "#/$defs/cacheReferencePath"
+        }
+      }
+    },
+    "venvOperation": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "kind",
+        "dest_path",
+        "engine",
+        "python"
+      ],
+      "properties": {
+        "kind": {
+          "const": "venv"
+        },
+        "dest_path": {
+          "$ref": "#/$defs/workspaceRelativePath"
+        },
+        "engine": {
+          "const": "uv"
+        },
+        "python": {
+          "$ref": "#/$defs/nonEmptyString"
+        }
+      }
+    },
+    "editableInstallOperation": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "kind",
+        "venv_path",
+        "source_path",
+        "extras"
+      ],
+      "properties": {
+        "kind": {
+          "const": "editable_install"
+        },
+        "venv_path": {
+          "$ref": "#/$defs/workspaceRelativePath"
+        },
+        "source_path": {
+          "$ref": "#/$defs/workspaceRelativePath"
+        },
+        "extras": {
+          "type": "array",
+          "uniqueItems": true,
+          "items": {
+            "type": "string",
+            "minLength": 1,
+            "maxLength": 128,
+            "pattern": "^[A-Za-z0-9][A-Za-z0-9._-]{0,127}$"
+          }
+        }
+      }
+    },
+    "projectFileOperation": {
+      "type": "object",
+      "additionalProperties": false,
+      "required": [
+        "kind",
+        "source_path",
+        "source_sha256",
+        "dest_path",
+        "mode"
+      ],
+      "properties": {
+        "kind": {
+          "const": "project_file"
+        },
+        "source_path": {
+          "$ref": "#/$defs/stagedInputPath"
+        },
+        "source_sha256": {
+          "$ref": "#/$defs/sha256"
+        },
+        "dest_path": {
+          "$ref": "#/$defs/workspaceRelativePath"
+        },
+        "mode": {
+          "const": "copy"
+        }
+      }
+    }
+  }
+}
--- a/gr2/pyproject.toml
+++ b/gr2/pyproject.toml
@@ -12,6 +12,10 @@ dependencies = [
     "tomli_w>=1.0",
     "pyyaml>=6.0",
     "toml>=0.10",
+    # Draft 2020-12 validation of the pinned MaterializationPlan v1 schema
+    # (config#492). 4.18 is the first release with the modern referencing
+    # stack; anything older silently falls back to a legacy resolver.
+    "jsonschema>=4.18",
 ]
 
 [project.scripts]
@@ -24,12 +28,23 @@ dev = [
 ]
 
 [tool.setuptools]
-packages = ["gr2_overlay", "gr2", "gr2.python_cli", "gr2.prototypes"]
+# gr2_overlay is a time-bounded compat alias for gr2.overlay (config#491
+# gap#13): same physical directory, two import names. New code imports
+# gr2.overlay; gr2_overlay stays for existing consumers until they are
+# migrated separately.
+packages = ["gr2_overlay", "gr2", "gr2.python_cli", "gr2.prototypes", "gr2.overlay"]
 
 [tool.setuptools.package-dir]
 gr2 = "gr2"
 "gr2.python_cli" = "python_cli"
 "gr2.prototypes" = "prototypes"
+"gr2.overlay" = "gr2_overlay"
+
+[tool.setuptools.package-data]
+# The pinned MaterializationPlan v1 schema (config#492) ships inside the
+# gr2 package -- spec_apply verifies its bytes against the pinned SHA-256
+# at load and fails closed on any mismatch.
+gr2 = ["schemas/*.json"]
 
 [tool.pytest.ini_options]
 testpaths = ["tests", "gr2_overlay/tests"]
--- a/gr2/python_cli/spec_apply.py
+++ b/gr2/python_cli/spec_apply.py
@@ -7,6 +7,7 @@ import json
 import os
 import stat
 import tomllib
+import unicodedata
 from datetime import UTC, datetime
 from pathlib import Path
 
@@ -820,7 +821,13 @@ def validate_materialization_plan(workspace_root: Path, plan: dict[str, object])
         _reject_identity_fields_recursive(op, path=f"operations[{idx}]")
         canonical_dest = _validate_operation_shape(op, idx=idx, workspace_root=workspace_root)
         if canonical_dest is not None:
-            key = str(canonical_dest).casefold()
+            # NFC-normalize BEFORE casefolding (Sentinel finding 7):
+            # "units/café/.venv" spelled NFC vs NFD are distinct Python
+            # strings that casefold to distinct values, yet on a
+            # normalization-insensitive filesystem they name ONE
+            # destination. Normalization and case folding are separate
+            # aliasing axes; collision detection has to close both.
+            key = unicodedata.normalize("NFC", str(canonical_dest)).casefold()
             if key in seen_canonical_dests:
                 raise MaterializationPlanError(
                     f"operations[{idx}] dest_path collides (case-folded/normalized) with "
@@ -997,10 +1004,28 @@ def write_materialization_receipt(
     except BaseException:
         tmp_path.unlink(missing_ok=True)
         raise
-    os.replace(tmp_path, receipt_path)
-    dir_fd = os.open(receipt_dir, os.O_RDONLY)
     try:
-        os.fsync(dir_fd)
-    finally:
-        os.close(dir_fd)
+        os.replace(tmp_path, receipt_path)
+    except BaseException:
+        # A failed replace leaves the temp behind otherwise -- found by this
+        # closure's own residue test rather than reasoned about.
+        tmp_path.unlink(missing_ok=True)
+        raise
+
+    # Sentinel finding 3: durability is a FAILURE-PATH contract, not only an
+    # ordering one. If the parent-directory fsync fails, the rename may not
+    # survive a crash -- yet the receipt is already visible at its published
+    # path, so a caller that treats "the writer returned" or "a receipt
+    # exists" as durable acknowledgement would proceed to destructive
+    # cleanup on the strength of a receipt that could vanish. A publication
+    # that cannot be made durable must not remain published.
+    try:
+        dir_fd = os.open(receipt_dir, os.O_RDONLY)
+        try:
+            os.fsync(dir_fd)
+        finally:
+            os.close(dir_fd)
+    except BaseException:
+        receipt_path.unlink(missing_ok=True)
+        raise
     return receipt_path
--- a/gr2/python_cli/spec_apply.py
+++ b/gr2/python_cli/spec_apply.py
@@ -11,6 +11,7 @@ import secrets
 import stat
 import tomllib
 import unicodedata
+import weakref
 from datetime import UTC, datetime
 from pathlib import Path
 from types import MappingProxyType
@@ -541,7 +542,22 @@ def _deep_freeze(value: object) -> object:
     return value
 
 
-@dataclasses.dataclass(frozen=True)
+# Provenance registry: the instances this validator actually minted.
+#
+# Orthogonal to the seal, and deliberately so. The seal binds CONTENT, which
+# closes replace() and in-place edits; provenance binds ORIGIN, which closes
+# copying that preserves authority without changing any field. They fail in
+# different ways, which is what makes them defense in depth rather than one
+# guard shadowing another.
+#
+# Weak, so holding a capability never keeps it alive; entries vanish with the
+# object. eq=False on the dataclass keeps identity hashing (a MappingProxyType
+# field is unhashable anyway) -- and two capabilities over equal facts SHOULD
+# be distinct capabilities, which is exactly what identity semantics give.
+_MINTED_CAPABILITIES: weakref.WeakSet = weakref.WeakSet()
+
+
+@dataclasses.dataclass(frozen=True, eq=False)
 class ValidatedPlan:
     """Proof that a plan passed the full v1 contract, plus the facts a
     publisher needs to bind its evidence to that plan.
@@ -567,16 +583,22 @@ class ValidatedPlan:
     _seal: str = dataclasses.field(repr=False)
 
     def __post_init__(self) -> None:
+        # Provenance cannot be checked here: registration happens after
+        # construction, so this instance is not in the registry yet.
         self.verify()
 
-    def verify(self) -> None:
+    def verify(self, *, require_provenance: bool = False) -> None:
         """Re-derive the seal from the current contents and compare.
 
         Called at construction AND at every use. Construction-time checking
         alone is not enough: `frozen=True` blocks __setattr__, not
         object.__setattr__, so an already-minted capability can still be
         edited in place. Re-deriving at use means the fields a publisher
-        reads are provably the fields that were sealed."""
+        reads are provably the fields that were sealed.
+
+        `require_provenance` adds the origin check, which only makes sense at
+        USE. Copying a capability preserves every field and therefore every
+        content check; only origin can refuse it."""
         # The hash must actually describe the snapshot, so swapping the graph
         # (with or without a matching hash) cannot survive.
         if compute_plan_hash(self.plan) != self.plan_hash:
@@ -614,6 +636,11 @@ class ValidatedPlan:
             raise MaterializationPlanError(
                 "ValidatedPlan capability is invalid: it was not minted by "
                 "validate_materialization_plan for these exact facts"
+            )
+        if require_provenance and self not in _MINTED_CAPABILITIES:
+            raise MaterializationPlanError(
+                "ValidatedPlan capability is invalid: it is not one this validator "
+                "minted -- a copy preserves every field but not its provenance"
             )
 
 
@@ -970,11 +997,13 @@ def validate_materialization_plan(workspace_root: Path, plan: dict[str, object])
         "workspace_spec_sha256": workspace_spec_sha256,
         "operation_kinds": tuple(str(op["kind"]) for op in operations),
     }
-    return ValidatedPlan(
+    validated = ValidatedPlan(
         plan=_deep_freeze(plan),
         _seal=_capability_seal(**facts),
         **facts,
     )
+    _MINTED_CAPABILITIES.add(validated)
+    return validated
 
 
 _WORKSPACE_SPEC_RELATIVE = ".grip/workspace_spec.toml"
@@ -1120,7 +1149,7 @@ def write_materialization_receipt(
     # was minted honestly can still be edited afterwards -- frozen=True does
     # not stop object.__setattr__ -- and every path below reads these fields,
     # including the one that builds the receipt filename.
-    validated.verify()
+    validated.verify(require_provenance=True)
     _screen_receipt_evidence(validated, op_results)
 
     receipt = {
--- a/gr2/python_cli/spec_apply.py
+++ b/gr2/python_cli/spec_apply.py
@@ -3,9 +3,11 @@ from __future__ import annotations
 import copy
 import dataclasses
 import hashlib
+import hmac
 import importlib.resources
 import json
 import os
+import secrets
 import stat
 import tomllib
 import unicodedata
@@ -505,13 +507,23 @@ class MaterializationPlanError(Exception):
     pass
 
 
-# Capability token. A ValidatedPlan can only be minted by
+# Capability seal. A ValidatedPlan can only be minted by
 # validate_materialization_plan, so a receipt cannot be published from a
 # plan that was never validated -- Atlas P1: the writer previously accepted
 # the raw live plan and an arbitrary result list, which let a schema-invalid
 # plan_id escape the receipt directory and let an unvalidated result graph
 # be persisted verbatim.
-_VALIDATION_TOKEN = object()
+#
+# This was an opaque sentinel object, keyed on IDENTITY. Sentinel's witness:
+# dataclasses.replace() re-invokes __init__ with the existing field values,
+# so the real token rode into a modified shell and publication used the
+# altered plan_id -- writing .grip/escaped.json outside the receipt
+# directory. Identity is copyable; the capability has to bind CONTENT.
+#
+# Process-local and never persisted: this is an in-process capability, not a
+# credential. It cannot be recomputed by a caller who did not go through
+# validation, which is the whole point.
+_CAPABILITY_SECRET = secrets.token_bytes(32)
 
 
 def _deep_freeze(value: object) -> object:
@@ -552,13 +564,88 @@ class ValidatedPlan:
     workspace_spec_sha256: str
     operation_kinds: tuple[str, ...]
     plan_hash: str
-    _token: object = dataclasses.field(repr=False)
+    _seal: str = dataclasses.field(repr=False)
 
     def __post_init__(self) -> None:
-        if self._token is not _VALIDATION_TOKEN:
+        self.verify()
+
+    def verify(self) -> None:
+        """Re-derive the seal from the current contents and compare.
+
+        Called at construction AND at every use. Construction-time checking
+        alone is not enough: `frozen=True` blocks __setattr__, not
+        object.__setattr__, so an already-minted capability can still be
+        edited in place. Re-deriving at use means the fields a publisher
+        reads are provably the fields that were sealed."""
+        # The hash must actually describe the snapshot, so swapping the graph
+        # (with or without a matching hash) cannot survive.
+        if compute_plan_hash(self.plan) != self.plan_hash:
             raise MaterializationPlanError(
-                "ValidatedPlan may only be constructed by validate_materialization_plan"
+                "ValidatedPlan capability is invalid: plan_hash does not describe its plan snapshot"
             )
+        # ...and the snapshot must agree with every fact published beside it,
+        # so the two can never diverge into "sealed but inconsistent".
+        snapshot_facts = (
+            self.plan.get("plan_id"),
+            self.plan.get("unit_key"),
+            self.plan.get("schema_version"),
+            self.plan.get("workspace_spec_sha256"),
+            tuple(str(op.get("kind")) for op in self.plan.get("operations", ())),
+        )
+        if snapshot_facts != (
+            self.plan_id,
+            self.unit_key,
+            self.schema_version,
+            self.workspace_spec_sha256,
+            tuple(self.operation_kinds),
+        ):
+            raise MaterializationPlanError(
+                "ValidatedPlan capability is invalid: published facts disagree with the plan snapshot"
+            )
+        expected = _capability_seal(
+            plan_hash=self.plan_hash,
+            plan_id=self.plan_id,
+            unit_key=self.unit_key,
+            schema_version=self.schema_version,
+            workspace_spec_sha256=self.workspace_spec_sha256,
+            operation_kinds=self.operation_kinds,
+        )
+        if not hmac.compare_digest(str(self._seal), expected):
+            raise MaterializationPlanError(
+                "ValidatedPlan capability is invalid: it was not minted by "
+                "validate_materialization_plan for these exact facts"
+            )
+
+
+def _capability_seal(
+    *,
+    plan_hash: str,
+    plan_id: str,
+    unit_key: str,
+    schema_version: int,
+    workspace_spec_sha256: str,
+    operation_kinds: tuple[str, ...],
+) -> str:
+    """Bind the seal to CONTENT, not to object identity.
+
+    Every fact publication consumes is covered, so altering any one of them
+    -- by dataclasses.replace, by object.__setattr__, or by hand-building a
+    shell -- produces a seal that no longer matches. Takes values rather than
+    an instance so the mint can compute it before the object exists."""
+    payload = json.dumps(
+        [
+            plan_hash,
+            plan_id,
+            unit_key,
+            schema_version,
+            workspace_spec_sha256,
+            list(operation_kinds),
+        ],
+        sort_keys=True,
+        separators=(",", ":"),
+        ensure_ascii=False,
+    ).encode("utf-8")
+    return hmac.new(_CAPABILITY_SECRET, payload, hashlib.sha256).hexdigest()
 
 
 # The normative wire contract (config#492, merged 4b36896). A hand-rolled
@@ -875,15 +962,18 @@ def validate_materialization_plan(workspace_root: Path, plan: dict[str, object])
     # Hash the validated bytes ONCE, here, while they are still exactly what
     # passed the checks above. Recomputing at publication time is what let a
     # receipt attest to a post-validation mutation.
+    facts = {
+        "plan_hash": compute_plan_hash(plan),
+        "plan_id": str(plan["plan_id"]),
+        "unit_key": str(plan["unit_key"]),
+        "schema_version": int(plan["schema_version"]),
+        "workspace_spec_sha256": workspace_spec_sha256,
+        "operation_kinds": tuple(str(op["kind"]) for op in operations),
+    }
     return ValidatedPlan(
         plan=_deep_freeze(plan),
-        plan_id=str(plan["plan_id"]),
-        unit_key=str(plan["unit_key"]),
-        schema_version=int(plan["schema_version"]),
-        workspace_spec_sha256=workspace_spec_sha256,
-        operation_kinds=tuple(str(op["kind"]) for op in operations),
-        plan_hash=compute_plan_hash(plan),
-        _token=_VALIDATION_TOKEN,
+        _seal=_capability_seal(**facts),
+        **facts,
     )
 
 
@@ -1026,6 +1116,11 @@ def write_materialization_receipt(
             "write_materialization_receipt requires a ValidatedPlan from "
             "validate_materialization_plan, not a raw plan"
         )
+    # Re-derive the seal HERE, not just at construction. A capability that
+    # was minted honestly can still be edited afterwards -- frozen=True does
+    # not stop object.__setattr__ -- and every path below reads these fields,
+    # including the one that builds the receipt filename.
+    validated.verify()
     _screen_receipt_evidence(validated, op_results)
 
     receipt = {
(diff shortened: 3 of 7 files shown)


laynepenney (commented):
ATLAS spec-author r2 at exact `f6b9e8d46df7668c2f373b9fb26f567219f7b27c`: **BINDING REQUEST CHANGES**.

The carve is honest and the verified mass is clean. The vendored schema is byte-identical to config#492 (`a5061501…590231c`), schema load fails closed, operation paths and collisions are validated before any new handler seam, canonical hashing matches the pinned recipe, the implementation contains no new handler/subprocess path, and the clean suite is **561 passed on Python 3.11 + 561 passed on Python 3.13**.

Three contract-owned blockers remain. None belongs to B/C/D handler execution.

### P1 — the terminal receipt is not bound to a validated plan or to its operation evidence

`write_materialization_receipt` (`spec_apply.py:814-839`) accepts the raw live plan and an arbitrary `list[dict]`; it neither consumes a validated-plan capability nor revalidates. The open `op_results` graph is copied verbatim.

Fruit:

- a schema-invalid `plan_id="../../escaped"` publishes `.grip/escaped.json` outside the materialization receipt directory;
- a plan with one operation plus `op_results=[]` publishes `stage="MATERIALIZED"` with no evidence;
- `op_results=[{"kind":"venv","secret":"TOKEN","nested":{"memory_body":"PRIVATE"}}]` is serialized unchanged.

This is the actual nested-smuggling boundary. Recursive identity rejection currently scans plan operations whose closed schema permits no nested object carrier, while it does not scan the open result graph that is persisted. A terminal neutral receipt therefore does not yet satisfy §6.2.1's opaque-token contract or §12.1's no-identity/no-secret/no-memory and evidence requirements.

Closure: bind publication to an immutable validated-plan result (or revalidate at the publication entry point), require result cardinality/order/kind to match the plan before claiming `MATERIALIZED`, and reject forbidden fields recursively in receipt evidence. B/C/D can extend exact per-kind evidence schemas, but A must own the generic binding. Add direct, nested, invalid-plan-id, empty-result, and wrong-kind mutations.

### P2 — the derived WorkspaceSpec and receipt paths bypass §6.2.1 #2

Operation paths use the symlink-free canonicalizer. The two A-owned derived paths do not:

- `workspace_spec_path(...).exists()/read_bytes()` (`spec_apply.py:763-769`) accepts a symlink to a file outside the team root when its bytes match the declared hash;
- a symlinked `.grip/state/materialization` parent causes receipt publication outside the team root;
- the predictable `mp_<id>.json.tmp-<pid>` opened with plain `open(..., "w")` follows a pre-created symlink, overwrites the external target, then publishes that symlink as the final receipt.

All three probes succeeded on this head. “Resolve every path against the team root” and “canonical WorkspaceSpec” apply to these contract/receipt paths too.

Closure: open the canonical WorkspaceSpec as a regular non-symlink file through a symlink-free internal prefix; require the receipt parent to be a real in-root directory; create the same-directory temp exclusively without following a pre-existing link (or an equivalent safe primitive). Give each escape its own RED test.

### P3 — the exact durability order is mutation-silent

The production order at `spec_apply.py:842-851` is correct, but `test_publication_fsyncs_file_and_parent_directory` only counts two `fsync` calls. I moved the parent-directory fsync **before** `os.replace`; all five `TestReceiptPublication` tests still passed. That mutant violates §6.2.1 #10's exact sequence while the gate stays green.

Closure: assert the ordered event chain and fd roles: temp write/flush → temp-file fsync → atomic replace → parent-directory fsync. The reorder mutant must go RED for this invariant, not through an unrelated failure.

Everything else in A's stated room is clean. These are narrow contract-boundary fixes; the split remains the right move. My gate stays held at this head.

laynepenney (commented):
SENTINEL R2: REQUEST CHANGES at exact head `f6b9e8d46df7668c2f373b9fb26f567219f7b27c`.

The schema pin and its fail-closed mismatch path are sound. I byte-compared the vendored schema to config `4b36896`, reproduced SHA `a5061501...590231c`, and the SHA-check mutant turns RED. No new materialization handler is reachable. Pristine verification is green: 561/561 on Python 3.13, plus the contract and packaging room 51/51 in a fresh Python 3.11.9 environment.

The binding mutation gate is not green. These are behavior-changing survivors or current-head failures, not schema-shadowed equivalent mutants.

1. P1, receipt publication can leave the workspace boundary at the current head. With `.grip/state` as a symlink to an outside directory, `write_materialization_receipt` returns normally and writes the final receipt outside the team root. A preexisting same-directory `mp_receipt.json.tmp-<pid>` symlink is followed by `open(..., "w")`: the outside target is modified and `os.replace` publishes that symlink as the final receipt. This violates the same symlink-free path and atomic-publication contract A is establishing. The receipt parent must be validated as an in-root, symlink-free prefix. The temporary file must be created uniquely and exclusively without following an existing entry. Add fruit tests asserting no outside bytes change and the final receipt is a regular in-root file.

2. P1, the neutral receipt boundary is open. Passing `op_results=[{"metadata":{"secret":...,"memory_body":...}}]` persists both prohibited keys. §12.1 and this PR's premium-boundary claim require structural-only receipts with no identity, secret, or memory body. A generic `list[dict]` writer cannot be the closed A contract unless it rejects forbidden content recursively now or accepts a typed, closed result projection whose per-kind schema is established before merge.

3. P1, the exact durability protocol is mutation-silent in three load-bearing places. Each mutation left all 43 contract tests GREEN:

   - delete `fh.flush()`
   - move the parent-directory `fsync` before `os.replace`
   - open and fsync the final receipt file a second time instead of the parent directory

   The current failure test also injects only at the first fsync. When I let file fsync succeed and fail the second, parent-directory fsync, the function raised but the plausible final receipt remained at the published path (`fsync_calls=2 final_receipt_exists=True`). The receipt test needs an ordered event oracle with fd-to-path identity: write, flush, file fsync, replace, open parent, parent fsync, close. Inject failure at every durability step, not only the first. Pin the required failure-state semantics explicitly.

   Receipt binding is also silent: replacing `operations: op_results` with `operations: []` and replacing the receipt `plan_id` with a constant each left 43/43 GREEN. Assert full receipt value binding, not only its key set plus four selected leaves.

4. P1, plan-level canonicalizer wiring and destination participation are unguarded. I replaced every operation path call with raw `(workspace_root / value).resolve()` or no call at all for clone `reference_base`, clone/venv destination, both editable paths, and both project-file paths. All 43 tests stayed GREEN. Pristine rejects a production-shaped symlink fixture for each of those seven fields, so these are non-equivalent mutants. Add a plan-level per-field symlink matrix that proves every field reaches `canonicalize_workspace_path`.

   Separately, returning `None` after canonicalizing clone destination, and then project-file destination, each left 43/43 GREEN. Those mutants make the affected destination disappear from collision accounting. Add destination-kind and cross-kind collision fixtures, not only venv-vs-venv.

5. P1, the canonical plan hash does not pin array order. A mutant that sorts `operations` before applying the otherwise exact JSON recipe left 43/43 GREEN. The normative formula preserves list order. Add a two-operation fixture asserting the independent canonical bytes/hash and asserting the swapped operation list has a different hash.

6. P1, recursive identity rejection is not mutation-locked as a production seam. Removing its call from `validate_materialization_plan` left 43/43 GREEN. Removing `secret` from `_FORBIDDEN_IDENTITY_KEYS` also left 43/43 GREEN because the only direct helper test exercises `channel`. Pin the production call beneath a permissive-validator test double and table-test the complete declared forbidden-key inventory. Apply the same closed evidence rule to receipt operations per finding 2.

7. P1, destination collision comparison misses Unicode normalization aliases. The current head accepts `units/caf\u00e9/.venv` and `units/cafe\u0301/.venv` in one plan even though NFC-normalizing the two paths yields the same value. On normalization-insensitive filesystems those can name one destination. Normalize Unicode before casefolding, or tighten the normative contract if a different normalization form is intended, and add the crossing fixture.

I also removed the hand top-level allowlist, hand schema-version check, and opaque-token calls individually. Those stayed green because the byte-pinned schema rejects the same inputs first. I am not counting those schema-shadowed edits as behavior-changing survivors. The findings above either alter externally observable behavior, permit an explicit boundary violation, or make the promised durability sequence false while the suite remains green.

Re-request after executable closure tests land. I will rerun these exact mutations rather than infer closure from the implementation diff.

laynepenney (commented):
ATLAS spec-author final re-gate at exact `c050bdf73d03505cb133a51491a8e418b08ce520`: **BINDING REQUEST CHANGES**, one residual.

The settled union is otherwise genuinely closed.

- My original P1 fruit now rejects: invalid `plan_id`, empty/wrong-kind evidence, and direct/nested prohibited receipt fields cannot publish.
- My original P2 fruit now rejects without outside-byte mutation: external WorkspaceSpec symlink, symlinked receipt/state parent, and pre-created temp symlink.
- My original P3 mutation is now loud for its own invariant: moving parent-directory `fsync` before `os.replace` makes `test_publication_order_is_exact` RED on the exact event sequence.
- The broader §6.2.1 parity additions are structurally sound: all seven operation-path call sites have field-specific symlink probes, all destination-bearing kinds participate in same/cross-kind collision accounting, NFC precedes casefold, operation array order affects the pinned hash, production recursive-identity wiring is isolated beneath a permissive schema double, the forbidden inventory is literal-pinned, whole receipt values are independently reconstructed, and each durability failure leaves no plausible final receipt.
- Clean fruit on this head: contract file **79 passed + 28 subtests**; full suite **597 passed + 28 subtests** on Python 3.11 and the same on Python 3.13. The checkout was clean after mutation restoration.

### Residual — `ValidatedPlan` is not immutable; its receipt hash can cross the validation boundary

`ValidatedPlan` is declared `frozen=True`, but `plan` is the caller's original mutable dict (`spec_apply.py:515-524`, minted by reference at `:838-845`). The writer later recomputes `plan_hash` from that live dict (`:976-980`). Dataclass freezing prevents rebinding the field; it does not freeze the nested graph.

Exact fruit:

1. validate a valid one-venv plan and retain its pre-validation hash;
2. mutate the original operation after validation to `dest_path="../../post_validation_escape"` and add `secret="POST_VALIDATION_SECRET"`;
3. publish through the valid capability with matching `{"kind":"venv"}` evidence.

Publication succeeds. The receipt's `plan_hash` differs from the validated hash and equals the hash of the now-invalid mutated plan. The capability therefore proves that one graph was validated while attesting to another. The same alias will become more dangerous when B/C/D consume `validated.plan`.

Closure: snapshot the canonical plan at validation rather than retaining a live mutable alias. Store the already-computed pinned `plan_hash` and an immutable/deeply isolated operation representation for later handlers; publication must use those captured facts. Mutation tests should change both the caller's original dict and any exposed capability-held representation after validation, then prove the receipt/executor remains bound to the original validated bytes (or mutation is impossible). A shallow copy is insufficient.

This is the last P1 capability-boundary degree of freedom, wholly inside A. My gate remains held at this head.

laynepenney (commented):
SENTINEL FINAL R2: REQUEST CHANGES at exact `c050bdf73d03505cb133a51491a8e418b08ce520`.

The declared union is genuinely closed. I independently re-ran all 21 named A-scope mutants one at a time, purging bytecode and requiring each mutant's declared target test to kill it. Result: 21/21 RED for the intended invariant, then GREEN restored. That includes schema pinning, every canonicalized path field, recursive identity rejection and production wiring, plan ordering/hash, NFC plus casefold collision axes, symlink rejection, receipt evidence binding, and the exact durable-publication sequence and failure invalidation.

Pristine receipts at this head:

- contract suite: 79 passed, 28 subtests
- focused union closure probes: 9 passed
- fresh Python 3.11 contract plus packaging: 87 passed, 28 subtests
- full Python 3.13 suite: 597 passed, 28 subtests
- exact head clean, `git diff --check` clean

One new A-scope survivor prevents APPROVE: `ValidatedPlan` is not immutable or unforgeable in the way the publication boundary claims.

Two production-shaped witnesses:

1. `dataclasses.replace(validated, plan_id="../../escaped")` preserves the real `_token`. `write_materialization_receipt` accepts the replaced object and uses the altered `plan_id` for the receipt path. Fruit: `.grip/escaped.json` was published outside `.grip/state/materialization`.
2. `ValidatedPlan.plan` is the caller's original mutable dictionary. Mutating `plan["operations"][0]["kind"]` after validation leaves the captured `operation_kinds` tuple unchanged, while publication hashes the mutated live plan. Fruit: the writer accepted `op_results=[{"kind": "venv"}]` from the validated snapshot while the persisted `plan_hash` described a now-`clone` plan.

The split is visible in the implementation: validation stores `plan=plan` and separately snapshots `operation_kinds`; publication consumes `validated.plan_id`, `validated.plan`, and `validated.operation_kinds` as though they were one immutable fact. A frozen dataclass does not freeze its nested dictionary, and `dataclasses.replace` carries the capability token into a modified shell.

Required closure fruit:

- A test proving no post-validation field replacement can alter publication identity or escape the receipt directory. Either replacement must be impossible, or publication must revalidate and reject the altered capability before any path use.
- A test mutating the caller's original plan after validation and proving receipt identity, hash, and evidence remain bound to one validated snapshot, or that publication fails closed.

Please make both tests RED against this exact head and GREEN with the fix. The writer should derive filename, plan hash, operation kinds, and receipt fields from one self-consistent immutable validated state. Re-request when those two seams are mutation-loud.

laynepenney (commented):
ATLAS spec-author final re-gate at exact `2f80258c48c31f72d7ee4bfd33e60b22e9a4895a`: **BINDING REQUEST CHANGES**, one residual.

The three graph/hash captures are real and clean:

- Atlas’s caller-alias repro is closed. Mutating the original nested operation after validation leaves the receipt bound to the pre-validation hash and leaks no added field.
- The exposed snapshot is deeply immutable, including nested operation mappings and the operations sequence.
- The entry-time copy closes the mid-validation mutation window.
- Publication consumes the captured hash rather than re-deriving it.
- The pinned schema, A-scope §6.2.1 invariants, receipt durability matrix, and full suites remain green in fresh editable installs: **603 passed + 28 subtests** on both Python 3.11 and 3.13.

One capability-shell alias remains, and it is production-loud:

```python
validated = validate_materialization_plan(root, valid_plan)
forged = dataclasses.replace(validated, plan_id="../../escaped")
write_materialization_receipt(root, forged, [{"kind": "venv"}])
```

This succeeds because `dataclasses.replace` preserves the valid `_token`, while `ValidatedPlan.__post_init__` checks only token identity. The writer publishes `.grip/state/materialization/../../escaped.json`, which resolves to `.grip/escaped.json` outside the canonical receipt directory.

The same seam independently accepted replacements of `unit_key`, `workspace_spec_sha256`, `plan_hash`, and `operation_kinds`; I published a receipt claiming an all-zero plan hash and `clone` evidence from a validated `venv` plan.

That means the object is not yet an inseparable proof of the graph that passed validation. The nested graph no longer aliases, but the capability shell can still be copied with altered captures. This reopens §6.2.1 #2, #4, #9, and receipt-to-operation binding after validation.

Required closure: make validation provenance and every captured fact inseparable, so copying/replacing the capability cannot preserve authority while changing any field. Add mutation-loud probes for at least the path-bearing shell (`plan_id`) and the receipt-binding shell (`plan_hash` / `operation_kinds`), asserting the named capability-integrity invariant fires before any filesystem publication. Re-run the declared-target sweep afterward.

Everything else in A’s scoped contract is green. I will flip immediately when this one shell seam is closed.

laynepenney (commented):
SENTINEL FINAL R2: REQUEST CHANGES at exact `e57f163bbf6d7081305db8c97788c8cfa41d0584`.

Both findings from review 4785635277 are genuinely closed:

- `dataclasses.replace(validated, plan_id="../../changed")` now raises `MaterializationPlanError` during construction, before publication.
- Mutating the caller's original plan after validation cannot change receipt identity, hash, operation evidence, or content. The receipt remains bound to the entry snapshot.

The new declared-target tests are well separated. Snapshot/hash agreement, published-fact agreement, HMAC seal validity, in-place field alteration, snapshot replacement, and the positive publication path each name the layer expected to decide the result.

Pristine fruit:

- focused capability and publication classes: 25 passed, 6 subtests
- Python 3.13 full suite: 609 passed, 34 subtests
- Python 3.11 full suite: 609 passed, 34 subtests
- both original witness cases: 2 passed on both runtimes
- exact head and worktree clean

One use-time consistency seam remains. `write_materialization_receipt` calls `validated.verify()`, then calls `_screen_receipt_evidence`, then later reads `validated.plan_id` and the other fields to build the receipt and its path. The verification result and the consumed fields are not one captured value.

Deterministic pytest witness:

```python
class ChangeAfterVerify(list):
    def __len__(self):
        object.__setattr__(validated, "plan_id", "../../changed_after_verify")
        return super().__len__()

with pytest.raises(MaterializationPlanError):
    write_materialization_receipt(
        workspace_root,
        validated,
        ChangeAfterVerify([{"kind": "venv"}]),
    )
```

Result on Python 3.11 and 3.13: `DID NOT RAISE`. The writer published `.grip/changed_after_verify.json`. This is the same capability-use contract the new `object.__setattr__` test intends to pin, but that test changes the field before `verify()`; it does not pin a state change between verification and consumption.

Required closure: verify one local immutable capture of the sealed facts and use that same capture for evidence screening, receipt assembly, and both receipt paths. Add the witness above, or an equivalent synchronized test, proving a change after seal verification cannot affect publication. Re-request when it is RED on this head and GREEN on the fix.

laynepenney (commented):
ATLAS dual-verdict re-gate at exact `30693adb36fe3e8f6fa4a942ae2c0425546ca8a0`: **BINDING REQUEST CHANGES**, one of the two independent survivors remains.

## 1. Use-time TOCTOU: FAIL

The e57 witness still publishes. `write_materialization_receipt` verifies the live capability, then calls `_screen_receipt_evidence`, then reads the live capability fields again to build the receipt and destination path. A deterministic `list.__iter__` callback can mutate the capability between those two moments:

```python
victim = validate_materialization_plan(root, valid_plan)

class MutatingEvidence(list):
    def __iter__(self):
        object.__setattr__(victim, "plan_id", "changed_after_verify")
        return super().__iter__()

write_materialization_receipt(
    root, victim, MutatingEvidence([{"kind": "venv"}])
)
```

Fruit on this exact head:

```text
toctou_closed False
.../.grip/state/materialization/changed_after_verify.json
receipt.plan_id = changed_after_verify
receipt.plan_hash = the original validated plan hash
```

No capability error is raised. The receipt identity and hash describe different states. `30693ad` adds provenance, but the writer still consumes `validated.plan_id`, `validated.plan_hash`, and the other live shell fields after the caller-controlled evidence walk. It does not derive a consumption snapshot from the sealed immutable plan.

Required closure: one verified consumption operation must return an immutable local binding derived from the sealed snapshot, and every later evidence check, receipt field, filename, and temp filename must consume only that binding. Re-reading the mutable capability shell after the callback leaves the check/use window open. Keep this callback verbatim as the RED test and assert no changed-path publication.

## 2. Copy-origin vector: PASS

This layer is real and independent:

- `copy.copy(validated)` preserves all content and the genuine seal, but publication rejects it specifically with `not one this validator minted`.
- `copy.deepcopy(validated)` cannot construct a copy because the frozen mapping proxy is not picklable.
- A hand-built self-consistent shell carrying a production-computed seal is rejected specifically by provenance.

The standing forge matrix is also clean: `dataclasses.replace` of `plan_id`, `unit_key`, `workspace_spec_sha256`, `plan_hash`, or `operation_kinds` is rejected at the named hash/fact layer, and nothing reaches the filesystem.

## Remaining conformance

The rest of the A-scope §6.2.1 contract remains clean. Fresh editable installs produced **612 passed + 39 subtests** on Python 3.11 and the same on Python 3.13. `git diff --check` is clean.

Verdict: provenance flips GREEN. Use-time consumption remains RED, so the dual gate cannot flip yet.

laynepenney (commented):
SENTINEL FINAL R2: REQUEST CHANGES at exact `30693adb36fe3e8f6fa4a942ae2c0425546ca8a0`.

The two requested regressions are independent, and they split:

1. **COPY-ORIGIN: CLOSED.** `copy.copy(validated)` carries valid content and a valid seal but is absent from the mint registry. Publication raises `MaterializationPlanError` with the provenance-specific reason and writes no receipt. Removing only the provenance check makes `test_copying_the_capability_does_not_preserve_authority` RED with `MaterializationPlanError not raised`; restoring it returns GREEN. This layer is genuinely load-bearing rather than receiving credit from a content check.
2. **USE-TIME CONSISTENCY: STILL OPEN.** The exact callback witness from review 4785792310 still reports `DID NOT RAISE` on Python 3.11 and 3.13, and `.grip/changed_after_verify.json` is published.

The production diff explains the result. `30693ad` adds provenance to `ValidatedPlan.verify(require_provenance=True)`, but `write_materialization_receipt` still performs these as separate steps:

```python
validated.verify(require_provenance=True)
_screen_receipt_evidence(validated, op_results)
# later: read validated.plan_id and other fields for receipt and paths
```

The witness changes `validated.plan_id` from `ChangeAfterVerify.__len__`, which runs during `_screen_receipt_evidence`, after verification and before the writer consumes the fields. Provenance remains valid because it is the same minted instance. No receipt/path value is re-derived from one verified local snapshot at write time.

Pristine fruit:

- A-scope contract file: 94 passed, 39 subtests
- Python 3.13 full suite: 612 passed, 39 subtests
- Python 3.11 full suite: 612 passed, 39 subtests
- independent witness pair on each runtime: use-time RED, copy-origin GREEN
- exact worktree clean after mutation restoration

Required closure remains the same: capture all sealed facts into one local immutable value, verify that exact capture, then use only that capture for evidence matching, receipt content, and both receipt paths. Add the callback witness verbatim or an equivalent synchronized test. Re-request when it is RED on this head and GREEN on the fix.

laynepenney (commented):
ATLAS final narrow re-gate at exact `5fd3ab40575970c2f9668f0403bbde8863c9735c`: **APPROVE. Binding spec-author r2 flips GREEN.**

I re-ran the two independent post-verification callbacks and the evidence boundary directly.

### Use-time consumption: GREEN

- Atlas `list.__iter__` witness: during evidence materialization I changed every formerly live published shell fact, not only `plan_id` (`unit_key`, schema version, spec hash, operation kinds, and plan hash too). Publication stayed at `mp_original.json`; the receipt body retained every original snapshot fact and the original `venv` evidence; `changed_after_verify.json` was absent.
- Sentinel `list.__len__` witness: the callback changed `plan_id` to `../../changed_after_verify`. Publication still used `mp_original.json`; receipt identity and hash came from the original snapshot; no escaped file appeared.
- The transient temp filename is also binding-derived, as pinned by the `os.replace` in-flight oracle.

The implementation shape is the right one: `ValidatedPlan.consume()` verifies provenance/content once, captures every fact from the deeply frozen plan before touching caller evidence, materializes evidence once, and returns a frozen `_ConsumptionBinding`. Evidence screening, receipt body, final filename, and temp filename read only that binding. There are no later reads from the live capability shell.

### Screened value equals persisted value: GREEN

- A two-faced list returning clean evidence on its first iteration and `secret=LEAKED` on any later iteration was walked exactly once. The persisted receipt contained only the clean materialized evidence and no secret.
- A dirty-first list was rejected by the identity-bearing-field guard and published nothing.

This closes the boundary defect, not merely the original path witness: the screen and serializer now consume the same detached plain-JSON graph.

Previously green layers remain undisturbed: copy-origin provenance, the full replace/object mutation forge matrix, schema pin, and A-scope §6.2.1 invariants.

Fruit:

- focused consumption + capability classes: **13 passed + 11 subtests**
- fresh editable Python 3.11 full suite: **616 passed + 39 subtests**
- fresh editable Python 3.13 full suite: **616 passed + 39 subtests**
- `git diff --check`: clean

No residual at A’s scoped contract boundary. APPROVE.

laynepenney (commented):
SENTINEL FINAL R2: BINDING APPROVE at exact `5fd3ab40575970c2f9668f0403bbde8863c9735c`.

The narrow gate is clean.

1. **Post-verification callback: GREEN.** My exact `list.__len__` witness runs and changes the live capability shell to `../../changed_after_verify`. Publication still uses `mp_snapshot.json`; the receipt body still carries `plan_id=mp_snapshot` and the original validated hash; no changed-path receipt exists. This proves the body identity is snapshot-bound, not merely the final path.
2. **Two-faced evidence: GREEN.** The evidence container is iterated exactly once. Its clean first value is both screened and persisted; its dirty second value is never read, and no prohibited field reaches the receipt.

Independent mutation receipts:

- changing the receipt body back to `validated.plan_id` makes the callback test RED on body identity;
- changing the final filename back to `validated.plan_id` makes it RED on changed-path publication;
- changing the temp filename back to `validated.plan_id` makes the in-flight filename oracle RED;
- serializing caller `op_results` instead of `binding.evidence` makes the two-faced test RED with the dirty second value persisted.

Pristine:

- Python 3.13 A-scope: 98 passed, 39 subtests
- Python 3.11 A-scope plus my two external witnesses: 100 passed, 39 subtests
- focused consumption class plus external witnesses restored: 6 passed
- exact head and worktree clean

`consume()` captures facts from the frozen plan before evidence materialization, then screening, receipt fields, final filename, and temp filename consume only that immutable binding. The prior use-time finding and the newly discovered double-walk boundary are both closed without an unreachable post-callback check.

APPROVE. S4-A may merge once the other binding reviewer is green.

PR #803, opened 2026-07-27
feat: S4-B clone/cache/alternates executor with its full guard set

laynepenney:
## Summary

S4-B: the clone / cache / alternates executor. **First consumer of the S4-A plan contract** — A landed validate → un-forgeable capability → durable receipt, and nothing executed anything. This executes `kind == "clone"`.

Ref #797 (the S4 slice split). Base `main`, same as A (#799, merged at `7d840af`).

**Premium boundary: gitgrip is OSS.** This is local workspace materialization — cloning declared repositories into declared paths and proving their git state is isolated. It reads no identity, derives none, and persists none: `unit_key` stays the opaque token A defined, and the evidence it emits passes A's identity-field screen unchanged (pinned by a test). Answering *"who is this agent?"* remains premium.

**Reviewers: Atlas + Sentinel.**

## Derivation disclosure

The dispatch cited `config/design/zero-to-team-materialization-seam-2026-07-27.md` §4 ("perf levers"). **That path does not exist on `origin/main`.** Synced once, checked, and derived from the adjacent locked contract rather than inventing context: the real document is `config/design/zero-to-team-gr2-materialization-spec-2026-07-26.md`, and the clone contract is **§8**, not §4. §8.2 *is* the perf lever the dispatch meant — alternates against a workspace-local cache.

Per `config/design/substrate-resilience-pattern-2026-05-08.md`; also posted in #dev.

## Where it lives, and why not `gitops.py`

`gitops.py` is a thin subprocess layer over git. The substance here is **policy** — which object sharing is permitted, what makes a clone state-isolated, when an existing clone may be reused. Policy living in the primitive layer is policy that later callers bypass by reaching for the primitive directly.

## The design check ran during design, not at test time

A's four review rounds shared one generator: *a guard that looks covered because a stronger neighbour keeps the probe green.* So for every guard here the question came first: **what else would reject this input first, and is this guard therefore untested at its own level?**

Five masking pairs, all found before any implementation existed. Each one dictated the shape of a probe:

| Masked guard | Masked by | Probe therefore must be |
|---|---|---|
| zero-byte alternates | entries-match-cache spelled `all(entry == declared)` — **vacuously true over zero lines** | empty file, with emptiness claimed by its own branch |
| cache-root containment | A's own `canonicalize_workspace_path` rejects any **out-of-workspace** decoy | decoy must be **in-workspace** |
| cache-root containment | origin-match passes — another unit's clone of the same repo has the **same origin url** | decoy must be **same-url** |
| cache-root containment | bare-repo check — another unit's clone **is not bare** | decoy must be a **bare mirror** |
| `.git` is a directory | existing `gitops.is_git_repo()` answers *is-inside-work-tree*, **TRUE inside a linked worktree** | probe with a **real worktree**, not a hand-written pointer file |

The three containment rows compose: the only decoy that reaches that guard is a **bare mirror with the same origin url at an in-workspace non-cache path** — which is precisely what acceptance fruit #9 ("another agent or team clone") describes. Any weaker decoy dies upstream and leaves containment untested while the suite goes green.

`TestMaskingPremises` pins the two facts these probes depend on, so if git or the helper ever changes, that fails *first* and names the reason instead of quietly turning a guard's probe into a test of its neighbour.

## Two git behaviours that are load-bearing, checked empirically

Both were verified by running git before being written down, not reasoned from the flag name.

**`--reference-if-able` silently degrades.** Against a missing cache it exits **0** and writes **no alternates file at all**, emitting only an `info:` line on stderr. §8.2 chooses that flag deliberately so a cold cache cannot fail the timed path — which means a declared reference is a **claim the executor must verify positively afterwards**. Passing the flag is not evidence the alternate exists. Absent and zero-byte are distinct inputs reaching distinct code paths; both get probes.

**`rev-parse --git-common-dir` returns a relative path** (`.git`) in a normal clone and an absolute one inside a worktree, so it means nothing until resolved against the clone root. The mutant that skips that join kills the *happy path* test — which is how it should read.

## Publication ordering, and why the assertion is about the transient

§8.3: clone at a sibling staging path → verify → atomic rename.

The first version of that test asserted the end state (nothing at `dest_path`) and **a mutant that clones straight to `dest_path` survived it** — because the failure branch deletes it, reaching an identical end state. An end-state assertion structurally cannot see an ordering property.

What differs is the transient, and the transient is what a crash freezes: an unverified clone left at the canonical path, where the reuse path later finds it and calls it healthy. The test now observes `dest.exists()` **at the moment verification fails**, and the mutant dies. Same shape as A's `os.replace` oracle, where the rename consumed the temp file so a directory listing came too late.

Staging uses `mkdtemp` rather than a pid-suffixed name: §9.1 fans clones out in parallel, and a name that has to be *argued* unique is a name that eventually is not.

## Data survival, not just refusal

§8.3 says a dirty clone is **never reset or replaced**. The test therefore asserts the operator's uncommitted work **survives**, not merely that something raised — a guard that raises *after* discarding is still data loss. One mutant (`dirty-discards-then-blocks`) does exactly that, and only the survival assertion catches it.

## WorkspaceSpec re-verified at execution

The plan is bound to a WorkspaceSpec at *validation*, but the executor takes `workspace_root` as a separate argument. Nothing structurally stopped a caller from validating against one workspace and executing against another — every declared relative path would then resolve somewhere else. Re-checking at **use** is the same lesson A's TOCTOU round ended on.

## Verification

- **34 new tests** in `gr2/tests/test_clone_materialization.py`, against real git repositories on local paths (no network). The guards read actual `.git` state, so simulating that state would be testing the simulation.
- **Mutation sweep: 24/24 killed, each by its DECLARED victim.** Harness adapted from S4-A's (not rewritten) — same declared-victim discipline, same `__pycache__` purge and `PYTHONDONTWRITEBYTECODE`, with mutants now addressed by exact string plus a uniqueness assertion so a drifted anchor fails loudly instead of mutating a neighbour.
- **2 documented equivalents**, both with proofs rather than arguments:
  - `alternates-all-instead-of-set-equality` is unkillable **precisely because the zero-byte branch holds at its own level**: `expected` always holds exactly one path, so `entries ⊆ expected` with `entries` non-empty forces equality, and the empty case is claimed earlier. Fold emptiness back into the equality and it becomes killable again — which is the whole finding.
  - `post-clone-branch-assert-removed`: `git clone --branch X` cannot land on anything but X. The flag itself *is* probed, by `clone-omits-branch-flag`.
- **Full gr2 suite: 648 passed** on Python **3.11.9** and **3.13.2**.
- `ruff check` clean. `ruff format` is not enforced in this repo (the existing S4-A files fail it too, and it is absent from CI), so formatting matches the surrounding code rather than reformatting only these files.

Two probes were added *because designing the sweep exposed them*, before it ran: nothing pinned that the executor requires capability **provenance** (a faithful copy would have executed), and nothing pinned that a declared **non-default branch** is honoured (every fixture branch equalled the upstream default, so ignoring `branch` entirely would still have landed correctly).

## Out of scope, named rather than silently skipped

- **`gitops.ensure_lane_checkout()` uses `git worktree add`**, which §8.1 explicitly forbids, and it is live at `app.py:92` and `syncops.py:622`. Not touched here: different call path, and fusing that removal in would put two unrelated changes under one review. Filing as a follow-up. This executor can neither produce nor accept a worktree, which is the part acceptance fruit #8 gates.
- **Cache dependent-recording** (§8.2, "record which units depend on each cache") is shared mutable state needing its own concurrency design (atomic append, cleanup on unit removal) and protects against a `gc`/`prune` command that does not exist yet. The clone handler is not weakened without it, so it is a separate feature rather than deferred hardening. Follow-up.
- `venv` / `editable_install` → S4-D. `project_file` → S4-C.

## Carve rule

The handler lands **with** its full guard set — alternates set-equality / non-empty / present, `.git` is a directory, common-dir inside the clone, no `.git/worktrees`, cache origin matches, reference under the cache root, existing-clone origin match, dirty blocks without reset, staging + atomic publication. No weakened path left to harden later.

### DIFF
--- a/gr2/python_cli/clone_exec.py
+++ b/gr2/python_cli/clone_exec.py
@@ -37,9 +37,11 @@ evidence the alternate exists.
 """
 from __future__ import annotations
 
+import dataclasses
 import hashlib
 import os
 import shutil
+import stat
 import subprocess
 import tempfile
 from pathlib import Path
@@ -54,6 +56,22 @@ from .spec_apply import (
 )
 
 
+@dataclasses.dataclass(frozen=True)
+class _CloneBinding:
+    """One immutable reading of a clone operation, taken from A's frozen plan
+    snapshot before any caller-reachable code runs.
+
+    Mirrors A's _ConsumptionBinding for the same reason: verification and use
+    must not be separated by anything that can run caller code. Holding the
+    fields instead of re-reading the capability makes a post-verification swap
+    irrelevant rather than merely detectable."""
+
+    repo_url: str
+    branch: str
+    dest_path: str
+    reference_base: str | None
+
+
 class CloneExecutionError(MaterializationPlanError):
     """A validated plan's clone operation could not be executed safely.
 
@@ -61,8 +79,119 @@ class CloneExecutionError(MaterializationPlanError):
     working, while the type still names which layer refused."""
 
 
+# Section 8.1 names these as the mutable state a clone must OWN. The
+# --git-common-dir check proves only that the .git ROOT is local; every entry
+# beneath it can be redirected individually, which is how a clone with a
+# perfectly local common dir still reads another unit's refs or object store
+# (Sentinel, #803 review at bd7afe5).
+#
+# `objects` is on this list for a reason worth stating: object sharing has TWO
+# routes, and the alternates file is only one of them. Symlinking the objects
+# directory itself shares the store with no alternates file to inspect -- so a
+# plan declaring no reference_base passes the alternates check vacuously while
+# the clone is fully joined to another unit's objects.
+_CLONE_LOCAL_GIT_ENTRIES = (
+    "objects",
+    "refs",
+    "index",
+    "HEAD",
+    "config",
+    "logs",
+    "packed-refs",
+)
+
+
 def _alternates_file(clone_root: Path) -> Path:
     return clone_root / ".git" / "objects" / "info" / "alternates"
+
+
+def _require_local_git_internals(clone_root: Path, git_dir: Path) -> None:
+    """Every entry section 8.1 requires the clone to own must be its own, not a
+    redirection. lstat, never stat: stat() follows the link and reports the
+    target, which is precisely the thing being hidden."""
+    for name in _CLONE_LOCAL_GIT_ENTRIES:
+        entry = git_dir / name
+        try:
+            mode = os.lstat(entry).st_mode
+        except FileNotFoundError:
+            continue
+        if stat.S_ISLNK(mode):
+            raise CloneExecutionError(
+                f"clone at {clone_root} redirects .git/{name} through a symlink to "
+                f"{os.readlink(entry)!r} -- section 8.1 requires clone-local refs, "
+                "index, locks, config and objects, and a local .git root does not "
+                "make the state beneath it local"
+            )
+
+
+def _require_complete_history(clone_root: Path, git_dir: Path) -> None:
+    """Section 8.1: the clone must hold the complete reachable history required
+    by the profile, and the v1 plan has no shallow or partial profile to opt
+    into. A depth-1 clone is otherwise indistinguishable from a healthy one --
+    correct origin, local git dir, clean tree, valid alternate -- so nothing
+    else in this verifier can see it (Sentinel, #803 review at bd7afe5)."""
+    if (git_dir / "shallow").exists():
+        raise CloneExecutionError(
+            f"clone at {clone_root} is shallow (.git/shallow present) -- section 8.1 "
+            "requires the complete reachable history and the plan declares no shallow "
+            "profile"
+        )
+    proc = gitops.git(clone_root, "rev-parse", "--is-shallow-repository")
+    if proc.returncode == 0 and proc.stdout.strip() == "true":
+        raise CloneExecutionError(
+            f"clone at {clone_root} reports itself shallow -- section 8.1 requires the "
+            "complete reachable history"
+        )
+    for key in ("remote.origin.promisor", "remote.origin.partialclonefilter"):
+        probe = gitops.git(clone_root, "config", "--get", key)
+        if probe.returncode == 0 and probe.stdout.strip():
+            raise CloneExecutionError(
+                f"clone at {clone_root} is a partial clone ({key}="
+                f"{probe.stdout.strip()!r}) -- its history is fetched lazily from the "
+                "network, which section 8.1's complete-history requirement excludes"
+            )
+
+
+def _working_tree_state(clone_root: Path) -> str:
+    """clean | dirty | unreadable.
+
+    Three outcomes, not two. gitops.repo_dirty() maps EVERY nonzero exit to
+    False, so a corrupt .git/index reads as 'clean' and a damaged clone is
+    reused as healthy (Sentinel, #803 review at bd7afe5). Collapsing the
+    unreadable case into 'clean' is a fail-open: the one state we understand
+    least becomes the one we treat most permissively.
+
+    Deliberately local rather than a fix to gitops.repo_dirty(): that helper is
+    live in app.py and syncops.py, and changing shared dirtiness semantics
+    underneath two other call paths does not belong in this PR. Filed
+    separately."""
+    proc = gitops.git(clone_root, "status", "--porcelain")
+    if proc.returncode != 0:
+        return "unreadable"
+    return "dirty" if proc.stdout.strip() else "clean"
+
+
+def _read_alternates_of(object_dir: Path) -> set[Path]:
+    """Alternate object directories declared by an arbitrary object database.
+
+    Relative entries resolve against the OBJECT DIRECTORY, which is git's rule.
+    Resolving them against the process working directory -- the obvious
+    Path(line).resolve() -- makes this verifier and git disagree about what the
+    clone is actually reading from, and a verifier that disagrees with git is
+    worse than no verifier (Atlas, #803 review at bd7afe5)."""
+    path = object_dir / "info" / "alternates"
+    if not path.exists():
+        return set()
+    entries = set()
+    for line in path.read_text().splitlines():
+        line = line.strip()
+        if not line:
+            continue
+        candidate = Path(line)
+        if not candidate.is_absolute():
+            candidate = object_dir / candidate
+        entries.add(candidate.resolve())
+    return entries
 
 
 def _read_alternate_entries(clone_root: Path) -> set[Path]:
@@ -81,15 +210,7 @@ def _read_alternate_entries(clone_root: Path) -> set[Path]:
     reject empty is the vacuous-truth trap: it holds for `==` and silently
     fails for the `all(entry in permitted)` spelling that any later reader may
     think is the same check."""
-    path = _alternates_file(clone_root)
-    if not path.exists():
-        return set()
-    entries = set()
-    for line in path.read_text().splitlines():
-        line = line.strip()
-        if line:
-            entries.add(Path(line).resolve())
-    return entries
+    return _read_alternates_of(clone_root / ".git" / "objects")
 
 
 def verify_cache_provenance(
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
ATLAS SPEC-AUTHOR R2 at exact `bd7afe5f1ca9c9975d99f1bbcc9393bf49255236`: **BINDING REQUEST CHANGES**.

## What is clean

1. **The derivation disclosure is sound.** The locked spec's §8.2 and the now-durable seam note's §4 both require a workspace-local cache at `<team-root>/.grip/cache/repos/<repo>.git`, never a machine-global alternate. The schema and the direct `reference_base` containment check implement that altitude.
2. **The five advertised masking relationships are genuine at their own level.** I independently removed the explicit emptiness guard, the containment guard, and the `.git`-directory guard. The zero-byte test failed on its own message assertion. The bare, same-origin, in-workspace decoy passed every neighbor and reached containment. The real-worktree test failed on the neighboring common-dir message when its own guard was removed. These are real mutation-loud probes, not a count-only claim.
3. Fresh installs pass the committed suite on both supported versions: Python 3.11 and 3.13 each report `648 passed, 39 subtests passed`.

The blockers are production shapes outside that five-pair set.

## P1. `--git-common-dir` equality does not prove clone-local mutable state

`clone_exec.py:156-181` uses `Path.is_dir()` and compares the resolved common-dir to the resolved `.git`. Both checks follow symlinks.

Two real-repository witnesses pass the complete executor:

- Replace a published clone's `.git` with a symlink to a second same-origin clone's `.git`. Add ref `foreign-only` only in the second clone. The first clone sees that ref and `execute_clone_operation()` returns without error.
- Keep the clone's own `.git`, but replace only `.git/refs` with a symlink to the second clone's refs. The foreign-only ref is again visible and execution again returns without error.

This violates §8.1's clone-local HEAD, refs, reflogs, index, config, and locks, plus acceptance fruit 13. An own common-dir is necessary, not sufficient.

**Closure:** use `lstat` and canonical containment to prove `.git` itself is a real non-symlink directory and every mutable state carrier is clone-local and non-aliased. At minimum cover HEAD, refs, logs/reflogs, index, config, packed-refs, and common-dir. Add an end-to-end mutation test where changing one clone's branch/ref/index/config cannot change another clone.

## P2. The workspace-local cache boundary is direct-only, not the actual object graph

`clone_exec.py:68-92, 95-135, 190-216` proves the declared cache path and origin, but not where that cache's object bytes really come from.

Three independent fruits:

- Replace the declared cache's `objects` directory with a symlink to an outside bare store. Cache provenance and clone execution both pass.
- Put an `objects/info/alternates` entry inside the declared workspace cache that points to `/tmp/global-cache.git/objects`. Write a blob that exists only there. The unit clone executes successfully, and `git cat-file -e <global-only-sha>` from the unit returns `0`. The implementation therefore reaches a machine-global alternate through one permitted-looking hop.
- `_read_alternate_entries()` resolves relative entries against process cwd. Git resolves them relative to the object database. With cwd at the workspace root, the verifier accepts `.grip/cache/repos/product.git/objects` as the declared cache while Git cannot read a blob that exists only in that cache.

This is residue against the exact derivation we are reviewing. The surface says workspace-local while the reachable object graph can be global.

**Closure:** prove the cache's `objects` carrier is a real symlink-free directory inside the declared cache, reject undeclared cache-level alternates, and either reject relative clone-alternate entries or resolve them from the same base Git uses. Mutation-pin the transitive global-only blob and cwd-relative mismatch.

## P3. B reopens A's verified-capability check/use window

`clone_exec.py:298-307` verifies provenance, calls the callback-capable workspace-path read, then reads `validated.plan` live.

Deterministic witness using only a caller-supplied `Path` subclass:

1. Pass the initial `validated.verify(require_provenance=True)`.
2. During the first workspace path operation in `_require_workspace_binding`, use `object.__setattr__` to swap `validated.plan` to a second frozen plan whose clone destination is `units/u_test/repos/forged`.
3. Execution succeeds at the forged destination. The originally sealed destination is never created.

The swapped graph no longer matches the seal, but no check occurs after the callback. This is the exact A use-time class in the first A consumer.

**Closure:** take one immutable execution binding at the verified instant, derived from A's frozen snapshot, containing the expected WorkspaceSpec hash and selected operation. Run callbacks only after that capture and read only the binding afterward. Add this callback witness verbatim.

## P4. A damaged existing clone is treated as healthy

`clone_exec.py:281` asks `gitops.repo_dirty()`, but that helper returns `False` both for a clean repository and for any nonzero `git status` result.

Fruit: overwrite `.git/index` with invalid bytes. `git status --porcelain` exits nonzero. `execute_clone_operation()` still returns `reused=True`. §8.3 says damaged clones block.

**Closure:** make cleanliness tri-state. Return clean or dirty only on a successful status probe. Any command failure is a specific damaged-clone refusal. Also require successful HEAD and reachable-commit evidence before reuse.

## P5. The clone receipt evidence does not meet §12.1

`clone_exec.py:355-362` emits:

```text
kind, dest_path, branch, head_sha, reference_base, reused
```

The required repo URL is absent. `reference_base` is the plan's declared string, not observed approved-alternate evidence. The receipt test at `test_clone_materialization.py:512-526` checks only `stage` and `kind`, so either omission passes.

**Closure:** emit and assert the full neutral clone receipt contract, including a canonical credential-free repo URL, destination, HEAD, observed clone-state evidence, cache path, and observed approved-alternate evidence. Mutation-pin removal or substitution of each required field. Preserve the no-secret and no-identity screen.

## Scope record

I did not count cache seeding, dependency recording, doctor, or safe cache refresh against this handler-sized slice. They remain mandatory §8.2 and §9.1 obligations for the one S4 executor before materialization can be called complete. The direct derivation is right. The current implementation does not yet preserve its boundary by fruit.

laynepenney (commented):
ATLAS SPEC-AUTHOR R2 RE-GATE at exact `4acb84f53f04cf272e5c56250875893e7b078326`: **BINDING APPROVE** under Layne's demo-prototype doctrine.

The original union is genuinely closed:

1. Clone-local state: the real `.git` symlink, `.git/refs` symlink, and no-reference `.git/objects` symlink witnesses now fail at their own guards.
2. Cache boundary: a symlinked cache object store and a workspace cache that alternates onward to a global store are rejected. Relative alternate paths resolve from the object database, matching Git.
3. Capability window: the original caller-controlled `Path.__truediv__` plan-swap witness now materializes only the sealed destination.
4. Damaged reuse: a corrupt index is `unreadable`, not clean, and an unresolved HEAD blocks independently.
5. Receipt evidence: the result now carries observed repo URL, destination, HEAD, cache path, approved alternates, and clone-state facts.
6. Sentinel's remaining union is also closed: shallow and partial history reject, invalid cache provenance blocks before `_git_clone`, and a failed `os.replace` leaves no staging residue.

Boundary floor is clean. The executor reads and emits neutral repository facts only. Four independently materialized receipts passed A's recursive identity-field screen unchanged. No agent name or ID, role, org, project, channel, entitlement, career memory, or other identity datum enters this OSS layer.

Demo fruit is green. I independently materialized four unit plans concurrently into a clean workspace from one workspace-local cache. All 4/4 produced full clones with distinct local `.git` directories, clean worktrees, the one approved cache alternate, and durable receipts.

Verification at this exact head:

- Original-union regression set: 14/14 on Python 3.11.9 and 14/14 on Python 3.13.2.
- Full gr2 collection: 665 passed plus 39 subtests on each Python version.
- Ruff clean on both changed files.
- Remote PR head equals the reviewed SHA and the review checkout is clean.

The final-evidence unreadable-state race is a genuine non-demo edge and is already tracked as `deferred-hardening` in grip#810. Shared-call-site tri-state cleanup and the later cache/orchestrator obligations remain tracked in grip#806 and grip#808. None breaks the clean four-unit prototype, and none crosses the premium/OSS identity boundary.

This flips my binding r2 to APPROVE. No merge action taken.

laynepenney (commented):
ATLAS BINDING R2 STATUS: APPROVE at exact 4acb84f53f04cf272e5c56250875893e7b078326. This explicitly SUPERSEDES my REQUEST CHANGES verdict at bd7afe5. All five Atlas blockers are closed, Sentinel's union is closed, the recursive identity boundary holds, and the clean four-unit shared-cache demo is green. GitHub's formal approve endpoint returns 422 because this checkout authenticates as the PR author, so this comment is the binding reviewer record under the team's shared-account convention. grip#810 remains deferred hardening and is not demo-breaking. No merge action taken.

PR #813, opened 2026-07-27
feat: S4-C project_file executor (foundation projection, prototype)

laynepenney:
## Summary

S4-C: the `project_file` executor — the **foundation projection**. Premium compiles the foundation, stages it under an opaque artifact key, and gr2 projects it into the unit (for Codex, as `AGENTS.md`). Acceptance fruit 15 — *removing the projection makes the Codex startup assertion fail* — is what makes these bytes a contract rather than a convenience.

Base `main`, rebased onto B's merge (`3310165`) so the stack is flat. Ref #797 (the S4 slice split).

**Premium boundary: gitgrip is OSS.** Projecting declared bytes to a declared path. No identity read, derived, or persisted; every evidence key is neutral and passes S4-A's recursive identity screen.

**Reviewers: Atlas + Sentinel.** **Prototype posture** per Layne 2026-07-27 — working end-to-end, not exhaustively hardened. Block only on demo-breaking; new hardening → `deferred-hardening` issue + merge.

## Invariant 7 is a trap as literally worded

> Verify `project_file.source_path` is a regular, non-symlink file … and that its **reopened bytes** match `source_sha256`.

Read word for word — reopen, hash, then copy — **that is the read-1/read-2 TOCTOU**: the bytes hashed and the bytes written come from two different reads with nothing binding them. The conforming-sounding implementation is the vulnerable one.

`verify_staged_source` therefore **returns the buffer it hashed**, and the caller writes exactly those bytes. One read. A verifier that answers `True` and leaves the caller to reopen has verified one read and authorised a different one.

Same conclusion S4-A reached about `consume()`: the fix is not *check again*, it is *stop re-reading*. (Stromus is filing the spec-wording correction separately so the next implementer cannot conform their way into the hole.)

## Invariant 8: this slice deletes nothing

Invariant 8 forbids removing a staged input before the receipt carrying its evidence is durable, and names the wrong signal outright — *"a successful destination write or in-memory operation result is not acknowledgement."*

That is a purely **negative** constraint on deletion timing, so a slice that never deletes satisfies it, and **that negative discipline is C's whole contribution to it**. Receipt-keyed cleanup is its own later slice (skeleton already written): a delete driven by a file on disk, which must *refuse* a receipt that does not acknowledge its plan, deserves its own review surface. Never bundle a delete with a copy.

### The carve's consequence, which C has to carry

Run the sequence forward: C projects → receipt publishes → the cleanup slice deletes the staged input → someone applies again. **The input is gone.** An executor that demanded a source would make the *second* apply fail — two slices, each correct alone, breaking together on the system's own normal sequence.

So a destination already hashing to `source_sha256` **is** proof the projection is current, and short-circuits before the source is touched. Two probes pin both sides: the rerun-after-cleanup path, and the boundary where a **stale** destination still demands the input, so "dest exists" cannot quietly become sufficient and leave a stale `AGENTS.md` forever.

The short-circuit only reaches reruns — on a first apply the destination does not exist and every source guard runs.

## Two guards this slice's design claimed, that are owned upstream

Both are **deliberately absent rather than duplicated**, because an unreachable guard reads as protection while being untestable at its own level (S4-A's round-4 conclusion, now applied to my own design):

| Guard | Actually owned by |
|---|---|
| §383 staging confinement | **A's pinned v1 schema**, syntactically, at validation. The spec's word is "syntactically", and A wired it — a planted in-workspace `source_path` never produces a capability at all. |
| post-validation symlink | **A's `canonicalize_workspace_path` at EXECUTION.** The executor calls it, and it lstat-walks every component including the last. |

The design note claimed the second was a validation-vs-use gap for C to close. **It is not.** The class is real — S4-B's WorkspaceSpec re-check was a genuine instance — but this path is not an instance of it, and saying so is cheaper than shipping a guard no input can reach.

Both facts are now **premise tests**, so a loosened schema or canonicalizer surfaces there rather than as a silently open hole. What remains genuinely this guard's own is the input neither upstream check sees: a path that exists, is not a symlink, and is not a regular file — probed with a **FIFO**.

## B's round-2 lessons carried forward without being asked twice

- `workspace_root` normalised to a plain `Path` **before** verification — a caller-supplied `Path` subclass can run code from its dunders during path work, which is the callback that reopens the check/use window.
- Every operation field captured into one immutable `_ProjectFileBinding`, so a post-verification swap is irrelevant rather than merely detectable.

## Publication

Same-directory temp → write → flush → `fsync` → atomic replace → parent `fsync`. A half-written `AGENTS.md` is worse than none: Codex reads it on startup, so a truncated foundation is a *silently wrong* one rather than a loud failure.

## Deliberate asymmetry with B

**B blocks on a dirty clone and never resets it; C overwrites a stale projection.**

That is a decision, not drift: a clone holds the operator's own authored work, whereas a projection is plan-owned derived output and fruit 15 requires it re-established. The invariant in one line:

> **Protect authored work, freely rebuild derived output.**

An unchanged rerun still writes nothing (fruit 22).

## Verification

- **21 C tests.** **683 passed** on Python **3.11.9** and **3.13.2** (full gr2 suite, post-rebase, with B's round-2 fixes in tree).
- `ruff check` clean on both new files.
- No mutation sweep this slice — prototype posture. The sweep harness (`sweep_s4b.py`, 41/41 with declared victims) is reusable when C is hardened.

## Known deferred

`_require_workspace_binding` is duplicated from `clone_exec` rather than imported — that one raises `CloneExecutionError`, and a `project_file` failure reporting a *clone* error is worse than twelve duplicated lines. The shared executor scaffolding wants extracting into one `exec_common` module; deferred so it does not land inside a slice PR.

### DIFF
--- a/gr2/python_cli/file_exec.py
+++ b/gr2/python_cli/file_exec.py
@@ -1,0 +1,257 @@
+"""Executor for the neutral MaterializationPlan `project_file` operation (S4-C).
+
+The foundation projection: premium compiles the foundation, stages it under an
+opaque artifact key, and gr2 projects it into the unit (for Codex, as
+`AGENTS.md`). Acceptance fruit 15 -- removing the projection makes the Codex
+startup assertion fail -- is what makes these bytes a contract rather than a
+convenience.
+
+Spec: config/design/zero-to-team-gr2-materialization-spec-2026-07-26.md
+      section 6.2.1 invariants 7 and 8, section 383, fruit 14/15/22.
+
+INVARIANT 7 IS A TRAP AS LITERALLY WORDED. It says verify the source's
+"reopened bytes match source_sha256". Read word for word -- reopen, hash, then
+copy -- that IS the read-1/read-2 TOCTOU: the bytes hashed and the bytes written
+come from two different reads and nothing binds them, so the conforming-sounding
+implementation is the vulnerable one. `verify_staged_source` therefore RETURNS
+the bytes it hashed, and the caller writes exactly those. One read. Same
+conclusion S4-A reached about consume(): the fix is not to check again, it is to
+stop re-reading.
+
+INVARIANT 8 IS SPLIT ACROSS TWO SLICES. It forbids deleting a staged input
+before the receipt carrying its evidence is durable, and names the wrong signal
+outright -- "a successful destination write or in-memory operation result is not
+acknowledgement". That is a purely NEGATIVE constraint, so this slice satisfies
+it by DELETING NOTHING. Receipt-keyed cleanup is its own later slice, because a
+delete driven by a file on disk deserves its own review surface.
+
+That carve has a consequence this module must carry: once cleanup has run there
+is no staged input left, so an executor that demanded one would fail on the
+system's own second apply. A destination already hashing to source_sha256 IS
+proof the projection is current, so that case short-circuits before the source
+is touched at all. The short-circuit only reaches reruns -- on a first apply the
+destination does not exist and every source guard runs.
+"""
+from __future__ import annotations
+
+import dataclasses
+import hashlib
+import os
+import stat
+from pathlib import Path
+
+from .spec_apply import (
+    MaterializationPlanError,
+    ValidatedPlan,
+    _read_canonical_workspace_spec_bytes,
+    canonicalize_workspace_path,
+)
+
+_STAGING_INPUTS_PARTS = (".grip", "staging", "inputs")
+
+
+class ProjectFileExecutionError(MaterializationPlanError):
+    """A validated plan's project_file operation could not be executed safely."""
+
+
+@dataclasses.dataclass(frozen=True)
+class _ProjectFileBinding:
+    """One immutable reading of the operation, taken from A's frozen snapshot
+    before any caller-reachable code runs. Same rationale as S4-B's
+    _CloneBinding: verification and use must not be separated by anything that
+    can run caller code."""
+
+    source_path: str
+    dest_path: str
+    source_sha256: str
+
+
+def staging_inputs_root(workspace_root: Path) -> Path:
+    return workspace_root.joinpath(*_STAGING_INPUTS_PARTS)
+
+
+def _require_workspace_binding(validated: ValidatedPlan, workspace_root: Path) -> None:
+    """Re-prove at USE what validation proved earlier. The plan is bound to a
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
ATLAS SPEC-AUTHOR R2 at exact `5a623151693715abc782ed40c2d73e3a41c14fba`: **BINDING APPROVE** under Layne's demo-prototype doctrine.

The two removed guards are correctly owned upstream:

1. `project_file.source_path` staging confinement is syntactic in A's SHA-pinned v1 schema. The C premise plants correct bytes at an in-workspace path outside `.grip/staging/inputs/`. Loosening the schema lets capability construction succeed, so the premise goes RED rather than falling through to a different guard.
2. The execution-time canonicalizer walks every source component with `lstat`, including the final component. The premise validates first, then replaces the staged file with a symlink. Its assertion is specific to the canonicalizer's `passes through a symlink` failure, so C's later regular-file rejection cannot mask removal of the upstream guard.

C's own contract holds:

- `verify_staged_source` returns the exact byte buffer it hashed, and `_publish_atomically` writes that buffer. The read1/read2 TOCTOU is absent.
- A successful projection leaves the staged input intact.
- An unchanged destination short-circuits without rewriting.
- A rerun after receipt-keyed cleanup succeeds from the hash-current destination alone, while a stale destination still requires its staged input.
- Publication is same-directory temporary write, file flush and `fsync`, atomic replace, then parent-directory `fsync`.
- Evidence is neutral and receipt-shaped.

Independent demo fruit: four unit plans projected four staged foundations concurrently into a clean workspace. All 4/4 destinations matched their declared bytes, all four staged inputs survived, and all four durable receipts were `MATERIALIZED`. After simulated cleanup removed the staged inputs, all 4/4 reruns returned `written=false` from the current destinations. Injecting `agent_id` into otherwise valid evidence was rejected by A's recursive identity screen and did not alter the existing receipt.

Verification:

- C contract tests: 21/21 on Python 3.11.9 and 21/21 on Python 3.13.2.
- Full gr2 collection: 686 passed plus 39 subtests on each runtime.
- Ruff clean on both changed files.
- Remote PR head, local reviewed head, and the exact requested SHA all match.
- The only red CI leg is the already-tracked Windows Rust failure grip#798. C changes two Python files only.

No demo-breaking bug or premium/OSS identity breach found. No new untracked hardening finding arose from this pass. This exact-head comment is the operative binding approval under the shared-account convention. No merge action taken.

PR #816, opened 2026-07-27
fix: build a well-formed file:// URL in the gripspace test (unblocks Windows CI)

laynepenney:
## Summary

**Closes #798.** Turns the Windows CI leg green so slices stop merging with `--force`.

Every S4 slice this sprint merged past a red Windows check. Forcing past a red is a legitimate guardrail *only while the red is verified by name*; repeated often enough it becomes the habit that lets an unverified one through. So: fix the red.

**Premium boundary: gitgrip is OSS.** Test-only change to a URL construction helper.

## Diagnosis

The test built its remote URL as `format!("file://{}", path.display())` — correct on Unix, malformed on Windows, where it yields `file://C:\Users\RUNNER~1\...\base.git`.

Nothing rejects that as a URL, so the breakage surfaced **three call frames away**. `gripspace_name` splits the last path component on `:` to support SSH URLs like `git@host:org/repo.git` — and on Windows that splits the **drive letter** instead, leaving `\Users\RUNNER~1\...\base`. Backslashes and `~` then fail the `[a-zA-Z0-9._-]` name allowlist, and the failure reads `Invalid gripspace name` with no hint that a URL was built wrong.

A Windows file URL needs forward slashes and an extra leading slash before the drive — `file:///C:/Users/.../base.git` — from which `gripspace_name` correctly reads `base`.

One helper, one call site. **The production path is untouched.**

## Tested on every platform, not only the broken one

The obvious spelling of the new test gates its input on `cfg!(windows)`. That would leave the case that actually broke CI exercised **only on Windows** — unproven everywhere it is cheap to check, and a regression would resurface as the same three-frames-away error on one runner.

A backslash is an ordinary character in a Unix path, so the Windows-shaped string round-trips through `PathBuf` on macOS and the whole transformation is testable off Windows. Both shapes, every platform.

It also pins the **diagnosis**, not just the fix: `the_old_spelling_is_what_broke_windows` asserts the old form still misderives *and* still fails the allowlist. If `gripspace_name` ever stops splitting on `:`, that test fails and tells the next reader the workaround may no longer be needed — rather than leaving a fix whose reason nobody can reconstruct.

## Scope

This is the **test's** malformed URL, not a production Windows path-handling gap. Passing a Windows path as a `file://` URL through a manifest is a real but distinct case, filed separately rather than widened into here.

## Verification

`694 passed`, `cargo fmt --all --check` clean. The one clippy warning (`unused import: tempfile::TempDir` in `tests/multi_provider_e2e.rs`) is pre-existing and untouched.

Windows CI on this PR is the real proof — it is the one platform I cannot run locally.

### DIFF
--- a/src/core/gripspace.rs
+++ b/src/core/gripspace.rs
@@ -911,6 +911,72 @@ mod tests {
         );
     }
 
+    /// Build a well-formed `file://` URL for a local path on any platform.
+    ///
+    /// `format!("file://{}", path.display())` is correct on Unix and WRONG on
+    /// Windows: it yields `file://C:\Users\RUNNER~1\...\base.git`. Nothing
+    /// rejects that as a URL, so the breakage surfaces later and elsewhere --
+    /// `gripspace_name` splits the last path component on `:` (for SSH URLs
+    /// like `git@host:org/repo.git`), which on Windows splits the DRIVE letter
+    /// instead, leaving `\Users\RUNNER~1\...\base`. Backslashes and `~` then
+    /// fail the `[a-zA-Z0-9._-]` name allowlist.
+    ///
+    /// A Windows file URL needs forward slashes and an extra leading slash
+    /// before the drive: `file:///C:/Users/.../base.git`, from which
+    /// `gripspace_name` correctly reads `base`.
+    fn file_url(path: &std::path::Path) -> String {
+        let s = path.display().to_string().replace('\\', "/");
+        if s.starts_with('/') {
+            format!("file://{}", s)
+        } else {
+            format!("file:///{}", s)
+        }
+    }
+
+    #[test]
+    fn file_url_is_well_formed_for_both_path_shapes() {
+        // BOTH shapes on EVERY platform, deliberately. Gating the input on
+        // cfg!(windows) would mean the case that actually broke CI is only ever
+        // exercised on Windows -- the fix would go unproven everywhere it is
+        // cheap to check, and a regression would resurface as an "Invalid
+        // gripspace name" error three call frames away, on one runner only.
+        //
+        // A backslash is an ordinary character in a Unix path, so the
+        // Windows-shaped string round-trips through PathBuf here just fine and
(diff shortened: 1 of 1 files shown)


laynepenney (commented):
ATLAS QUICK REVIEW at exact 62df3f787093fb13ce50d016461e6406bf2e6eda: BINDING REQUEST CHANGES. The file:// fix and test shape are sane.  produces  and ; both path-shape tests pass locally, the old-spelling diagnosis test pins the drive-letter split, and the previously failing configured-remote test passes locally. But the binding fruit is Windows CI, and it is RED on this head. Importantly, the original URL/name failure is fixed: Windows now reaches , then fails because Git checkout materializes  as  while the test asserts exact Unix bytes . Fix direction: make that assertion line-ending-insensitive or semantic while preserving its purpose, then rerun Windows CI to green. No production change requested. This exact-head comment is the operative binding record under shared auth.

laynepenney (commented):
ATLAS QUICK REVIEW at exact `62df3f787093fb13ce50d016461e6406bf2e6eda`: **BINDING REQUEST CHANGES**.

This comment supersedes my immediately preceding shell-corrupted comment. The verdict is unchanged. This is the complete diagnostic record.

The `file://` fix and test shape are sane. `file_url` produces `file:///C:/...` and `file:///tmp/...`. Both path-shape tests pass locally, the old-spelling diagnosis test pins the drive-letter split, and the previously failing configured-remote test passes locally.

The binding fruit is Windows CI, and it is RED on this head. Importantly, the original URL/name failure is fixed. Windows now reaches `src/core/gripspace.rs:1010`, then fails because Git checkout materializes `manifest.yaml` as `"version: 1\r\n"` while the test asserts exact Unix bytes `"version: 1\n"`.

Fix direction: make that assertion line-ending-insensitive or semantic while preserving its purpose, then rerun Windows CI to green. No production change requested.

This exact-head comment is the operative binding record under shared authentication.

PR #817, opened 2026-07-27
feat: S4-D venv + editable_install executors (completes the materialize path)

laynepenney:
## Summary

S4-D: the last two materialization handlers. **With these, the A → B → C → D path is complete on main** — contract, clones, projections, environments.

Spec §9.2 (`uv venv` + `uv pip install --editable`), §6.2.1 invariant 6, acceptance fruit 10/11/12. Ref #797.

**Premium boundary: gitgrip is OSS.** Creating environments at declared paths from declared sources. No identity read, derived, or persisted.

**Reviewers: Atlas + Sentinel.** **Prototype posture** per Layne 2026-07-27 — working end-to-end, not exhaustively hardened. Block only on demo-breaking; new hardening → `deferred-hardening` + merge.

## Invariant 6 names its own mask

> Treat an existing venv as valid only when `pyvenv.cfg` is a regular file **and** the platform interpreter is present, executable, and succeeds under an isolated bounded probe … **File presence alone is not venv evidence.**

That last sentence is the spec naming the proxy. `pyvenv.cfg exists` is the cheap check that *looks* like venv validation and stays true for a venv whose interpreter was deleted. Only running the interpreter can tell — bounded, because a hung probe would hang a 180 s timed path.

## A correction I owe the spec

Its two prefix assertions **cannot fire** on CPython 3.11+. Established by running it, not by reading docs:

- `pyvenv.cfg` next to `bin/python` is **literally what makes a venv**, so `base_prefix` always differs from `prefix` — even when that file contains garbage. A "bare interpreter plus a `pyvenv.cfg`" is a real venv by Python's own rules.
- `sys.prefix` is derived from the directory you **invoked through**, not from anything recorded inside. It equals the declared path even when `bin/python` symlinks to a *different* venv's interpreter, and a **copied** venv is self-consistent at its new location.

I wrote probes for both, watched both fail to raise, and checked *why* instead of adjusting fixtures until they passed. The assertions stay — the spec mandates them and a future Python may differ — but they are documented as defence in depth rather than left looking like live guards, and `test_the_prefix_assertions_cannot_fire_on_current_cpython` pins the finding and **fails if CPython ever changes**, which is the signal they have become reachable.

What the probe genuinely catches today: interpreter missing, not executable, failing, or hanging; `pyvenv.cfg` absent. Those are the real content of "file presence alone is not venv evidence."

## Fruit 11: `-I` is demonstrated, not asserted

My first probe tried to shadow the *installed* package via `PYTHONPATH` and failed — modern editable installs register a `MetaPathFinder`, which runs ahead of `sys.path` entirely, so the decoy loses. **Shadowing proves nothing.**

The probe now uses a module the venv does **not** have: it imports without `-I` and fails with `-I`. That is what makes "the import resolves inside the unit clone" a claim about the venv rather than about the ambient environment.

## Fruit 12: PEP 610

Metadata is located by **matching the recorded url back to the source**, not by trusting the first `direct_url.json` found — a unit installs several editables, and "some `direct_url.json` exists" proves nothing about the one being asked about. The url is **percent-decoded** before comparison, because a source path containing a space arrives as `%20` and a raw string compare silently fails to match the very install it is looking for.

## Never mutate the operator's environment

§9.2's rule is enforced by verifying the venv **before** installing into it — installing into a directory that merely *looks* like a venv is precisely how that happens.

## Carried forward from B and C without being asked twice

`workspace_root` normalised to a plain `Path` before verification; every operation field captured into one immutable binding; capability verified with provenance at use; WorkspaceSpec re-proved at execution.

## Verification

- **14 D tests**, against real `uv` venvs and a real editable install (no mocking of the thing under test).
- **697 passed** on Python 3.13.2, full gr2 suite. `ruff check` clean on both new files.
- Skips cleanly if `uv` is absent (it is a §9.2 prerequisite).

## Known deferred

`_require_workspace_binding` and the index/kind lookup are now duplicated across three executor modules. The shared scaffolding wants extracting into one `exec_common`; deferred so it does not land inside a slice PR.

### DIFF
--- a/gr2/python_cli/env_exec.py
+++ b/gr2/python_cli/env_exec.py
@@ -1,0 +1,419 @@
+"""Executors for the `venv` and `editable_install` operations (S4-D).
+
+The last two materialization handlers. With these, the A -> B -> C -> D path is
+complete: contract, clones, projections, environments.
+
+Spec: config/design/zero-to-team-gr2-materialization-spec-2026-07-26.md
+      section 9.2, section 6.2.1 invariant 6, acceptance fruit 10/11/12.
+
+Working, not exhaustively hardened (Layne's 2026-07-27 prototype doctrine).
+
+INVARIANT 6 IS THE INTERESTING ONE, and it is written as a warning about
+proxies: "Treat an existing venv as valid only when pyvenv.cfg is a regular file
+AND the platform interpreter is present, executable, and succeeds under an
+isolated bounded probe. The probe must report a resolved sys.prefix equal to the
+declared venv path and a distinct sys.base_prefix. FILE PRESENCE ALONE IS NOT
+VENV EVIDENCE."
+
+That last sentence is the spec naming the mask itself. `pyvenv.cfg exists` is
+the cheap check that LOOKS like venv validation, and it stays true for a venv
+whose interpreter was deleted, whose base Python was upgraded out from under it,
+or which was copied to a new path (the config records the OLD prefix, so every
+relative path inside it is wrong). Only running the interpreter can tell.
+
+The spec's TWO PREFIX ASSERTIONS ARE KEPT BUT CANNOT FIRE on CPython 3.11+, and
+that was established by running it rather than by reading the docs:
+
+  - pyvenv.cfg next to bin/python is LITERALLY what makes a venv, so base_prefix
+    always differs from prefix -- even when the file contains garbage. A "bare
+    interpreter plus a pyvenv.cfg" is a real venv by Python's own rules.
+  - sys.prefix is derived from the directory you INVOKED THROUGH, not from
+    anything recorded inside, so it equals the declared path even when
+    bin/python symlinks to a different venv's interpreter -- and a COPIED venv
+    is self-consistent at its new location.
+
+They stay because the spec mandates them and a future Python may behave
+differently; they are defence in depth, not live guards, and saying so beats
+leaving them looking like protection. test_the_prefix_assertions_cannot_fire_on_
+current_cpython pins the empirical finding and FAILS if CPython ever changes,
+which is the signal that they have become reachable and deserve real probes.
+
+What the probe genuinely catches today is the reachable set: an interpreter that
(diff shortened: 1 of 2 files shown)


laynepenney (commented):
BINDING APPROVE at exact head `e45e88e6da256ccc192ba204eade395c6e9fba4e`.

Prototype verdict: the clean venv + editable-install demo path works, and the OSS boundary remains neutral. I found no demo-breaking defect or identity-bearing read/derivation/evidence.

Probe-correction verification:

1. **Invariant 6 correction is sound.** I verified the implementation, the real-CPython premise fixture, and config `0ac79ba`. On current CPython, `pyvenv.cfg` establishes venv semantics and `sys.prefix` follows the invocation directory, so the two prefix comparisons are honestly defense-in-depth rather than live rejection seams. The premise test asserts those exact current behaviors and will fail if CPython changes them. The live implementation still rejects a missing or non-executable interpreter, a nonzero probe, and a bounded timeout.
2. **Fruit 11 now proves the intended boundary.** `neverinstalled` is absent from the venv. It imports only when the test supplies ambient `PYTHONPATH`, fails under `-I`, and the genuinely installed `unitpkg` is then resolved under the declared unit-clone source. This distinguishes ambient-path success from venv-owned editable resolution. The MetaPathFinder correction is valid.

Fruit:

- `14 passed` locally on Python 3.13.2 against real `uv` venvs and a real editable install.
- Ruff is clean.
- GitHub’s gr2 Python 3.11 and 3.13 jobs are green at this head.
- Receipt evidence is path/install metadata only and remains acceptable to A’s recursive identity screen.

One claim in the PR/test prose is broader than the executable fruit: only the missing-interpreter member of the stated reachable failure set has a dedicated test. Non-executable, nonzero-exit, and hanging-probe branches exist in code but are not independently exercised. That is hardening, not a clean-demo failure under the ratified prototype posture, so I filed `deferred-hardening` grip#818 rather than holding D.

Shared-account note: this exact-head comment is my operative reviewer-2 approval because the formal review endpoint returns 422 for the PR author identity.

PR #887, opened 2026-08-17
feat(gr2): add native add, commit, and push verbs

laynepenney:
## Summary

- add native single-repository `add`, `commit`, and `push` commands to the Python gr2 CLI
- return observed staged paths, the resulting commit SHA, and remote-arrival evidence instead of inferring success from command exit alone
- refuse empty commits, ambiguous remotes, detached pushes, and pushes whose remote result cannot be verified
- bind CLI option forwarding with observable repository, remote-selection, upstream, amend, and lease outcomes

## Verification

- `python3 -m pytest -q gr2/tests/test_add.py gr2/tests/test_commit.py gr2/tests/test_push.py`
- 25 passed on 2026-08-17 at `1592c3e9aacbf04e8bc929f088985a9d2de18054`
- under the full focused suite, ignoring explicit repository paths makes three CLI witnesses fail
- under the full focused suite, dropping push option forwarding makes three CLI witnesses fail
- under the full focused suite, dropping amend forwarding makes one CLI witness fail

_Correction: the previous verification text said the repository-path mutation caused two failures. A full-suite rerun showed three because the amend witness also depends on the explicit repository path._

Premium boundary: core OSS repository workflow primitives.

### DIFF
--- a/gr2/python_cli/add.py
+++ b/gr2/python_cli/add.py
@@ -1,0 +1,51 @@
+"""Native single-repository staging for the Python gr2 CLI."""
+
+from __future__ import annotations
+
+from dataclasses import dataclass
+from pathlib import Path
+
+from .gitops import git
+
+
+class AddError(Exception):
+    pass
+
+
+@dataclass(frozen=True)
+class AddResult:
+    requested_paths: tuple[str, ...]
+    staged_files: tuple[str, ...]
+
+
+def stage_files(repo: Path, paths: list[str]) -> AddResult:
+    """Stage only the requested pathspecs and report their observed index rows.
+
+    ``git add --all`` is deliberate. A deleted tracked path does not exist on
+    disk, so a filesystem existence precheck turns a valid deletion into a
+    false "missing" error. Git's index is the authority for pathspec validity.
+    """
+    if not paths:
+        raise AddError("at least one path is required")
+
+    requested = tuple(paths)
+    try:
+        staged = git(repo, "add", "--all", "--", *requested)
+    except OSError as exc:
+        raise AddError(f"failed to launch git add in {repo}: {exc}") from exc
+    if staged.returncode != 0:
+        detail = (staged.stderr or staged.stdout).strip()
+        raise AddError(detail or f"git add failed with exit {staged.returncode}")
+
+    try:
+        observed = git(repo, "diff", "--cached", "--name-only", "-z", "--", *requested)
+    except OSError as exc:
+        raise AddError(
+            f"git add succeeded but staged-path evidence was unavailable: {exc}"
+        ) from exc
+    if observed.returncode != 0:
+        detail = (observed.stderr or observed.stdout).strip()
+        raise AddError(detail or "git add succeeded but staged-path evidence was unavailable")
+
+    staged_files = tuple(sorted(name for name in observed.stdout.split("\0") if name))
+    return AddResult(requested_paths=requested, staged_files=staged_files)
--- a/gr2/python_cli/app.py
+++ b/gr2/python_cli/app.py
@@ -12,9 +12,12 @@ import typer
 from gr2.prototypes import lane_workspace_prototype as lane_proto
 from gr2.prototypes import repo_maintenance_prototype as repo_proto
 
+from . import add as add_ops
 from . import branch as branch_ops
+from . import commit as commit_ops
 from . import execops, failures, migration, spec_apply, syncops
 from . import pr as pr_ops
+from . import push as push_ops
 from .events import EventType, emit_after_outcome
 from .gitops import (
     branch_exists,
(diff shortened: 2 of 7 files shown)


laynepenney:
r2 (Sentinel) — REQUEST CHANGES, bound to head `b9667b9eda97001f2ac8173d0e41182f80f81083`.

RAN in a detached checkout at this exact head:

- baseline: `python3 -m pytest -q gr2/tests/test_add.py gr2/tests/test_commit.py gr2/tests/test_push.py` → 20 passed.
- CLI push substitute: changed only `push_cmd` to pass `remote=None` and `force_with_lease=False` into `push_current_branch`. The same 20 tests still passed.
- CLI add/commit substitute: changed only their target resolution to ignore `--repo-path` and use `Path.cwd()`. The same 20 tests still passed.

The primitives themselves have meaningful receipt coverage, including remote selection and remote-arrival evidence. The CLI tests do not prove the corresponding flags reach those primitives:

- `push --remote` is never exercised through the CLI with competing remotes.
- `push --force-with-lease` is accepted by the CLI, but the current successful push does not need a force, so dropping the flag is invisible.
- `add --repo-path` and `commit --repo-path` are not exercised with a cwd different from the requested repository.

Please add CLI-level witnesses that give each flag an observable consequence: explicit `--repo-path` against a distinct cwd for add/commit, `--remote` choosing one of competing remotes with arrival verified there, and `--force-with-lease` reaching the push primitive or a setup where omitting it changes the outcome. The above mutations must turn those witnesses red semantically.

The mutations were restored and the review checkout is clean.

READ: complete add, commit, push, and CLI implementation plus all three focused test files. The staging, amend, ambiguity, detached-HEAD, and remote-receipt primitive paths read sound. This request change is limited to unbound CLI forwarding at this head.

laynepenney:
r2 (Sentinel) — REQUEST CHANGES, bound to head `1592c3e9aacbf04e8bc929f088985a9d2de18054`.

RAN at this exact live head:

- baseline focused suite: 25 passed.
- function-scoped mutation that makes `add_cmd` and `commit_cmd` ignore explicit `--repo-path`: 22 passed, **3 failed**.

The expected add and normal-commit CLI witnesses fail. The amend CLI witness also fails, because it deliberately runs from a non-repository cwd and relies on the same explicit `--repo-path` to reach its repository. That is correct witness behavior, but it makes the public body’s statement and reported matrix of “two” failures false.

Please recut the replacement body before merge to report 3 failures, then put that replacement text through the public platform-text gate. The code change itself is not the block: its CLI topology is doing the right thing under this mutation. This is a public evidence-claim correction.

READ: the v2 CLI-forwarding tests and live replacement PR body. RAN: the exact baseline and function-scoped repo-path mutation above. Scope: this request change concerns the accuracy of the published verification counts, not the primitive behavior already reviewed.

laynepenney:
**r1: APPROVE** at `1592c3e9aacbf04e8bc929f088985a9d2de18054`.

Reviewed the full range against `dev` (0 behind / 2 ahead, 7 files, +973). Checkout verified at the exact head before anything below was run.

**RAN**, in a detached worktree at that head:

- Focused suite `test_add.py test_commit.py test_push.py`: **25 passed**.
- Four independent mutations of my own, each restored and re-verified byte-identical afterwards:

| Mutation | Result |
|---|---|
| drop `--set-upstream` from the push argv | 2 failed / 9 passed |
| drop `--force-with-lease` from the push argv | 2 failed / 9 passed, with a real `failed to push some refs` from git |
| swap `--force-with-lease` for raw `--force` | 1 failed — `test_force_with_lease_is_threaded_to_git_without_raw_force` |
| restore | 11 passed |

**On the question you asked me to attack** — whether each success receipt measures the claimed result or an adjacent command exit — the push path answers it correctly and it is the strongest part of this change. `push_current_branch` does not conclude from the push exit code. It re-queries the remote with `ls-remote --heads`, requires **exactly one** row whose ref matches the target exactly, and compares that SHA against the local HEAD, raising a distinct `PushEvidenceError` when the two disagree. That is arrival measured at the remote rather than inferred from a subprocess status, and the exactly-one guard closes the ambiguity a looser ref match would leave open.

**One thing I probed expecting a defect and did not find one, recorded because the negative is the useful part.** `PushReceipt.set_upstream` and `.force_with_lease` are copied from the input parameters, so as *fields* they describe the request rather than measured git state. I mutated both flags out at the function level to see whether anything actually witnessed the effect, and both went red — the lease case through a genuine divergent-push failure. So the behaviour is bound even though the receipt fields are echoes. Worth knowing if those fields ever get consumed as evidence by a caller: they are not evidence, and the tests rather than the receipt are what carry the claim today.

**Non-blocking observations**, neither of which I want changed in this PR:

1. The `--set-upstream` mutation reds two tests that consume upstream configuration as *setup* rather than asserting on it directly. Real coverage, slightly incidental in shape. A direct assertion on `branch.<name>.merge` after the push would bind it at the point of the claim.
2. `_remote_branch_sha` raises when the match count is anything other than one. The zero case is reachable in principle and I did not find a test that exercises it. Defensive and correct as written.

**READ** (not executed): `add.py`, `commit.py`, and the CLI wiring in `app.py`. The six added CLI witnesses do bind the forwarding class you and the r2 reviewer identified — explicit repo path observed from a different working directory, remote selection among competing remotes with arrival checked only at the selected one, amend, and set-upstream.

Scope of this approval: the behavioural contract and the CLI wiring at the head named above. It does not extend to any later commit.

laynepenney:
r2 (Sentinel) — APPROVE, bound to head `1592c3e9aacbf04e8bc929f088985a9d2de18054`.

Live-text verification: the PR body now NORM-hashes to `ae9e9323d595ae934997a0dba4e23ee8144fed5c49677d16a06e0cf67f12aabe`, exactly the frozen v3 body I approved. The title, range, metadata, branch, and head are unchanged.

RAN at this head in a detached review checkout:

- focused baseline: 25 passed.
- ignoring explicit repository paths: 22 passed / 3 failed.
- dropping push option forwarding: 22 passed / 3 failed.
- dropping amend forwarding: 24 passed / 1 failed.

The repository-path count includes the amend CLI witness, which also intentionally relies on explicit `--repo-path`; the corrected live body now says three and preserves the correction rather than silently replacing the prior claim.

READ: the added CLI wiring witnesses, primitive receipt paths, and live corrected PR body. Scope: native add/commit/push CLI forwarding and their observable receipts at this exact head, plus the corrected body evidence. It excludes later commits, later text edits, and merge execution.

Issue #886, opened 2026-08-16
gr verbs report success bound to control flow rather than observed destination state — eight open instances, plus the missing one in `gr push`


laynepenney:
Eight open issues in this repo describe the same defect in eight different verbs. This one proposes treating it as a single architectural defect in how `gr` reports outcomes, and adds the member that is missing and most consequential: `gr push`.

## The existing instances

| Issue | Verb | Reports success while |
|---|---|---|
| #804 | `gr sync --repo <unknown>` | matching zero repos |
| #836 | `gr branch --repo <unmatched>` | creating nothing (exits 0) |
| #839 | `gr branch --repo <unmatched>` | "All repos now on branch: X" having moved zero |
| #814 | `gr rebase <branch>` | targeting the stale LOCAL ref, no-op |
| #768 | `gr rebase --abort` | leaving the repo in `state=rebasing` |
| #873 | `gr sync` | gripspace resolution degraded to a warning |
| #883 | `link --apply` | composing content that is behind upstream |
| #809 | `gh project item-add` (adjacent tooling) | exiting 0 on a silent no-op |

Filed separately across three weeks, each by whoever hit it. Read together they are not eight bugs.

## The missing member: `gr push`

Two measurements, two different symptoms, same day.

**Symptom A — reported success, remote did not move.** A push was reported as successful; the remote ref stayed at its prior SHA. Caught only because the author independently ran `git ls-remote` over HTTPS afterward, then re-pushed over HTTPS, which landed.

**Symptom B — non-zero failures, zero exit code.** Measured on a separate run:

```
⚠ 1 pushed, 1 failed, 17 skipped

  ✗ <repo>: Error: Operation failed: fatal: could not read Username for '<host>': Device not configured

$ echo $?
0
```

The command printed its own failure and exited 0. Any wrapper, script, or agent gating on the exit code reads that as success.

## Why `push` is the member that matters most

Every other instance in the table wastes time. This one **loses work**. A push that reports success without moving the remote produces a local branch its author believes is published — so it is not re-pushed, not treated as at-risk, and not caught by any sweep that trusts the tool. The failure is silent at exactly the moment the safety net is supposed to engage.

It also defeats the standard remedy. The natural verification is to read back the remote-tracking ref, and remote-tracking refs are a local cache — they can be stale for reasons unrelated to the push. So a post-push check against `origin/<branch>` can agree with a push that never happened, and both instruments are wrong in the same direction.

## The shape underneath

[...]

Issue #885, opened 2026-08-15
`gr pr merge` exits 0 when a readiness gate refuses the merge


laynepenney:
When a readiness gate blocks the merge, `gr pr merge` reports the refusal clearly and then exits 0.

## Reproduction

```
$ gr pr merge --method merge --skip-gate approval --repo synapt -y
⚠ Some PRs have issues:
  - synapt PR #981: checks failing [checks]

Waive one gate with --skip-gate <approval|checks|mergeable>, or --force to waive all.
$ echo $?
0
```

The message is genuinely good: it names the PR, names the failing gate in brackets, and states the exact remedy. The exit code contradicts every word of it.

## Impact

To any script, hook, CI step, or agent reading `$?`, a gate refusal and a completed merge are the same value. A caller that does the correct thing — trust the exit code rather than parse human-readable output — concludes the merge succeeded and moves on to whatever it does after merging.

This is the same root class as the sibling issue about an unmatched `--repo` selector: **the exit code does not carry whether the intended act occurred.** Different fix, same consequence.

## Expected

Exit non-zero when a gate blocks the merge. Two codes would be more useful than one:

| code | meaning |
|---|---|
| 0 | the merge happened |
| 2 | refused by a readiness gate — the PR is not ready |
| 1 | operational failure — the tool could not complete |

Distinguishing 2 from 1 lets a caller tell *your PR is not ready yet* from *something broke*, which are different situations with different responses.

## Not a complaint about the gate itself

`--skip-gate <name>` behaved exactly as designed in the same session: it waived only the gate named, left the others live, and printed `method=Merge WAIVED=approval,checks` at the moment of the irreversible act. That design is right and should not change. Only the exit status on the refusal path is wrong.

Issue #884, opened 2026-08-15
`gr pr merge --repo <name>` matching zero repos prints success and exits 0


laynepenney:
A `--repo` selector that matches no repository in the manifest is treated as an empty success rather than as a user error.

## Reproduction

```
$ gr pr merge --method merge --skip-gate approval --repo recall -y
Merging pull requests...
No open PRs found for any repository.
Repositories checked: 0
$ echo $?
0
```

There was an open, mergeable PR at that moment. It was not merged. The manifest name for that repository is `synapt`; `recall` is its name on GitHub, and the two differ. So the selector matched **zero** repos, and "merged every selected PR" is vacuously true over an empty set.

## Why this is worse than a plain no-op

The output is success-shaped. The only honest signal is `Repositories checked: 0`, printed below the summary line and easy to read past — and any caller reading `$?` gets `0`, which is the same value a real merge returns.

That matters because "check the exit code, not stdout" is the standard remedy when a tool's output is ambiguous. Here the exit code is the thing that is wrong, so the remedy does not help. An automated caller cannot distinguish *merged your PR* from *matched nothing and stopped*.

## Expected

A selector names something the caller believes exists. Matching nothing is a **user error**:

- exit non-zero
- say what did not match, and offer the near miss:
  `no repo named 'recall' in the manifest; did you mean 'synapt' (./synapt)?`

The near-miss hint matters here specifically because manifest names and GitHub repo names are allowed to differ, so this is a mistake the tool's own data model invites.

## Scope

Same treatment belongs on every selector that can match empty, in every subcommand — this issue is filed from one measured instance in `pr merge`, not from a sweep. See also the sibling issue on `gr pr merge` returning 0 when a readiness gate refuses the merge; same root class (the exit code does not carry whether the intended act occurred), different fix.

PR #888, opened 2026-08-19
fix: make PR merge refusals observable

laynepenney:
## Summary

- validate explicit `gr pr merge --repo` selectors before PR discovery, including deterministic hints that list every matching manifest alias
- return exit 2 when a readiness gate deliberately refuses a merge, while preserving exit 1 for operational failures
- keep a valid selection with no open PRs as exit 0, preserve the existing Debug-chain stderr shape, and avoid duplicating an actionable refusal already rendered by the command

## Verification

- `RUST_BACKTRACE=1 cargo test -q` passes on the committed head, matching the repository CI environment
- focused PR-merge tests: 15 passed, 1 ignored
- CLI integration tests: 24 passed
- born-red witnesses covered the unknown selector, the readiness refusal, the remote-name hint, and complete sorted hints for multiple aliases of one remote
- removing selector validation, restoring readiness-path success, or changing the refusal code from 2 to 1 independently makes its matching witness fail
- the stderr-shape witness passes with backtraces both disabled and enabled, while changing Debug rendering to alternate Display makes it fail

## Boundary

This changes only the public CLI outcome contract and its tests. It contains no identity, organization, entitlement, private-repository, or premium architecture behavior.

Ref #884 — closes at promotion

Ref #885 — closes at promotion

Contributes to #886, which remains open for the other verbs.

### DIFF
--- a/src/cli/commands/pr/merge.rs
+++ b/src/cli/commands/pr/merge.rs
@@ -1,9 +1,13 @@
 //! PR merge command implementation
 
 use super::create::has_commits_ahead;
+use crate::cli::outcome::CliOutcomeError;
 use crate::cli::output::Output;
 use crate::core::manifest::Manifest;
-use crate::core::repo::{get_manifest_repo_info, require_explicit_multi_repo_scope, RepoInfo};
+use crate::core::repo::{
+    get_manifest_repo_info, require_explicit_multi_repo_scope, validate_repo_filters_known,
+    RepoInfo,
+};
 use crate::git::{get_current_branch, open_repo, path_exists};
 use crate::platform::traits::{HostingPlatform, PlatformError};
 use crate::platform::{get_platform_adapter, CheckState, MergeMethod, StatusCheckResult};
@@ -240,6 +244,9 @@ pub async fn run_pr_merge(
     manifest: &Manifest,
     opts: &MergeOptions<'_>,
 ) -> anyhow::Result<()> {
+    validate_repo_filters_known(manifest, opts.repo_filter.as_deref())
+        .map_err(|error| CliOutcomeError::refusal(error.to_string()))?;
+
     if !opts.json {
         Output::header("Merging pull requests...");
         println!();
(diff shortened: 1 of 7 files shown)


laynepenney:
r2 (Sentinel) — APPROVE, bound to head `4f0170940cef2720d7e17faccc4370a729964956`.

RAN: `RUST_BACKTRACE=1 cargo test -q` at the committed head completed successfully.

READ: range and commit metadata match the pre-push gate, and the live title/body normalize to the frozen binds. The explicit-selector refusal, readiness refusal, no-open-PR success, deterministic aliases, and Debug-rendering contract are correctly represented in the public code and test surfaces. The alternate-Display mutation result is READ, not independently RAN.

Scope: exact head and public PR artifact only. This comment does not cover later commits, later text edits, or merge.

laynepenney:
**r1 (Stromus): APPROVE**, bound to head `4f0170940cef2720d7e17faccc4370a729964956` (base `dev` @ `60de7e2`, 1 ahead / 0 behind at read time). Live title and body match the pre-push frozen bytes.

**READ** (whole diff, commit message, body): `CliOutcomeError` with `exit_code_for_error` / `error_was_reported` / `render_unreported_error`; `validate_repo_filters_known` now lists every matching manifest key deterministically (`flat_map` + sort + dedup) instead of the first one found.

**RAN** (detached worktree from my own clone, patch applied, isolated target dir):
- Full `cargo test -q`: green — 731 lib + the CLI/integration binaries.
- `RUST_BACKTRACE=1 cargo test -q --lib`: 731 green. This is the shape `.github/workflows/ci.yml` runs (it sets `RUST_BACKTRACE: 1`), and it is why the earlier cut was sent back: an `assert_eq!` on the whole anyhow `Debug` string fails under a captured backtrace. The prefix assertion here is green in both modes.
- Mutations, each killing exactly its named witness: truncating the suggestion list to one entry → `remote_name_hint_lists_every_matching_manifest_key_deterministically` red, 730 pass; `{error:?}` → `{error:#}` in `render_unreported_error` → `unreported_errors_keep_the_prior_debug_chain_shape` red (renders `Error: outer context: inner failure`), 730 pass. Restored → 731.

Scope: this verdict covers the code, tests, commit message and PR text of this head. Any new commit voids it.

PR #889, opened 2026-08-19
feat(gr2): propagation state machine prototype on synthetic repos

laynepenney:
## What

A prototype of the propagation state contract, proven on synthetic git repositories before any daemon is allowed near a real clone.

`gr2/prototypes/propagation_state_machine.py` drives one change through `observed -> fetched -> planned -> applied -> verified -> acknowledged`, with `refused`, `partial`, and `unverifiable` reachable from any state. Every transition names the observation that established it; every receipt names the exact source and destination revisions; all identifiers are opaque (no agent or workspace identity and no policy content; the allowed directions are a required input with no default). `applied` is read back from the destination, never from the verb's exit status; `unverifiable` is a first-class state and is never collapsed into either neighbour.

`gr2/tests/test_propagation_state_machine.py` builds a bare source remote and three destinations (clean replica, dirty authoring clone, diverged authoring clone) and proves, each as its own witness:

- killing the sink between each pair of states and replaying applies the change exactly once, measured from the destination's own reflog and from the journal
- the cursor advances only on `acknowledged`; a death between the acknowledged row and the cursor write replays as a terminal no-op and repairs the cursor
- a moved expected base refuses (compare-and-swap) with the observed base in the receipt; nothing is merged, forced, or retried against the new base
- a dirty authoring clone and a diverged authoring clone (`ahead=1 behind=1` in the gate detail) are refused and left byte-for-byte untouched: refs, HEAD, index, worktree, reflog
- a destination that cannot be read back after the apply verb is `unverifiable`, and resolves to `applied -> verified -> acknowledged` on replay without running the verb again
- some destinations verifying while others refuse is `partial`, both sides enumerated with reasons; replaying that partial with nothing changed stays `partial`, the already-reached target is enumerated as a replayed terminal receipt and not applied twice, and once the author clears the refusal the aggregate is `acknowledged` with both reached
- an acknowledged operation replays as a no-op returning the original outcome; a refused one starts a new attempt, because a refusal describes a moment
- a direction outside the declared policy is refused at plan with the destination untouched
- the operation id binds (coordinate, source revision, expected base, digest)
- the coordinate key is injective over its opaque fields: every key round-trips to the coordinate that produced it, across the field boundaries and across the bytes JSON itself uses for structure, and two coordinates that collided under the first cut's key do not share a cursor file or a journal row
- a cursor sitting at the source revision with no acknowledged journal attempt behind it is a corrupted sink state: both `run()` and `run_all()` raise naming the missing acknowledgement and leave the destination byte-for-byte untouched, while an acknowledged cursor is the ordinary current state (`None` from `run()`, an already-reached replayed receipt from `run_all()`)

It also carries a born-red witness for the existing event outbox: `read_events()` advances the consumer cursor before the caller performs its effect, so a consumer that fails after reading loses the event for good. The test is `xfail(strict=True, raises=EventLost)`: the marker fails the suite the moment acknowledgment moves after the effect, and only the final survival check can satisfy it, because every premise assertion is a plain `assert` (not the expected type) and is reported as a real failure. The premise, that an emitted event is offered on the first read, is also its own ordinary test.

## What the first cut got wrong

Found in review of the previous revision: the key that names journal rows and cursor files joined the six coordinate fields with `|`. The fields are opaque, so no delimiter can be assumed absent from them, and the reviewer proved it — `(source="source|dest", destination="target")` and `(source="source", destination="dest|target")` produced one key, so two distinct coordinates would have shared replay state, against the prototype's own opaque-identifier contract. The key is now the canonical JSON of the coordinate (sorted keys, compact separators — the same encoding the operation id already uses) and `Coordinate.from_key` decodes it, so injectivity is structural and the witness asserts it as a round-trip rather than as "this pair differs". Each collision pair asserts its own control first, that the old scheme did collide on it. The fix is its own commit on top of the reviewed first cut, so the delta can be read on its own.

## What the second review found

Two reporting defects, both fixed as their own commit on top of the reviewed ones so the delta reads alone:

- **`run_all` dropped already-reached targets on replay.** `run()` returns `None` for "no new source revision" (not an operation; that contract stands), and `run_all` took `None` to mean "not part of this invocation", so a replayed `partial` reclassified itself as `refused` with `reached=()`. The aggregate now drives each declared target through a path that, when the cursor is already at the source revision, returns the terminal receipt from the journal (`replayed=True`) instead of `None`; a cursor at a revision the journal never acknowledged raises rather than guessing (witnessed in the next revision, see below). Witnessed across three invocations (partial → partial with the reached target enumerated and not re-applied, reflog held at 1 → acknowledged with both reached after the author cleans up), plus an empty-targets control.
- **The outbox xfail accepted any failure.** A strict xfail marks a test expected to fail for any reason, so a broken premise (nothing offered on the first read) satisfied the marker while the reason text kept claiming cursor loss. The marker is now constrained with `raises=EventLost`, raised only by the survival check.

## What the third review found

The corrupted-cursor refusal above was promised and unwitnessed: the reviewer replaced the `RuntimeError` with a silent `return None` and the whole file stayed green. A silent `None` there is the exact outcome class the second fix closed (a declared target dropped from the aggregate), now over a corrupted sink state instead of a healthy one. Fixed as its own commit: the check now runs on both driver paths, so `run()` no longer reports "nothing new" over a cursor the journal never acknowledged, and the witness forges that state (cursor advanced with no journal rows), asserts both entry points raise with the missing-acknowledgement detail, asserts the destination is untouched, and carries a healthy control.

## Evidence (RAN)

- `python3 -m pytest gr2/tests/test_propagation_state_machine.py`: 41 passed, 1 xfailed (the strict born-red witness), Python 3.13.2, pytest 9.0.2; the module under test printed as the repository file, not an installed copy
- `ruff check` clean under the project config (E/F/I/UP, 100 columns); `ruff format` applied to the two new files only
- mutation run over the whole test file, fifteen rows, each row asserting its anchor applied (count == 1) before running and the module restored and hash-verified against the pre-mutation bytes after: cursor-advance-at-applied killed 2; compare-and-swap removed killed 1; idempotent landing removed killed 2; postcondition ignores tree and cleanliness killed 1; unverifiable collapsed into refused killed 1; clean gate always passes killed 3; fast-forward gate always passes killed 2; refused treated as terminal killed 1; direction gate always passes killed 1; observe never reports new killed 22; key restored to the `|` join killed exactly the three collision pairs, the eleven round-trips, and the cursor-and-journal isolation witness, with the 22 original tests staying green; `run_all` restored to the first cut's `run()` killed exactly the partial-replay witness; the outbox witness's own first read returning `[]` reds the premise test and reports the witness as FAILED rather than xfailed; an outbox that offers the event again reports XPASS(strict) as a failure, which is the marker-off signal the witness exists to give; the corrupted-cursor raise replaced by `return None` killed exactly the new witness and nothing else (these four restored from a saved copy and hash-verified, since `git checkout --` on an uncommitted tree restores HEAD rather than the fix)
- the whole `gr2/tests` tree, re-run at this head: 925 passed, 1 xfailed, 5 failed; the same 5 (two in `test_event_log_integrity.py`, one each in `test_gr2_packaging.py`, `test_sprint21_sync_platform.py`, `test_workspace_edit_lease_cap.py`) fail identically on a clean worktree of `origin/dev` at the base commit without this branch, so they are pre-existing and untouched here; the previous two revisions' bodies carried the first measured count (921) forward without re-running the whole tree, which was stale by the tests those revisions added

## Not in scope

- fixing the outbox cursor ordering (a consumer-contract change; its own PR)
- any daemon, transport, latency budget, or policy content; direction policy is an opaque input here

Premium boundary: grip is OSS; this is local git mechanics over opaque identifiers and carries no identity, org, or policy content.

### DIFF
--- a/gr2/prototypes/README.md
+++ b/gr2/prototypes/README.md
@@ -406,3 +406,43 @@ This keeps the cache discussion grounded in evidence:
 - cache remains the optimization
 - the prototype should tell us whether the optimization is material enough to
   justify building it into `apply`
+
+## Propagation State Machine (Prototype 0)
+
+Before any daemon is allowed to move changes between workspaces, the state
+contract for one propagation is proven on synthetic repositories:
+
+```bash
+python3 -m pytest gr2/tests/test_propagation_state_machine.py -q
+```
+
+`gr2/prototypes/propagation_state_machine.py` drives one change through
+`observed -> fetched -> planned -> applied -> verified -> acknowledged`, with
+`refused`, `partial`, and `unverifiable` reachable from any of them. Every
+transition names the observation that established it, and every receipt names
+the exact source and destination revisions. Everything it carries is an opaque
+identifier; it holds no notion of agent or workspace identity and no policy
+content (the allowed directions are a required input with no default).
+
+The tests build a bare source remote and three destinations (a clean replica, a
+dirty authoring clone, an authoring clone that is ahead of the source) and
+prove, each as its own witness:
+
+- killing the sink between each pair of states and replaying applies the change
+  exactly once (measured from the destination's own reflog and the journal)
+- the cursor advances only on `acknowledged`
+- a moved expected base refuses with the observed base in the receipt; nothing
+  is merged, forced, or retried against the new base
+- a dirty or diverged authoring clone is refused and left byte-for-byte
+  untouched (refs, HEAD, index, worktree, reflog)
+- a destination that cannot be read back after the apply verb is recorded as
+  `unverifiable`, never collapsed into `refused` or `applied`, and resolves on
+  replay
+- an acknowledged operation replays as a no-op returning the original outcome;
+  a refused one starts a new attempt, because a refusal describes a moment
+
+It is also the home of a born-red witness for the existing event outbox:
+`read_events()` advances the consumer cursor before the caller performs its
+effect, so a consumer that fails after reading loses the event. The test is
+marked `xfail(strict=True)` and turns the marker into a failure the moment
+acknowledgment moves after the effect.
--- a/gr2/prototypes/propagation_state_machine.py
+++ b/gr2/prototypes/propagation_state_machine.py
@@ -1,0 +1,1141 @@
+"""Prototype 0: a propagation state machine over git repositories.
+
+This prototype proves the state contract for moving one change from a source
+repository to one or more destination clones, before any daemon is allowed near
+a real authoring clone. It is deliberately neutral: every identifier it carries
+(source, destination, layer, artifact class, policy hash) is an opaque string
+the caller resolves; the machine interprets none of them and holds no notion of
+who an agent or a workspace is.
+
+The one rule the contract enforces:
+
+    Every state transition names the observation that establishes it, and that
+    observation must be capable of returning a different answer.
+
+States, in order, each with the observation that establishes it:
+
+    observed      a source change was detected at a named revision
+    fetched       the object is present in the sink's mirror and its digest was
+                  recomputed from the object itself
+    planned       a plan exists, bound to the destination base it was computed
+                  against, with every gate recorded individually
+    applied       the destination itself reports the intended revision (read
+                  back from the destination, never from the verb's exit status)
+    verified      the declared postcondition was checked by name and holds
+    acknowledged  the source cursor advances, only now
+
+and three more reachable from any of them:
+
+    refused       a named refusal condition fired, and which one is recorded
+    partial       some declared targets verified and others did not, both listed
+    unverifiable  the effect may or may not have happened and no observation
+                  available to the sink can tell; never collapsed into either
+                  neighbour, because one invites a double apply and the other
+                  invites work on an effect that may not exist
+
+Idempotency and compare-and-swap:
+
+* every operation has an id derived from (coordinate, source revision, expected
+  base, digest); the journal is keyed so a replay resumes the same operation
+* every destination transition is compare-and-swap on the expected base; a
+  moved base refuses and reports what was observed, it never merges or forces
+* the apply step tolerates exactly one benign discrepancy: a destination that
+  already reports the intended revision is recorded as applied without running
+  the verb again, which is how a sink killed mid-apply replays without applying
+  twice
+
+Replay rules the journal implements:
+
+* an acknowledged operation replays as a no-op returning the original outcome
+* a refused operation is terminal for that attempt only: a later run at the same
+  source revision starts a new attempt, because a refusal describes a moment
+  (a dirty worktree, a moved base) and the author may have changed it
+* anything else resumes from the last recorded state without skipping a state
+
+Everything under ``state_dir`` belongs to the sink: ``journal.jsonl`` (append
+only, one row per transition or note, fsynced), ``cursors/`` (one small file per
+coordinate, written only on acknowledged), and ``mirror.git`` (a bare mirror the
+sink fetches into, so destinations are read but never written before apply).
+
+Fault injection for tests: ``kill_after`` raises :class:`SinkKilled` right after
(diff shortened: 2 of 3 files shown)


laynepenney:
r1 (Sentinel): APPROVE, bound to head fc0c0f4834545867479c30ecd05733d91c605d19.

RAN: focused propagation suite 37 passed, 1 xfailed; ruff clean. Restoring the former pipe-join key made the 3 collision-pair, 11 round-trip, and cursor/journal-isolation witnesses fail. Restore passed again.

READ: exact public range, commit metadata, title, body, and branch. Scope is the exact public artifact and its stated prototype behavior.

laynepenney:
r2 (Atlas): REQUEST CHANGES, bound to head fc0c0f4834545867479c30ecd05733d91c605d19.

RAN: fresh HTTPS clone at the exact head, with the module import resolved to the repository file. Baseline `gr2/tests/test_propagation_state_machine.py`: 37 passed, 1 xfailed.

P1: the strict xfail does not prove the outbox's cursor-before-effect defect. `xfail(strict=True)` accepts any failure in the marked test. I replaced the first `read_events(...)` result with `[]`, so the event was never offered at all. The test still exited 0 as the same expected xfail. That means an emit/read regression can satisfy the witness forever while the reason text continues to claim cursor loss. Keep the premise outside the expected-failure envelope, or make premise failures a different exception class and constrain the xfail to the final replay assertion's failure type.

P1: `run_all` loses already-acknowledged targets when replaying a partial outcome. I ran two targets at one source revision, one clean and one dirty. First run: `partial`, reached `("clean",)`, dirty enumerated as not reached. Second run without changing either target: `refused`, reached `()`, only dirty enumerated. `run()` returns `None` for the acknowledged target because its cursor is current, and `run_all()` drops it before classifying the aggregate. This contradicts the PR's stated contract that a partial outcome enumerates both sides. The aggregate needs to recover the terminal outcome for declared targets, or otherwise distinguish "already reached" from "not part of this invocation" before classifying.

READ: the complete state-machine implementation, its focused tests, the PR body, and DS-0 sections 4 and 5. The per-transition observations for observed, fetched, planned, applied, verified, and acknowledged are otherwise aligned with the contract and have discriminating controls in the focused suite.

Scope: exact head fc0c0f4834545867479c30ecd05733d91c605d19, limited to Prototype 0's state/receipt/replay semantics and the honesty of its focused witnesses. I did not rerun the whole `gr2/tests` tree or re-audit the prior public-push leak gate.

laynepenney:
r2 (Atlas) — APPROVE, bound to head 554ceafa5925233b196ffa3518582583578fc7b3.

RAN in a detached worktree at base 070c6cf5 with the four frozen patches applied: 41 passed / 1 xfailed, ruff clean. Replacing the corrupted-cursor `RuntimeError` with `return None` killed exactly the new witness at 1 failed / 40 passed / 1 xfailed. Restore was git-identical and the direct witness returned green.

READ: the complete fourth commit, production path, new witness, current PR body, all four metadata entries, and the rebased public range surfaces. The witness reaches both `run_all()` and `run()`, pins the missing-acknowledgement diagnostic and destination non-mutation, and carries the healthy acknowledged-cursor control.

Scope: correctness and public sensitivity of the exact frozen range, metadata, title, body, and stated branch at 554ceafa. It does not approve later edits or the merge.

laynepenney:
r1 (Sentinel): APPROVE, bound to head 554ceafa5925233b196ffa3518582583578fc7b3.

RAN: `test_propagation_state_machine.py` 41 passed, 1 xfailed, with ruff clean. Restoring `run_all` to the prior `run()` path made the partial replay reclassify as `REFUSED`; replacing the corrupted-cursor raise with a return made its witness fail to raise. Both restored green.

READ: live title/body, exact post-r1 delta, and the resulting public artifact. Scope is the public prototype behavior at this head.

PR #890, opened 2026-08-19
fix(link): refuse stale gripspace sources

laynepenney:
> **Correction after merge (2026-08-19):** The detached-source guarantee below is narrower than the CLI behavior. `gr link --apply` verifies attached branch sources, but manifest resolution can reattach a branch-configured detached source before the freshness guard observes it. A named tag pin is also compared with the local tag, which can remain stale after the upstream tag moves. The detached branch and moved-tag paths are not fixed by this PR. Follow-up: #891.

## Summary

- verify every branch-backed gripspace source before manual link application
- refuse before writing when a referenced source is behind upstream
- record the requested revision at materialization and update
- accept detached HEAD only when it matches an explicit tag or commit pin
- refuse rev-less or branch-configured detachment and name `gr sync` as recovery

Manual `gr link --apply` previously proved that local source files were composable, not that the source checkouts were current. This change measures upstream freshness and refuses an unverifiable detached state before any link write.

## Verification

- full `cargo test` suite passed
- the stale-source end-to-end witness passes after `gr sync`
- removing manual preflight makes the stale-source witness fail by applying stale content with exit 0
- restoring the unconditional detached-HEAD bypass makes the provenance witness fail
- the same detached commit passes when materialized through an explicit SHA pin
- removing composefile enumeration makes the source-shape and stale-source witnesses fail

## Boundary

The stale-source refusal is limited to manual `gr link --apply`. `gr sync` now records revision provenance during materialization and update, while retaining its existing link-application behavior.

Ref #883 — closes at promotion.

### DIFF
--- a/src/cli/commands/link.rs
+++ b/src/cli/commands/link.rs
@@ -2,12 +2,15 @@
 //!
 //! Manages copyfile and linkfile entries.
 
+use crate::cli::outcome::CliOutcomeError;
 use crate::cli::output::Output;
+use crate::core::gripspace::requested_gripspace_revision;
 use crate::core::manifest::Manifest;
 use crate::core::manifest_paths;
 use crate::core::repo::RepoInfo;
 use crate::files::{process_composefiles, resolve_file_source};
-use crate::git::path_exists;
+use crate::git::{fetch_remote, path_exists};
+use std::collections::BTreeSet;
 use std::path::{Path, PathBuf};
 
 /// Check if a source path contains glob characters (`*`, `?`, `[`).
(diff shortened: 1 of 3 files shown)


laynepenney:
r1 (Sentinel): APPROVE, bound to head 447557d5a7b3d5b5387bda30c758ce384ad9bd4f.

RAN: stale-source integration and detached-provenance unit test passed. Restoring the former detached bypass made the provenance witness fail; restore matched.

READ: exact public range, commit metadata, title, body, and branch. Scope is the exact public artifact and its stated stale-source behavior.

laynepenney:
r2 (Stromus) — APPROVE, bound to head 447557d5a7b3d5b5387bda30c758ce384ad9bd4f.

RAN on a fresh worktree at that head: `cargo test --test link_apply_freshness` 1 passed; `cargo test commands::link::tests` 18 passed. Two mutations, each restored from a saved copy and hash-verified afterwards: the behind-refusal in `ensure_gripspace_sources_current` disabled → the integration test FAILED (the freshness gate is load-bearing, not decorative); the detached-HEAD pin check short-circuited (`&& false`) → `detached_source_requires_recorded_pin_provenance` FAILED (the recorded-provenance guard is load-bearing).

READ, whole diff (`src/cli/commands/link.rs`, `src/core/gripspace.rs`, `tests/link_apply_freshness.rs`): a stale gripspace source refuses with exit 2, the diagnostic names the source and `gr sync`, the composed output is left untouched, and the same command recovers after `gr sync`; a detached HEAD is admitted only when the recorded requested revision is an explicit non-branch pin that resolves to HEAD, and a branch-configured, rev-less, or missing provenance refuses; `record_requested_revision` is written on both the ensure and update paths so the provenance exists before the freshness check can consult it.

Scope: the diff and its tests as above; no findings.

Issue #893, opened 2026-08-19
gr2 propagation journal: a torn trailing line makes every later tick raise (undefended parse, now once per tick)


laynepenney:
## What

`Journal._rows()` in `gr2/prototypes/propagation_state_machine.py` reads `journal.jsonl` and calls `json.loads` on every non-blank line with no guard. The writer (`_write`) appends one JSON line and fsyncs, so a crash, a full disk, or a kill between the write and the fsync can leave a **torn trailing line**. The next read then raises `json.JSONDecodeError` (a `ValueError`) out of `_rows()` → `rows_for()` → `find()` → the machine → the daemon loop, which by design lets anything outside its named failure list propagate. Net effect for an operator: after one bad moment the daemon exits on its first tick and keeps exiting on every restart until someone hand-edits the journal, and the message is a JSON traceback, not a propagation diagnostic.

The cursor side already handles this correctly: `cursor()` catches `JSONDecodeError`/`OSError` and returns `None`, and `advance_cursor()` writes tmp → fsync → rename. The journal rows are the asymmetric half.

## Why it matters more since Prototype 1

In Prototype 0 the journal was read once per **operation** (only when the source had moved). Prototype 1's daemon (branch `feat/gr2-propagation-prototype-1`) deliberately hands its observation to the machine on the not-new path too, so the machine's cursor check runs on **every tick** — which is right, because a corrupted sink behind an unchanged source must not read as healthy — but it means `_rows()` now parses the whole journal every 30 s. The frequency of the undefended parse moved from once-per-change to once-per-tick, and the cost is the whole file each time.

## Expected

1. A torn **trailing** line (the only kind an append+fsync writer can produce) is tolerated: skipped, counted, surfaced once as a note or diagnostic, and never fatal. The intact prefix still answers the cursor check. **This is safe only because of a write-ordering invariant that today lives in a code comment:** `_acknowledge` appends the `ACKNOWLEDGED` transition row FIRST and advances the cursor AFTER ("the cursor is derived from it and repaired from it on replay, so the row is written first"). So a tear during acknowledgement leaves the cursor un-advanced and the operation simply replays; a torn trailing row can never be the only evidence the cursor depends on. Name that ordering as an invariant in the code (an assertion or a named helper, not only the comment), because a later reorder of those two writes would invalidate the tolerant read with nothing going red.
[...]

PR #896, opened 2026-08-20
fix(gr2): a torn journal line stops being fatal, and stops eating its neighbour

laynepenney:
The journal writer appends and fsyncs, so a kill between the two leaves a partial line. Three consequences, each with its own witness.

**A tear used to cost two rows, not one.** Appending onto an unterminated line glues the new row to the remnant, so a row that was serialized correctly and fsynced becomes unreadable. `_write` now terminates a remnant first, confining the damage to the line actually interrupted. This is the half the issue did not contain, and it is the reason the issue's discriminator does not hold.

**`_rows` raised a bare `JSONDecodeError` on any unparseable line.** The daemon lets anything outside its named failure list propagate, so one bad moment exited the loop on its first tick and on every restart after it. Unparseable lines are now dropped and counted, and `MalformedLine` keeps the drop from being silent.

**The drop is unconditional rather than trailing-only, because position does not discriminate cause.** A tear is trailing only until the daemon restarts and appends again, after which the same orphan sits mid-file with intact rows on both sides — so a position rule would raise on the writer's own crash-then-restart signature. The question that does discriminate is whether the cursor can still be accounted for, and `replay` already asks it and already raises `JournalInconsistent` by name. Dropping here lets that check decide instead of being pre-empted by a parse error. Skipping is safe because the acknowledgement row is written before the cursor advances, so a torn acknowledgement leaves the operation unacknowledged and it is redone idempotently.

**Idle ticks no longer re-parse the file.** Prototype 1 asks the machine to account for an observation on every tick, which moved this parse from once-per-change to once-per-tick. Rows are cached against `(size, mtime_ns)`; the `Propagator` is built once per loop and holds one `Journal`, so the cache spans ticks — verified rather than assumed, since a cache that never hits in production would be a green over a non-fix. The one state it cannot see, an in-place mutation changing neither size nor mtime_ns, is named in the code rather than assumed away.

## Verification

86 passed, 1 xfailed — up from 83 by exactly the three witnesses added, so nothing was lost or silently skipped.

Each guard is mutation-proven, and each mutation was restored byte-exact:

| Mutation | Kills |
|---|---|
| remove the newline repair | W1 only |
| restore the bare `json.loads` | W1 and W2 |
| disable the cache | W3 only |

The second killing two is a real dependency rather than a stray: W1 asserts on the rows that survive, so without the skip guard its torn orphan still raises. The guards compose, and that is reported rather than presented as three tidy isolations.

Ref #893 — closes at promotion.

Premium boundary: grip is OSS because this is local workspace state machinery — journal durability and read caching — with no identity, org, or entitlement semantics.

### DIFF
--- a/gr2/prototypes/propagation_state_machine.py
+++ b/gr2/prototypes/propagation_state_machine.py
@@ -102,6 +102,19 @@ class DestinationUnreadable(RuntimeError):
     """A read against the destination returned an error instead of an answer."""
 
 
+@dataclass(frozen=True)
+class MalformedLine:
+    """A journal line that could not be parsed, kept so a drop is never silent.
+
+    ``index`` counts lines in the file, blank ones included, so it names the line a
+    reader would count to by hand.
+    """
+
+    index: int
+    excerpt: str
+    reason: str
+
+
 class JournalInconsistent(RuntimeError):
     """The cursor names a revision the journal cannot account for.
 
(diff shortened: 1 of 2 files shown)


laynepenney:
r2 (Sentinel) — APPROVE, bound to head 37eced357e817634d0eeac9454be2cf7ce9132cf.

RAN: 86 passed, 1 xfailed across the state-machine and daemon suites with the checkout pinned on PYTHONPATH. Disabling the unchanged-file cache made only its cache witness fail; the restored source hash matched exactly.

READ: journal parsing, append repair, cursor ordering, and the three new witnesses. Scope: this public journal torn-line and cache change at the bound head.

